PR #175, opened 2010-10-28
Allow the Rack handler to be configured via Capybara.server

jfirebaugh:
Followup to discussion at http://github.com/jnicklas/capybara/commit/ac4a340d04831630629aa4a23d1d8e139f5a8d5f

For example, to use mongrel rather than thin or webrick:

```
Capybara.server do |app, port|
  require 'rack/handler/mongrel'
  Rack::Handler::Mongrel.run(app, :Port => port)
end
```

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -108,6 +108,38 @@ module Capybara
       @drivers ||= {}
     end
 
+    ##
+    #
+    # Register a proc that Capybara will call to run the Rack application.
+    #
+    #     Capybara.server do |app, port|
+    #       require 'rack/handler/mongrel'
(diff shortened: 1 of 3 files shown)


mathieujobin:
this is a great feature.
is there documentation somewhere so can configure it using unicorn, or puma or thin without knowing the ins and out of each?

thanks

twalpole:
@mathieujobin puma is already there, just set `Capybara.server = :puma` - https://github.com/jnicklas/capybara/blob/master/lib/capybara.rb#L481 - for unicorn or thin that would be up to you - see the docs for Capybara::register_server.  In the future please ask howto questions in the mailing list or on gitter as requested in the README

Issue #327, opened 2011-04-20
API for cookie manipulation?


agibralter:
Would it make sense/be possible to support an API for cookie manipulation?

```
get_cookie(name)
#=> return String with cookie that is set in user's browser session

set_cookie(name, value)
#=> set a cookie manually in the user's browser session

delete_cookie(name)
```

I don't know if it would work for all drivers... but it would certainly work for Rack, Selenium, and Webkit.

PR #357, opened 2011-05-23
Querying State of Cookies

mcrmfc:
Hi,

This is something I had previously been patching in but seeing as there is now a discussion on [issue #327](https://github.com/jnicklas/capybara/issues/327) I thought I would share this.

I feel that there are valid use cases for checking the state of the cookies, for example my team builds core components which will be shared across many sites...we still want to do full stack testing on our reference applications/example pages but we need to be absolutely sure the state of cookies we add is exactly as we expect as this will impact many other teams.

Have only added querying at the moment, maybe the argument for setting and deleting cookies is a little less strong...but as people have said on [#327](https://github.com/jnicklas/capybara/issues/327) I can see scenarios where you might want to.

So...feel free to reject this if you are planning to implement something more structured in a future release.  I can see an argument that it would be better to change/improve the Rack Test api than maybe hack out the cookies as we have to now.

Cheers,

Matt.

### DIFF
--- a/lib/capybara/rack_test/driver.rb
+++ b/lib/capybara/rack_test/driver.rb
@@ -48,7 +48,8 @@ class Capybara::RackTest::Driver < Capybara::Driver::Base
         :value => cookie.instance_variable_get(:@value),
         :path => cookie.instance_variable_get(:@options)['path'],
         :domain => cookie.instance_variable_get(:@options)['domain'],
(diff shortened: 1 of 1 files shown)


frewsxcv:
In case anyone discovers this, it's now possible to access the cookie jar without using `instance_variable_get`:

```ruby
browser
  .rack_mock_session
  .cookie_jar
```

Issue #418, opened 2011-07-18
Disable reset of session in rspec


kberridge:
By default capybara always resets the session after each test.  But there are times when I would prefer that management of the session state was left up to me (ex: to avoid the need to login before every test).

PR #419, opened 2011-07-18
Adds reset_session_after_each config option

kberridge:
Closes issue #418.

Allows user to set Capybara.reset_session_after_each = false.
This will prevent the session from being cleared after each rspec test.


sethuster:
I would also like this to be reconsidered.  There are times when it would be preferable to not have the session reset after each spec.

abotalov:
+1
User may want to use other mechanism to make sure that tests are relatively independent (https://github.com/rspec/rspec-core/issues/1811)

z41:
Migrating from page-object to site-prism, getting into the same issue. Monkey patching again is my tool, that's pity.

z41:
One of scenarios when having such an option can really help: 
I have pages with loading time about 30 seconds or even more (search results pages) and this would hardly be improved. So instead of loading the page every time I just reset its state. And the only option I have with capybara is to monkey patch it =(

PR #505, opened 2011-09-24
Adding support for getting an element's parent

andyw8:
It seemed Capybara lacked a simple way to get an element's parent ( http://stackoverflow.com/questions/4861863/how-to-get-parent-node-in-capybara ) so I've added support for this.

### DIFF
--- a/lib/capybara/node/simple.rb
+++ b/lib/capybara/node/simple.rb
@@ -118,6 +118,16 @@ module Capybara
         native[:selected]
       end
 
+      ##
+      #
+      # The parent of the node
+      #
+      # @return [Capybara::Node::Simple]   The parent element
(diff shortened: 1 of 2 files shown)


janko-m:
What does `Capybara::Node::Base#parent` attribute currently do? It seems to be always returning `Capybara::Document` instead of the actual tag's parent.

twalpole:
@janko-m Its the context for the find which located the current element.

abotalov:
@janko-m `Base#parent` is internally used to get an ancestor using which the current element was located. For example in:
```ruby
el = find('a')
el.find('b').parent # => returns el
```

janko-m:
Got it! Thanks guys.

PR #660, opened 2012-03-06
Bubble up errors that aren't handled by applications/frameworks

raggi:
Currently, people do things like make servers puts or use Thread.abort_on_exception = true to do this. That leads to other problems, so I added some code to cache the errors and re-raise in the testing thread.

In general this will display errors that are released / uncaught by frameworks (e.g. rails in test mode). Sometimes, where the first request doesn't error, but subsequent asset / ajax type requests do, it will raise those errors instead. Whilst this may be slightly odd, it will help with having a faster feedback loop most of the time.

Ideally it wouldn't rely on driver.rack_server. I'd actually prefer that session handled server booting when drivers request servers, however, this would be a larger refactor that is not appropriate for rolling up into these changes.

### DIFF
--- a/lib/capybara/dsl.rb
+++ b/lib/capybara/dsl.rb
@@ -37,9 +37,22 @@ module Capybara
       Capybara.current_session
     end
 
+    ##
+    #
+    # If using a driver that runs a rack server, check for errors will raise any
+    # errors that were not handled inside the application.
+    #
+    # @return [NilClass] If no errors were raised
+    def check_for_errors
+      return unless page.driver.respond_to?(:rack_server)
+      e = page.driver.rack_server.last_error
+      raise e if e
+    end
+
     Session::DSL_METHODS.each do |method|
       class_eval <<-RUBY, __FILE__, __LINE__+1
         def #{method}(*args, &block)
+          check_for_errors
           page.#{method}(*args, &block)
         end
       RUBY


Systho:
I know this issue has been officially closed but at the same time it does not yet work as expected. @jnicklas  do you have any idea on how you want to tackle this problem ? I think more and more app are using asynchronous drivers so I guess this problem will only amplify.

yevgenko:
I come to the same requirements, i.e. bubble up errors, as my initial intent was to have capybara working with new thing (rich UI, ajax, etc) as it was with more "plain" rails apps, I mean the one where you can disable JS safely.

What I didn't realise in a first place is that the new thing comes with quite different architecture, where we previously had a humble client (browser), view and controller, **now** we have rich communication protocol `Client <-> Server`, where browser provides environment for client, but the client undertakes responsibility for making decision "when", "what" and "where" to send request. Even if we keep it thin, and ideally we should, it is now client's concern to interpret DOM events and issue  requests to the server.

Essentially it means more layer and more tests needed to cover this architecture in end-to-end fashion, basically:

1. (capybara, with full app) End-to-End, same as before
2. (capybara, with mock app) UI makes request, e.g.: `it 'makes request when button clicked'` 
3. (NOT capybara) Server handles request, e.g.: `it 'creates resource'` (in case of RESTful style, but can be something else)

When DOM element not found in end-to-end scenario, I've to go to my UI test and make sure it makes "proper" request, than I go to the request test (3rd level) to make it handle request.

By the time 3rd level pass, the 1st should pass too (assuming it is as simple as CRUD, otherwise there will be more levels)

It is request tests (3rd level) where I expect to see an unhandled server-side exceptions and it just works, they on the same thread.

But there is more, I can share matchers and helpers between 1st and 2nd layers, I can setup 2nd layer to throw JS exceptions too and at the end I have clear documentation and regression picture, consider answering this question otherwise: "Is this exception because frontend send wrong request, because backend didn't handle it properly or something else?".

Perhaps sample app would help to clarify this point, I haven't build one yet, but I've put some early outlines of how tests might looks like with RESTful architecture on Rails:
http://www.beardedhacker.com/blog/2018/07/29/growing-web-apps-guided-by-tests/

I'd appreciate some early critics on this approach.

Thank you in advance.

yevgenko:
I've made sample web app which can illustrate that the necessities for bubbling unhandled exceptions in capybara can indicate design issues (particularly tests design). Sample app demonstrates that feedback loop can be maintained efficiently without re-raising exceptions.

Source on GitHub with more details in PR description:
https://github.com/yevgenko/rich-and-restful/pull/1

And I'd appreciate feedback and questions.

twalpole:
@yevgenko I'm not sure why you're commenting on an issue closed 3 years ago, and nor do I understand what you're trying to show in your app.  If you have an actual issue with Capybara please create a new issue that explains what your actual issue is with a reproducible example.  Note however that we will not be removing the re-raising of server errors in Capybara since this is done specifically to catch issues in the error handling of the app.

Locking this issue.

PR #958, opened 2013-02-09
Adding asset host support

sdhull:
### Asset Host Support!

This adds asset host support to `save_and_open_page`

This takes the super-simple approach of prepending `Capybara.asset_server` to asset `src`'s or `href`'s (in the case of links). It may be simple, but it should be sufficient to get people's pages rendering with assets loading from other server locations (making no assumptions about where that server might be running).

So, for instance, if your assets were rendering in `save_and_open_page` like this:

``` html
<link href="/assets/application.css" media="all" rel="stylesheet" type="text/css">
```

Then you could add this line to `spec_helper.rb`:

``` ruby
Capybara.asset_server = "http://localhost:3000"
```

(please note, no trailing slash)

And then in your opened page from `save_and_open_page`, that same asset would be rendered as:

``` html
<link href="http://localhost:3000/assets/application.css" media="all" rel="stylesheet" type="text/css">
```

If you don't wish to accept as-is, please tell me what I can do to make it acceptable or if there are some untenable issues that prevent acceptance.

Thank you!

### DIFF
--- a/lib/capybara/spec/session/save_page_spec.rb
+++ b/lib/capybara/spec/session/save_page_spec.rb
@@ -56,6 +56,15 @@ Capybara::SpecHelper.spec '#save_page' do
       result.should include("<head><base href='http://example.com' />")
     end
 
+    it "doesn't prepend base tag to pages when asset_host is nil" do
+      Capybara.asset_host = nil
+      @session.visit("/with_js")
+      path = @session.save_page
+
+      result = File.read(path)
+      result.should_not include("http://example.com")
+    end
+
     it "doesn't prepend base tag to pages which already have it" do
       @session.visit("/with_base_tag")
       path = @session.save_page


MrHubble:
Should `asset_host` and its usage be described in the docs?

sdhull:
@MrHubble [it is in the docs](https://github.com/jnicklas/capybara/pull/958/files?diff=unified#diff-f6aa6c79750a9a5bd39ee17f3ade2974R40), however it may be worth putting in the readme as well?

daveharris:
Just in case this helps someone else, it seems that something changed in rails 5.1 or 5.2, where the asset URL contained the digest, meaning that even though the `<base href="http://localhost:3000">` was set, the CSS wasn't found because the sha1 didn't that in development (which presumably doesn't have `config.assets.digest` turned on.

I had an issue where the `test` environment was using digested assets, whereas in development it wasn't. Therefore, a the below CSS file would not be found:
```
<base href="http://localhost:3000">
<link rel="stylesheet" media="all" href="/assets/application-7212b5b37cedcaab6e0d0022913294a085afdcd8b68bdbc554717d865422d97e.css">
```

The fix for this is to edit `config/environments/test.rb` to instruct it not to created digested assets:
```
config.assets.digest = false
```

Now, the below is produced, and the CSS file is found
```
<base href="http://localhost:3000">
<link rel="stylesheet" media="all" href="/assets/application.css">
```

I hope this helps someone else!

twalpole:
@daveharris Generally when using `asset_host` you'd want to precompile the test environment assets which would leave them in the projects public directory and available (with digest in the name) for the dev server to provide.  That way you don't need to turn off digesting in the test environment, or deal with assets that may have been compiled with different settings than you expect (config.assets.debug true) and therefore missing parts.

Issue #923, opened 2013-01-05
Feature request: support css pseudo-selectors in 'find'


Trevoke:
I'd like to be able to do this:

```
find('input["name=desired_amount"]:invalid')
```

Right now, using Selenium, I get the following error:

```
     Selenium::WebDriver::Error::InvalidSelectorError:
       The given selector .//input[@name = 'desired_amount' and invalid(.)] is either invalid or does not result in a WebElement. The following error occurred:
       InvalidSelectorError: Unable to locate an element with the xpath expression .//input[@name = 'desired_amount' and invalid(.)] because of the following error:
       [Exception... "The expression is not a legal expression."  code: "12" nsresult: "0x805b0033 (SyntaxError)"  location: "file:///var/folders/gc/3hx7nhq5257_6dzxblc3s1k00000gn/T/webdriver-profile20130104-35671-uj1dp/extensions/fxdriver@googlecode.com/components/driver_component.js Line: 5697"]
```

Using Webkit:

```
Failure/Error: find("input[name='desired_amount']:invalid")
RuntimeError:
  xmlXPathCompOpEval: function invalid not found
```

PR #961, opened 2013-02-16
dont convert css to xpath

twalpole:
@jnicklas - Seeing issue #923 get closed this morning (request for :invalid pseudo selector support), it made me revisit something I'd been meaning to look at for a while.  Why always convert css to xpath when browsers can handle direct css searches?  This PR is the beginning of work on allowing css selectors to just flow through and be searched for as css, which would allow for css pseudo selectors to be used on browsers that support them.  Currently this code would require drivers to implement a find_css method which is obviously not optimal. I think a better way may be for find to  take a second parameter indicating whether the selector is xpath or css, and a way for the driver to indicate whether it will accept css, but this would break the current driver api.  I've also implemented the :disabled and :enabled pseudo selectors in the RackTest driver as proof that some of the css pseudo selectors can be implemented there (obviously not :invalid).  Your thoughts on how best to implement this, and whether or not you would consider it are appreciated before I finish it up

### DIFF
--- a/lib/capybara/node/finders.rb
+++ b/lib/capybara/node/finders.rb
@@ -145,7 +145,6 @@ module Capybara
 
       def resolve_query(query, exact=nil)
         elements = synchronize do
-          # base.find(query.xpath(exact)).map do |node|
           if query.selector.format==:css
             base.find_css(query.css)
           else


ioquatix:
Sorry to wake up an old issue, but if `e = find('textarea:invalid')` gives an element, wouldn't you also expect e.has_css?(':invalid')` to be true?

twalpole:
@ioquatix has_css? Looks for children that match the given css, it doesn't compare the element it's called on

PR #976, opened 2013-02-25
Hover test and support

twalpole:
This implements hover on an element which works with selenium and chrome. Unfortunately hover in selenium with firefox appears to be currently broken


lynndylanhurley:
Is hover in selenium with firefox still broken?

twalpole:
@lynndylanhurley You could try it rather than posting on a PR that was closed 2 years ago - or ask in the mailing list as requested in the README - but no its not still broken since the Capybara tests pass for it.

PR #1169, opened 2013-09-25
fixes #1168

### LINKED ISSUE #1168: Automatic string casting for selectors including "with: ..."
I observed an unexpected behaviour using the test

``` ruby
expect(page).to have_field 'foo_parent_id', with: predecessor.id
```

that fails. While fiddling around I found out that `predecessor.id` is in fact of type `Moped::BSON::ObjectId`

Explicit casing to string solves the problem

``` ruby
expect(page).to have_field 'foo_parent_id', with: predecessor.id.to_s
```

what in my point of view is not correct rails behaviour, because within such an input field I can just use a string. 

Therefore I suggest to use `filter(:with) { |node, with| node.value == with.to_s }` in https://github.com/jnicklas/capybara/blob/master/lib/capybara/selector.rb?source=cc#L106 or I am totally wrong?

gurix:
- Description with value
- Casting automatic values passed as object to string
- Added some tests for have_fields selector including a value

As discussed and fixing https://github.com/jnicklas/capybara/issues/1168

### DIFF
--- a/spec/capybara_spec.rb
+++ b/spec/capybara_spec.rb
@@ -31,7 +31,7 @@ describe Capybara do
     end
 
     it "should default to a proc that calls run_default_server" do
-      mock_app = mock('app')
+      mock_app = double('app')
       Capybara.should_receive(:run_default_server).with(mock_app, 8000)
       Capybara.server.call(mock_app, 8000)
     end


klondikemarlen:
And years later its still confusing ...

If you want the old way to work you can create a class that evaluates to nil when to_s is called.
e.g.
```ruby
class NilString
  def to_s
    nil
  end
end

expect(page).to have_field("Subject", with: NilString.new)
```

PR #1348, opened 2014-06-30
Allow Session#visit to visit absolute URLs with no scheme.

dstosik:
Hello,

Looking at [the code](https://github.com/jnicklas/capybara/blob/master/lib/capybara/session.rb#L209), it looks like Capybara considers any URL without scheme to be relative.
Actually, an absolute URL can have an empty scheme: when encountering `//example.com`, browsers will use the same scheme as the current page (thus https://example.com if you're currently on an HTTPS page, or http://example.com otherwise).

Please find my suggestions in this pull request.

Regards,

David

### DIFF
--- a/lib/capybara/spec/session/visit_spec.rb
+++ b/lib/capybara/spec/session/visit_spec.rb
@@ -17,6 +17,15 @@ Capybara::SpecHelper.spec '#visit' do
     expect(@session).to have_content('Another World')
   end
 
+  it "should fetch a response when absolute URI doesn't have a scheme" do
+    # Preparation
+    @session.visit('/')
+    root_uri = URI.parse(@session.current_url)
+
+    @session.visit("//#{root_uri.host}:#{root_uri.port}/")
+    expect(@session).to have_content('Hello world!')
+  end
+
   it "should fetch a response when absolute URI doesn't have a trailing slash" do
     # Preparation
     @session.visit('/foo/bar')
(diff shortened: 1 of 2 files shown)


twalpole:
I missed this one completely -- will look into it for 2.6

lib/capybara/spec/session/visit_spec.rb
```diff
@@ -77,6 +86,12 @@
       serverless_session.visit("http://#{@session.server.host}:#{@session.server.port}/foo")
       expect(serverless_session).to have_content("Another World")
     end
+
+    it "should visit an absolute URL without scheme" do
```
twalpole: Is this actually a valid test? Isn't this just testing that scheme defaults to http if none is specified, and is that really valid to do?

twalpole:
So I pulled the tests for this into another branch and am rewriting visit using URI.merge --- I added a line note on one of the tests since I'm not really sure what its testing -- feedback would be great (and sorry for this having taken so long)

jnicklas:
`visit` is a bit strange in the sense that it does accept relative URLs.

There isn't really any user-facing feature in browsers which behaves this way, the user can only ever enter absolute URLs, even thought the browser might fill in the scheme part. I tested in Safari, Chrome and Firefox and all of them do allow protocol relative URLs and actually fill in the scheme, however, the scheme they choose is `file:` and not `http:`.

But this isn't really the same behaviour as Capybara has for `visit` anyway, since `visit` does allow relative URIs, which is more like clicking a link behaves in browsers, though in that case the user wouldn't actually type the address.

So Capybara's current state of this method is probably a bit broken, since it doesn't actually simulate any user behaviour, which is what Capybara claims to do. Changing it would break compatibility though, so it's probably not a great idea.

I'm not sure what the conclusion of this is. It would be consistent with consistent with allowing relative URIs to allow protocol relative URIs as well, but allowing relative URIs is kinda broken in the first place so  ¯\_(ツ)_/¯.

twalpole:
After thinking about this and looking through the tests I think all this boils down to is defaulting scheme to http when not specified, which I'm not convinced is necessarily valid.  I definitely agree with @abotalov that we should not be resolving against current_url for calls to Session#visit, since calls to visit really are about setting up the state of the browser in order to then interact with it, not about loading stuff relative to the current page.

PR #1373, opened 2014-08-01
add method current_path_query to Capybara::Session

ice5050:
current_path_query is a method that return Path and query of the current page without any domain information


twalpole:
The functionality this would generally be useful for is provided by the has_current_path?/have_current_path/assert_current_path series of boolean and matchers

PR #1404, opened 2014-09-22
Add Selenium example project link to readme

shaunlebron:
People need examples for help getting started, and this example project helps a lot.


twalpole:
I don't think we want to add a random sample project to the readme -- might be better info for a wiki

PR #1405, opened 2014-09-23
Add current_fullpath method that returns path with query string.

cirosantilli:
As redundant as `current_path` and `current_host` (URI + `current_url`), but a more common use case than `current_host`, so I think that the convenience is worth it.

80 votes on SO: http://stackoverflow.com/questions/5228371/how-to-get-current-path-with-query-string-using-capybara/25999027#25999027

I will add tests if the feature is accepted.

### DIFF
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -42,7 +42,7 @@ module Capybara
       :title, :assert_title, :assert_no_title, :has_title?, :has_no_title?
     ]
     SESSION_METHODS = [
-      :body, :html, :source, :current_url, :current_host, :current_path,
+      :body, :html, :source, :current_url, :current_host, :current_path, :current_fullpath,
       :execute_script, :evaluate_script, :visit, :go_back, :go_forward,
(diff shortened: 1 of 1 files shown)


cirosantilli:
http://stackoverflow.com/questions/5228371/how-to-get-current-path-with-query-string-using-capybara/25999027?noredirect=1#comment44797830_25999027

woto:
+1 for this feature

twalpole:
@woto if you want stable tests you should be using the path assertions/matchers and have no need for this method

woto:
@twalpole There is a problem, that I have more query string params, which I do not want to test in this `test`. For now decided to do like this:
```
expect(::Addressable::URI.parse(current_url).query_values).to include('q' => q.to_s)

```

twalpole:
Since using any of the standard RSpec matchers with current_path/url is asking for flaky tests I would probably write that using the `have_current_path` matcher with a regex instead. Along the lines of

    expect(page).to have_current_path(/q=#{q}/)

woto:
Cool. I understood the reason. Thanks!

PR #1407, opened 2014-09-25
Enable the #fscenario example alias

elia:
Not sure why it's disabled, but since I expected to find it working I think it would be useful to enable it

cc: @twalpole (6c0674db9bb6896d4ced8f8ab12fc0e3eec6d71e)


e2:
:+1 - this should be released! 

I can confirm it works fine with RSpec monkey patching disabled.

PR #1421, opened 2014-10-13
Add https server launch control support

mender:
It allows to start https server inside `Capybara.server` block and control its readiness with `#responsive?` method.


twalpole:
his was implemented separately but very similarly for v 3.1 - and is going to have the client refactored out into its own class so will be very close to this code.  Closing this PR.

PR #1434, opened 2014-10-29
Added methods #previous #next on Node Element to locate closest previous...

neilma:
#next() finds the next sibling node of the current node.
#previous() finds the first previous element of the current node.
Specs are passing.

### DIFF
--- a/lib/capybara/node/element.rb
+++ b/lib/capybara/node/element.rb
@@ -258,6 +258,17 @@ module Capybara
       rescue NotSupportedByDriverError
         %(#<Capybara::Element tag="#{tag_name}">)
       end
+
+      def previous(html_tag=nil)
+        evaluated_html_tag = html_tag.nil? ? "*" : html_tag
+        self.first(:xpath, "(ancestor::#{evaluated_html_tag}[1]|preceding::#{evaluated_html_tag}[1])[last()]")
+      end
+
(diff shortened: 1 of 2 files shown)


AlexWayfer:
> IMO this isn't that useful in Capybara. Capybara's tagline is:
> 
> > Capybara helps you test web applications by simulating how a real user would interact with your app.
> 
> A real user would never care if two elements are adjacent to each other.

When you see a label like "Do you have something?" and following button toggles like "Yes" and "No" — user cares what about these are "Yes" and "No", they tracks its following after the label with description. So I think you're wrong and this is helpful ruined  PR.

twalpole:
@AlexWayfer This issue is from over 6 years ago.  Additionally, users care about page layout not document structure.  If you need to find elements based on the page location of other elements use the spatial filters supported by the finders - https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selector.rb#L12

AlexWayfer:
> @AlexWayfer This issue is from over 6 years ago.

I know. But didn't find such useful methods in actual Capybara.



> Additionally, users care about page layout not document structure. If you need to find elements based on the page location of other elements use the spatial filters supported by the finders - https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selector.rb#L12

OK, I'll look into it, thanks.

PR #1439, opened 2014-11-07
Make elements returned by `all` and `first` reloadable

### LINKED ISSUE #1410: Make elements returned by `all` and `first` reloadable
I think it would be a good idea to make elements returned by `Finders#all` reloadable, i.e.:
* set `Element#allow_reload` to true for every element from returned result. It may be done inside `Finders#all` or `Query#resolve_for`
* Pass `index` as the last argument inside `Query#resolve_for` (It will be not required argument that defaults to `0` just to not break backwards compatibility as `@api private` isn't added for `Element#initialize`)
* Inside `Element#reload` use `all(...)[index]` instead of `first(...)`

Do you like that idea?

abotalov:
Should fix #1410

All elements are considered reloadable. `#allow_reload!` may be removed from `Element` and `Simple` in next major version as it's not needed with this PR.

One of moments to discuss - currently behavior may be unexpected e.g. in the following case:
```ruby
el = all('li')[-1] # there are 4 'li' elements on the page
find('.remove_last_li').click
expect { el.text } not_to raise_error # I believe that will raise an error as that will search for third element, not for the last one
```

But I don't know at the moment a good solution for it. The problem is that element of enumerable doesn't know the way using which it was got from that enumerable.

### DIFF
--- a/lib/capybara/node/element.rb
+++ b/lib/capybara/node/element.rb
@@ -242,15 +242,12 @@ module Capybara
       end
 
       def reload
-        if @allow_reload
-          begin
-            reloaded = parent.reload.first(@query.name, @query.locator, @query.options)
-            @base = reloaded.base if reloaded
-          rescue => e
-            raise e unless catch_error?(e)
-          end
-        end
+        @index ||= 0 # just for backwards compatibility (in case if `Element` wasn't modified by `Result`)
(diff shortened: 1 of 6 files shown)


twalpole:
Closing due to @jnicklas huge :-1: and my misgivings about the potential unexpected behavior

PR #1448, opened 2014-12-01
Use monotonic clock for measuring passed time

ledestin:
Hello,

Capybara uses real clock to measure time which leads to races when clock is moved 1 hour ahead. This pull requests switches Capybara to use monotonic clock, which doesn't have such problems.

I've used a gem (that depends on hitimes) because supported Ruby includes 1.9.3 and monotonic clock is introduced only in Ruby 2.


twalpole:
Closing in favor of the lesser dependencies (and admittedly lesser platform support) in PR #1503

PR #1465, opened 2015-01-17
Update rubies; Remove BUNDLE_JOBS as it's not respected by Travis

abotalov:


twalpole:
Is not respecting BUNDLE_JOBS a new change on Travis?? I'm pretty sure it used to respect it

abotalov:
See:
https://github.com/calagator/calagator/commit/87169d023b7143d21d3cd399158dbd97a7ac0579
https://github.com/teampoltergeist/poltergeist/pull/499#issuecomment-54663015

After some search in travis-build/travis-build I found that first commit that added `--jobs=3` is probably this one - https://github.com/travis-ci/travis-build/commit/6cb7c4d8c0f9b89b7d95b46ef1174cb327dc1538 and current implementation is here (https://github.com/travis-ci/travis-build/blob/a9f5fe32fffb4d4ceee2062770c7c50aa3098d3a/lib/travis/build/script/shared/bundler.rb).

So travis-build sets `--jobs=3` without checking for BUNDLE_JOBS env variable. Probably this is a bug  in Travis but I think `--jobs=3` is fine for us.

cc @joshk

twalpole:
ok 3 vs 4 wouldn't make much difference anyway -- merging -- thanks

Issue #1459, opened 2014-12-23
Problem with send_keys usage


NameFILIP:
When sending a special keys, like arrow keys ("up", "down", "left", "right") there is a different syntax for different drivers

For the selenium's webdriver I use:
```
find('.some-input').native.send_keys('faci', :down, :down, :down, :down, :enter)
```
For the poltergeist:
```
find('.some-input').native.send_keys('faci', :Down, :Down, :Down, :Down, :Enter)
```

The root of this problem is that Selenium keys differ from PhantomJS keys:
https://selenium.googlecode.com/svn/trunk/docs/api/rb/Selenium/WebDriver/Keys.html
https://github.com/ariya/phantomjs/blob/master/src/modules/webpage.js#L465

It would be nice if capybara has some mapping or restrictions for this, so that you can easily switch between different drivers.

twalpole:
@jferris @jonleighton @mhoran  - any thoughts on supporting a unified definition/mapping for special keys?

jferris:
I'd be fine mapping a symbol or constant to a key press, like `send_keys(:down)` or `send_keys(Capybara::KEY::DOWN)`.

twalpole:
@jferris Would you see this being handled in each driver, or having the drivers provide the mapping and #send_keys in their node classes - and capybara doing the translation in Capybara::Node::Element#send_keys ?

jferris:
I think you'd need two mappings:

* The list of normalized key codes mapped to the list of driver-specific key codes
* The strategy for sending one or more key codes to a driver

I'm not sure how much added value there is in having the first mapping. Expecting the drivers to normalize the key codes seem okay to me, but as a driver author, I don't think it would be more difficult for me to implement one approach over the other, so I'm fine with either.

abotalov:
Some my opinions:
1. `Capybara::Node::Element#send_keys` shouldn't map keys (like `:down`, `:shift`, etc.). Mapping should happen at driver side.
2. Constants like `Capybara::KEY::SHIFT` won't be useful for users. Symbols are better.
3. It makes sense to standardize to Selenium keys (http://www.rubydoc.info/gems/selenium-webdriver/Selenium/WebDriver/Keys) as they are documented in [draft spec](https://w3c.github.io/webdriver/webdriver-spec.html#widl-WebElement-sendKeys-void-DOMString-value).

[...]

PR #1472, opened 2015-01-23
Add Node#Send keys to standardize usage across drivers

twalpole:
Implementation of item discussed in Issue #1459  -  @jferris @jonleighton @mhoran Thoughts on implementing this in your drivers for Capybara 2.5 ?

### DIFF
--- a/lib/capybara/node/element.rb
+++ b/lib/capybara/node/element.rb
@@ -148,6 +148,79 @@ module Capybara
       def double_click
         synchronize { base.double_click }
       end
+      
+      ##
+      #
+      # Send Keystrokes to the Element
+      #
+      # @param [String, Symbol, Array]
+      #
+      # Examples:
+      #
+      #     element.send_keys "foo"                     #=> value: 'foo'
+      #     element.send_keys "tet", :left, "s"   #=> value: 'test'
+      #     element.send_keys [:control, 'a'], :space   #=> value: ' '
+      #
+      # Symbols supported for keys
+      # :cancel
+      # :help
+      # :backspace
+      # :tab
+      # :clear
+      # :return
+      # :enter
+      # :shift
+      # :control
+      # :alt
+      # :pause
+      # :escape
+      # :space
+      # :page_up
+      # :page_down
+      # :end
+      # :home
+      # :left
+      # :up
+      # :right
+      # :down
+      # :insert
+      # :delete
+      # :semicolon
+      # :equals
+      # :numpad0
+      # :numpad1
+      # :numpad2
+      # :numpad3
+      # :numpad4
+      # :numpad5
+      # :numpad6
+      # :numpad7
+      # :numpad8
+      # :numpad9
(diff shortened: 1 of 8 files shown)


lib/capybara/node/element.rb
```diff
@@ -148,6 +148,87 @@ def right_click
...
+      # @param [String, Symbol, Array]
+      #
+      # Examples:
+      #
+      #     element.send_keys "foo"                     #=> value: 'foo'
+      #     element.send_keys "tet", :arrow_left, "s"   #=> value: 'test'
+      #     element.send_keys [:control, 'a'], :space   #=> value: ' '
+      #
+      # Symbols supported for keys
+      # :null
```
abotalov: trailing space

mhoran:
There are a few keys that we can't send in capybara-webkit

```ruby
:left_shift
:left_control
:left_alt
:arrow_left
:arrow_up
:arrow_right
:arrow_down
:separator
:decimal
:command
```

There is no differentiation between shift and left shift, control and left control, or alt and left alt in Qt. We only have left, up, right, and down mappings, but nothing corresponding to "arrow".

Is `:separator` `|`? Is `:decimal` `.`?

Command is not a cross-platform key, so it is not supported by Qt. On OS X, `Qt::Key_Control` maps to command, and `Qt::Key_Meta` maps to control. We could conceivably map these as appropriate on each individual platform, but that may be surprising for users.

Finally, what should `:null` send?

twalpole:
@mhoran - The list of symbols is what selenium understands
:left_shift, :left_alt, :left_control are just aliases for :shift, :alt, :passport_control: 
:arrow_up, :arrow_down, :arrow_right, :arrow_left  are aliases for :up, :down, :right, :left.    
:command is an alias for :meta.  
:decimal is the numeric keypad .  
:add, :divide, :subtract, :multiply are the numeric keypad instances of +, /, -, *

:separator is the numeric keypad "seperator" key (different from the :decimal key) - the keycode is 108 and I'm guessing its a key returned by non-english keyboards although I can't find info that specifically states that.

:null - appears to be used in lower level key routines in selenium to trigger release of all modifier keys.  It is probably not needed for this so I'm removing it from the list

jferris:
> There is no differentiation between shift and left shift, control and left control, or alt and left alt in Qt. We only have left, up, right, and down mappings, but nothing corresponding to "arrow".

I'm guessing that if we just sent shift, control, and alt, that would be least surprising to users. Same goes for up/arrow_up.

twalpole:
@jferris I have removed the arrow_x and left_x aliases from the list of required supported key symbols - For future info DOM Level 3 does support identifying left vs right instances of shift, alt, control via the location attribute of a KeyboardEvent however not all browsers support it at the moment https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent.location and selenium doesn't currently provide a way to set it, so we may have to revisit left_shift, right_shift, etc in the future

PR #1477, opened 2015-02-03
Allow empty attributes by default for Browser#submit and Driver#submit

gaganawhad:
Not sure if default value for `attributes` in `#submit` wasn't added intentionally, but adding it seems to improve the API


twalpole:
Pretty sure this was intentional since they're not really part of the public API and submitting a form without its parameters doesn't really make sense.

gaganawhad:
Got it! That makes sense. My use case was `page.driver.submit :delete, '/sign-out', {}` which got me wondering...

PR #1486, opened 2015-02-25
Add table of contents to README

eoinkelly:
The README is large (yay!) and I refer to it regularly so I thought a TOC would be helpful.


twalpole:
thanks

PR #1487, opened 2015-02-26
Fix API doc references to non-existent Capybara::Element

### LINKED ISSUE #1485: API docs refer to Capybara::Element not Capybara::Node::Element
Very minor thing but the API docs refer to `Capybara::Element` which does not exist e.g. https://github.com/jnicklas/capybara/blob/master/lib/capybara/node/finders.rb#L25 - I presume they should refer to `Capybara::Node::Element`?

eoinkelly:
Fixes #1485

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -116,7 +116,7 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
 
   ##
   #
-  # Webdriver supports frame name, id, index(zero-based) or {Capybara::Element} to find iframe
+  # Webdriver supports frame name, id, index(zero-based) or {Capybara::Node::Element} to find iframe
   #
   # @overload within_frame(index)
   #   @param [Integer] index                 index of a frame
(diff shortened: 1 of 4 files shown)


lib/capybara/selenium/driver.rb
```diff
@@ -116,7 +116,7 @@ def reset!
 
   ##
   #
-  # Webdriver supports frame name, id, index(zero-based) or {Capybara::Element} to find iframe
+  # Webdriver supports frame name, id, index(zero-based) or {Capybara::Node::Base} to find iframe
```
twalpole: Lets leave this as Capybara::Node::Element,  while its technically true that it can take a Base, Base isn't really meant for public use and is just used as the superclass for Element and Document. Document makes no sense being passed in as a frame element.

eoinkelly: :+1:

twalpole:
Ok - so if you make the one change I indicated and squash the two commits together, I'll merge this in -- thanks

eoinkelly:
Done :-)

Issue #1319, opened 2014-05-31
Propagate fill_in options to Element#set


sobrinho:
# TL;DR

We need to be able to customise the behaviour of `Node#set` somehow on our driver to trigger blur or not after filling the field, i.e.:

``` ruby
fill_in 'Employee', :name => 'Gabriel Sobrinho', :blur => false
```
# Long Version

Capybara's fill_in extract the :with option and pass it to `Node#set` and we are not able to receive any option from the user.

That's necessary in some specific use cases where the behaviour for fill_in needs to be slight different.

The default behaviour for poltergeist and probably other drivers, as can be seen [here](https://github.com/teampoltergeist/poltergeist/blob/master/lib/capybara/poltergeist/client/agent.coffee#L181-L203), is:
1. Clear input value
2. Trigger focus event
3. Trigger keydown event
4. Append the character
5. Trigger keypress event
6. Trigger keyup event
7. Repeat 3 to 6 for every received character
8. Trigger change event
9. Trigger input event
10. Trigger blur event

We need to be able to be able to receive on poltergeist if the blur event must happen or not.

It's really useful for autocompletes, i.e.: Typeahead and jQuery UI Autocomplete, where the autocomplete options is shown after you stop typing but is still focused on the input.

When the input loses the focus, it get's closed, so we can't chose any option from autocomplete.

There is some recommendations on web to circumvent this "problem" using a custom helper, like this:

``` ruby
  def fill_autocomplete(locator, options)
    value = options.delete(:with)

    field = find(:fillable_field, locator, options)
    field.set(value)

    page.execute_script "$('##{field[:id]}').autocomplete('search')"

    expect(page).to have_selector '.ui-autocomplete .ui-menu-item a'

    page.execute_script "$('.ui-autocomplete .ui-menu-item a:contains(#{value})').trigger('mouseenter').trigger('click')"
  end
```

Receiving the options on `Element#set`, would allow us to do this:

``` ruby
  def fill_autocomplete(locator, options)
    fill_in locator, options.merge(blur: false)

    click_link options[:with]
  end
```

At my first PoC, I changed the capybara from:

``` ruby
module Capybara
  module Node
    module Actions
      def fill_in(locator, options={})
        raise "Must pass a hash containing 'with'" if not options.is_a?(Hash) or not options.has_key?(:with)
        with = options.delete(:with)
        find(:fillable_field, locator, options).set(with)
      end
    end
  end
end

module Capybara
  module Node
    class Element < Base
[...]

PR #1498, opened 2015-04-08
Use native.clear to clear inputs in Selenium::Node

lencioni:
I am developing a React application where we have some controlled inputs. This means that whenever there is an onchange event, we update some application state that ends up re-rendering the input to the DOM. In practice, when testing these fields with Capybara, we've run into issues with the `fill_in` method because of how the inputs are cleared. Running some JS to set the value to empty ended up not working for us because React would immediately clobber the input with the value stored in the application state.

Using `native.clear` to clear an input before filling it in with the desired text more accurately mimics real-world behavior. This allows us to remove a warning comment and helps Capybara work better with codebases that maintain a tight control over the DOM, such as React.js.

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,7 @@ Release date: unreleased
 ### Added
 
 * 'formmethod' attribute support in RackTest driver [Emilia Andrzejewska]
+* Clear field using backspaces in Selenium driver by using `:fill_options => { :clear => :backspace }` [Joe Lencioni]
 
 #Version 2.4.4
 Release date: 2014-10-13
--- a/lib/capybara/selenium/node.rb
+++ b/lib/capybara/selenium/node.rb
@@ -23,7 +23,7 @@ class Capybara::Selenium::Node < Capybara::Driver::Node
     end
   end
 
-  def set(value)
+  def set(value, fill_options)
     tag_name = self.tag_name
     type = self[:type]
     if (Array === value) && !self[:multiple]
@@ -42,9 +42,17 @@ class Capybara::Selenium::Node < Capybara::Driver::Node
       elsif value.to_s.empty?
         native.clear
       else
-        #script can change a readonly element which user input cannot, so dont execute if readonly
-        driver.browser.execute_script "arguments[0].value = ''", native
-        native.send_keys(value.to_s)
+        if fill_options[:clear] == :backspace
+          # Clear field by sending the correct number of backspace keys.
+          backspaces = [:backspace] * self.value.to_s.length
+          native.send_keys(*(backspaces + [value.to_s]))
+        else
+          # Clear field by JavaScript assignment of the value property.
+          # Script can change a readonly element which user input cannot, so
+          # don't execute if readonly.
+          driver.browser.execute_script "arguments[0].value = ''", native
+          native.send_keys(value.to_s)
+        end
       end
     elsif native.attribute('isContentEditable')
       #ensure we are focused on the element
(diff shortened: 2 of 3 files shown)


lencioni:
I see that this broke some specs, but I don't think I have enough context on how important the previous behavior was. In any case, I'd love to hear your thoughts before I spend more time on this.

twalpole:
This is an area where Capybara has tried many different methods for clearing a field before filling it in and trying to keep it matching normal browser behavior.  native.clear doesnt match normal browser behavior because it fires a change event, which normally wouldn't be fired until an input loses focus, sending the number of backspaces to clear the field was tried but had performance issues in the past (not sure if those still exist or not), sending ctrl-a or cmd-a followed by backspace requires platform specific code, so the javascript clear thats currently implemented was settled on.   That being said, the fill_options parameter was added to #fill_in to allow for passing options through to #set so that drivers could implement things like different clearing mechanisms -   fill_in('Name', with: 'Bob', fill_options: {clear: :native})  or some such thing - although no fill_options have yet been implemented for the selenium driver

lencioni:
Thanks for the response! This is very helpful.

> native.clear doesnt match normal browser behavior because it fires a change event, which normally wouldn't be fired until an input loses focus

Agreed, this is not ideal.

> sending the number of backspaces to clear the field was tried but had performance issues in the past (not sure if those still exist or not)

I had considered this approach as well. I think it might make sense for compatibility to be the default over performance. 

> That being said, the fill_options parameter was added to #fill_in to allow for passing options through to #set so that drivers could implement things like different clearing mechanisms - fill_in('Name', with: 'Bob', fill_options: {clear: :native}) or some such thing - although no fill_options have yet been implemented for the selenium driver

Ah, this sounds good. Would you be open to a pull request that implements `fill_options` for the selenium driver where it clears by backspacing as the default and you can specify an option to clear by assignment for performance? `fill_options: { clear: :assignment }` Alternatively, I could leave the current behavior as default and create an option for backspacing--but as I previously mentioned, it seems like the right way to go is compatibility by default and give people who want to improve performance an option to do so, so that would be my preference.

Also, if you happen to know of a driver that has implemented `fill_options`, that would be helpful for me so I can see an example of how it is being used.

twalpole:
IIRC the performance hit (especially with chrome) was enough to make Capybara basically unusable so we won't be changing the default currently.  Issue #1319 and PR #1338 were where the need for fill_options was discussed and @sobrinho was the person initially requesting for poltergeist.  I'm not sure if it was ever used in poltergeist or not.  I would be open to a pull request that implements multiple clear methods for selenium and potentially a setting that will allow changing the default option (as long as its default is the current behavior)

twalpole:
as @abotalov mentioned it would probably be nice to have a :none option for clear too thereby allowing for appending to existing content

sobrinho:
What I'm currently doing in that cases is this:

``` ruby
  # Public: Fills an masked input using jQuery Input Mask.
  #
  # Examples:
  #
  #   fill_mask 'CPF', with: '000.000.000-00'
  #
  # Returns nothing.
  def fill_mask(locator, with: nil)
    field = find_field(locator)

    page.execute_script <<-JS.strip_heredoc
      $('##{field[:id]}').val('#{with}').trigger('complete')
    JS
  end
```

It uses jQuery and requires the input to have an id but you can get the idea easily.

You can change the triggered event as you like.

sobrinho:
Also note that the method is kind of naive, it is assuming you never have an apostrophe on id (which is invalid anyway) or on value.

But if you concerned about that you can the variables to be something like `$(...).val(#{with.to_s.to_json})`.

lencioni:
Alright, I've pushed a new version that implements `fill_options` with an option to use backspaces to clear the field. It looks like the only specs that are failing are the same ones that are failing on master. Let me know if you'd like me to do anything else with this commit.

shinnn:
@twalpole For future reference, can you update the title of this PR, for example `Clear field using backspaces in Selenium by using { clear: :backspace }` ? This PR finally didn't introduce `native.clear` to Capybara actually.

PR #1501, opened 2015-04-13
Update README.md

cerdiogenes:
Without requiring capybara-webkit you get an error: `no driver called :webkit was found, available drivers: :rack_test, :selenium`

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,7 @@ gem install capybara-webkit
 And you can use it by:
 
 ```ruby
+require 'capybara-webkit'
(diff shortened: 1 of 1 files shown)


twalpole:
If you've added capybara-webkit to your Gemfile then the require shouldn't be required, maybe we should just change it to say to add capybara-webkit to your Gemfile instead of adding the require line.  Any more in depth usage instructions really belong over in the capybara-webkit project rather than in Capybaras readme.

cerdiogenes:
I'm using it in an engine, maybe this is the problem.

Issue #1395, opened 2014-09-10
have_link matcher should throw an error if the options parameter is not a Hash


wgc-as:
I tend to forgot to wrap the URL into a Hash, like this:
```ruby
  it { should have_link 'Some link', some_path }
```
This is what I wanted to write:
```ruby
  it { should have_link 'Some link', href: some_path }
```

This is a false positive because have_link [ignores](https://github.com/jnicklas/capybara/blob/master/lib/capybara/query.rb#L11) the second parameter if it is not a Hash.
I think it would be nice to catch this potential error.

This is my own dirty hack to notify me:

```ruby
module Capybara
  module RSpecMatchers
    alias_method :original_have_link, :have_link

    def have_link(locator, options = {})
      fail 'Second parameter of `have_link` must be a Hash.' unless options.is_a? Hash
      original_have_link locator, options
    end
  end
end
```

PR #1504, opened 2015-04-16
warn when options for matchers aren't a Hash - Issue #1395

twalpole:


abotalov:
Do you think it makes sense to also output a warning for Finders and Actions?

twalpole:
While I want to say "yes it makes sense" we quickly run into issues where lots of warnings are produced for one call -- ie  the actions call finders and each will then issue its own warning, and I think that would start to get confusing.   I guess we could just blank out the options if the object passed in isn't a Hash - which would mean only the initial method receiving the non hash object would warn.  thoughts?

abotalov:
If I would do it I would probably parse `args` in Query#initialize.

If `args` are wrong, it's probably possible to invoke `caller` and find the last element of an array that contains a substring `/capybara/` or something like it. Then to get a method name we can invoke e.g.
```ruby
line[/in `(.+)'/, 1]
```

twalpole:
I'm not 100% sure that would always give us the method the user is calling though which would create confusing warnings.  I'll have to think about it a bit

twalpole:
Thinking about this,  I don't think it would work for chained finds or things like page.find(...), page.has_css?(....)  etc -- I guess we could filter out certain files (session.rb, etc)  but I think it would start to get messy.

abotalov:
It looks like chained methods don't get into trace:
```ruby
class A
  def a
    puts caller
  end

  def b
    self
  end
end
A.new.b.a # `b` method doesn't appear here
```

The only concern I have is a folder `capybara` that might appear in user's / some library's code. To mitigate it we can search for `/lib/capybara/node/finders.rb` and hope that some other library (or a user) won't have it...

So inspecting `caller` feels like a hack :(

twalpole:
hmmm -- interesting -- when I tried it with page.has_link?  the last "capybara" entry in caller was 
"/Users/twalpole/Projects/jnicklas-capybara/lib/capybara/session.rb:686:in `block (2 levels) in <class:Session>'",    -  which I assume came from the page method

twalpole:
@abotalov -  as you can see from the PR - I've implemented something for actions too, but finders gets more complicated due to the variable args functions -- maybe a hack using caller is a cleaner solution.  I'm going to think about it for a bit

PR #1506, opened 2015-04-21
add Capybara.first_default_waiting setting to control default behavior of Node#first

twalpole:
This adds Capybara.first_default_waiting as a setting to allow making the default behavior of Node#first be to wait for at least one element to match (same as passing minimum: 1 as an option), but returning nil if no matching elements show up rather than raising Capybara::ExpectationNotMet

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -22,7 +22,7 @@ module Capybara
     attr_accessor :asset_host, :app_host, :run_server, :default_host, :always_include_port
     attr_accessor :server_port, :exact, :match, :exact_options, :visible_text_only
     attr_accessor :default_selector, :default_wait_time, :ignore_hidden_elements
-    attr_accessor :save_and_open_page_path, :automatic_reload, :raise_server_errors, :server_errors
+    attr_accessor :save_and_open_page_path, :wait_on_first_by_default, :automatic_reload, :raise_server_errors, :server_errors
     attr_writer :default_driver, :current_driver, :javascript_driver, :session_name, :server_host
     attr_accessor :app
 
@@ -48,7 +48,7 @@ module Capybara
     # [ignore_hidden_elements = Boolean]  Whether to ignore hidden elements on the page (Default: true)  
     # [automatic_reload = Boolean]        Whether to automatically reload elements as Capybara is waiting (Default: true)  
     # [save_and_open_page_path = String]  Where to put pages saved through save_and_open_page (Default: Dir.pwd)  
-    #
+    # [wait_on_first_by_default = Boolean]   Whether Node#first defaults to Capybara waiting behavior for at least 1 element to match (Default: false) 
     # === DSL Options
     #
     # when using capybara/dsl, the following options are also available:
@@ -369,6 +369,7 @@ Capybara.configure do |config|
   config.raise_server_errors = true
   config.server_errors = [StandardError]
   config.visible_text_only = false
+  config.wait_on_first_by_default = false
 end
 
 Capybara.register_driver :rack_test do |app|
--- a/lib/capybara/node/finders.rb
+++ b/lib/capybara/node/finders.rb
@@ -166,7 +166,13 @@ module Capybara
       # @return [Capybara::Node::Element]            The found element or nil
       #
       def first(*args)
+        if Capybara.wait_on_first_by_default
+          options = if args.last.is_a?(Hash) then args.pop.dup else {} end
+          args.push({minimum: 1}.merge(options))
+        end
         all(*args).first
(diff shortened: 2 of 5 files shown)


lib/capybara.rb
```diff
@@ -48,7 +48,7 @@ class << self
     # [ignore_hidden_elements = Boolean]  Whether to ignore hidden elements on the page (Default: true)  
     # [automatic_reload = Boolean]        Whether to automatically reload elements as Capybara is waiting (Default: true)  
     # [save_and_open_page_path = String]  Where to put pages saved through save_and_open_page (Default: Dir.pwd)  
-    #
+    # [first_default_waiting = Boolean]   Whether Node#first defaults to Capybara waiting behavior for at least 1 element to match (Default: false)
```
abotalov: IMO it looks like `first` is an adjective in phrase `first default waiting`.

Do you think `enable_waiting_by_first` would be a better name?

twalpole: yeah -- the name I used isn't great -- but I don't think enable_waiting_by_first is great either since it seems to imply #first can't wait without that being true, which it can as long as you pass minimum, count, etc. - I'm open to other suggestions.

abotalov: `enable_default_waiting_for_first`
`make_first_to_wait_by_default`

RSO: +1 for `wait_on_first_by_default`! Seems to be the most descriptive approach.

abotalov: `make_first_to_wait_by_default` isn't good, however, as waiting will not happen if wait is set to 0.

lib/capybara/node/finders.rb
```diff
@@ -166,7 +166,13 @@ def all(*args)
       # @return [Capybara::Node::Element]            The found element or nil
       #
       def first(*args)
+        if Capybara.first_default_waiting
+          options = if args.last.is_a?(Hash) then args.pop.dup else {} end
+          args.push({minimum: 1}.merge(options))
+        end
         all(*args).first
+      rescue Capybara::ExpectationNotMet
+        nil
```
RSO: Am I correct in assuming that this is for backwards compatibility? Or is there some other reason why this method can not raise the `Capybara::ExpectationNotMet`?

I'm afraid that returning `nil` with `wait_on_first_by_default` could still cause confusion.

RSO:
Thanks for this PR! This would indeed resolve my issues (apart from the raised questions around returning `nil`).

lib/capybara/node/finders.rb
abotalov: I think the reason is to not have methods with duplicated behavior:
* `find` with `match: :first` raises an error if there are no elements
* `first` doesn't raise error if there are no elements

twalpole: The definition of #first is to return the first matching element, or nil if none -- I don't think changing that makes sense.

RSO: But doesn't this change existing behaviour? E.g, previously `first('.bla', minimum: 1)` would raise an exception, but now it won't.

twalpole: If passing options It does, but not for the common case of node.first().  the documentation states it returns the element or nil if none exists, so I would say the previous behavior with options is actually a bug. Also without this. Node.first() would raise if Capybara.wait_on_first... Was true, which again doesn't match the docs, and would break many more tests (since most people don't realize their are options that can be passed). What use case do you have where you would want #first to raise?

RSO: To be clear: I can live with the current spec, was just wondering if this PR wouldn't break existing flow.

That being said, I'm not quite sure why `first` should return `nil`. You asked for a use case so here's the line of code that actually raised the original error:

```
 page.first('a', text: link).click
```

The error `undefined method 'click' for NilClass` is in my opinion less useable than `can't find element for selector 'a'`.

twalpole: Ok, you make a good point, and it will be considered for v3 but for now we need to stick with matching current behavior of first() and the documented description of #first, as @abotalov pointed out you can do.  find(:link, link, match: :first).click  if you want a method that will raise

abotalov: Actually, v3 is completely another story.

I agree that it would be better if `first` would raise an error if there are no elements rather than return `nil`.

RSO: Cool! Thanks for the info! Any place in particular where I can read about/help with v3?

twalpole: There are some current issues tagged with a milestone of v3  -- It's basically in the collecting ideas phase, no real plans/timeframe yet.

Issue #1396, opened 2014-09-11
Method "#all" collision between Capybara and Rspec


mhodgson:
Unfortunately the `all` method in the Capybara DSL is now conflicting with the `all` matcher in Rspec (https://www.relishapp.com/rspec/rspec-expectations/v/3-1/docs/built-in-matchers/all-matcher). I'm not sure what happens now (fight to the death?), but it did lead to a very confusing error message:

```
The given selector #<RSpec::Matchers::BuiltIn::BePredicate:0x007fd94e726760> is either invalid or does not result in a WebElement. The following error occurred:
       InvalidSelectorError: An invalid or illegal selector was specified
```

Thoughts?

schnittchen:
I have a feature here where I'm not interested in accessing the Capybara method, but RSpec's builtin matcher instead. Worked around with `include RSpec::Matchers.clone`

mattwynne:
Would it be possible to implement a proxy that's intelligent enough to forward methods called on the result of `#all` to either Capybara or RSpec depending on the context?

PR #1509, opened 2015-04-23
add #find_all as an alias of #all - Issue #1396

twalpole:


abotalov:
It's useful only if RSpec::Matchers is included after Capybara::DSL, though.

twalpole:
@abotalov - Correct, if they are included the other way then Capybaras #all isn't hidden. Are you saying we shouldn't do it because of that? I don't think we want to deprecate #all currently, and I see no issue with having #all and #find_all accessible.

PR #1512, opened 2015-04-25
browsermob-proxy implementation for whitelisting urls

twalpole:
This is nowhere near final and will be changing quite a bit


twalpole:
PuffingBilly is a better solution for now

twalpole:
To clarify - by "better" I mean already implemented and not requiring any other work currently.  puffing-billy is also written in ruby so doesn't require having a java environment setup.  From a feature/flexibility standpoint browsermob-proxy support could probably provide more.

Issue #1510, opened 2015-04-23
have_link's :href option should accept RegEx


YSavir:
I've been trying to use `have_link` and test its `href` attribute using RegEx but to no avail. Here's a sample of the code:

```
scenario { should have_link 'Edit', :href => /\/todos\/\d+\/edit/ }
=> RSpec::Expectations::ExpectationNotMetError: expected to find link "Edit" with href /\/todos\/\d+\/edit/ but there were no matches. Also found "Edit", which matched the selector but not all filters.
```

It'll work with a string, but not with Regex:

```
scenario { should have_link 'Edit', :href => '/todos/7/edit' }
=> true
```

```
scenario { should have_link 'Edit', :href => /#{Regexp.escape('/todos/7/edit')}/ }
=> RSpec::Expectations::ExpectationNotMetError: expected to find link "Edit" with href /\/todos\/7\/edit/ but there were no matches. Also found "Edit", which matched the selector but not all filters.
```

twalpole:
Correct - have_link doesn't accept regex for the :href option, nor is it documented anywhere to. From the docs for click_link which uses the same selector as have_link - 

# @option options [String] :href    The value the href attribute must equal

If instead you're saying you would like to see this functionality expanded to allow for using regexps then fully tested PRs are always considered.

YSavir:
On it, thanks!

Also, was browsing through the tests, and as far as I can tell there isn't any test coverage for the `:href` option to begin with. Is that right?

twalpole:
yes there is in lib/capybara/spec/session/has_link_spec.rb  -- Unlike most projects most of the specs are in the lib/capybara/spec directory because they need to be accessible to all the other drivers that are written for Capybara

PR #1513, opened 2015-04-27
can find an click links with regex href values

YSavir:
Adds regex support when searching for links with `href` filters.

Solves #1510.

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -21,7 +21,7 @@ module Capybara
       #
       # @param [String] locator         text, id, title or nested image's alt attribute
       # @param options
-      # @option options [String] :href    The value the href attribute must equal
+      # @option options [String, Regexp] :href    The value the href attribute must equal
       #
       def click_link(locator, options={})
         find(:link, locator, options).click
--- a/lib/capybara/node/matchers.rb
+++ b/lib/capybara/node/matchers.rb
@@ -220,7 +220,7 @@ module Capybara
       #
       # @param [String] locator           The text or id of a link to check for
       # @param options
-      # @option options [String] :href    The value the href attribute must be
+      # @option options [String, Regexp] :href    The value the href attribute must be
       # @return [Boolean]                 Whether it exists
       #
       def has_link?(locator, options={})
--- a/lib/capybara/selector.rb
+++ b/lib/capybara/selector.rb
@@ -149,8 +149,12 @@ end
 
 Capybara.add_selector(:link) do
   xpath { |locator| XPath::HTML.link(locator) }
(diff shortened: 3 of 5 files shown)


lib/capybara/query.rb
```diff
@@ -43,6 +43,10 @@ def matches_filters?(node)
         regexp = options[:text].is_a?(Regexp) ? options[:text] : Regexp.escape(options[:text].to_s)
         return false if not node.text(visible).match(regexp)
       end
+      if options[:href]
```
twalpole: This section is for filters that are common to all queries/selectors  -- href is not so it shouldn't be here, it should be in the relevant selectors

lib/capybara/selector.rb
```diff
@@ -1,3 +1,5 @@
+require 'pry'
```
twalpole: non-necessary change

lib/capybara/query.rb
YSavir: Will change.

lib/capybara/selector.rb
YSavir: Woops. Missed that in the `git diff` :/.

twalpole:
@YSavir Other than the few line comments I made, its looking good -- please also update the has_link and click_link documentation to show they can take a regex for :href

lib/capybara/query.rb
YSavir: Turns out that the changes I made to the `link` selector made this filter unnecessary anyway.

YSavir:
Added changes. Should I squash the commits?

lib/capybara/node/actions.rb
```diff
@@ -21,7 +21,7 @@ def click_link_or_button(locator, options={})
       #
       # @param [String] locator         text, id, title or nested image's alt attribute
       # @param options
-      # @option options [String] :href    The value the href attribute must equal
+      # @option options [String] :href    The value the href attribute must equal. May be either a string or regular expression.
```
twalpole: This should be [String, Regexp]

lib/capybara/node/matchers.rb
```diff
@@ -220,7 +220,7 @@ def has_no_css?(path, options={})
       #
       # @param [String] locator           The text or id of a link to check for
       # @param options
-      # @option options [String] :href    The value the href attribute must be
+      # @option options [String] :href    The value the href attribute must be. May be either a string or regular expression.
```
twalpole: [String, Regexp]

twalpole:
Once the documentation fixes are made, yes please squash

YSavir:
Done and done.

twalpole:
Thanks

PR #1516, opened 2015-05-04
inject_asset_host should not raise if <head> is missing

tjgrathwell:
Previously, if you had Capybara.asset_host defined and tried to
`save_page` on a document without a `<head>` tag (like plain text)
`inject_asset_host` would crash trying to string-replace a `<base>`
tag onto the nonexistant `<head>`.

Since there's probably no need for it to successfully inject a
`<base>` into such documents, now it will return the original content
when there is no `<head>` present.

### DIFF
--- a/lib/capybara/helpers.rb
+++ b/lib/capybara/helpers.rb
@@ -42,10 +42,12 @@ module Capybara
     def inject_asset_host(html)
       if Capybara.asset_host && Nokogiri::HTML(html).css("base").empty?
         match = html.match(/<head[^<]*?>/)
-        html.clone.insert match.end(0), "<base href='#{Capybara.asset_host}' />"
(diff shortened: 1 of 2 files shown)


twalpole:
This seems good for plain text, but what if it is actually an html document, just without a <element> - I think in that case we should probably add a <head> containing the <base>

tjgrathwell:
I think the main use case for `Capybara.asset_host` is allowing the generated .html file to successfully find .css files relevant to it.

In most reasonable HTML documents, those are in the `<head>`, so I don't know if the extra complexity required to also add `<base>` to the `<body>` (or wherever) would be merited.

twalpole:
ok -- we'll go with this for now -- thanks

PR #1518, opened 2015-05-13
suppress FileNotFound exception in remote test, new config parameter

lazylester:
capybara may not be running on the same computer as the browser.

I'm testing Internet Explorer on a windows machine running selenium-server. Driving the test from RSpec/capybara running on my Mac. The (Rails) app is also running on the Mac.

However when testing the uploading files using attach_file, capybara raises an exception b/c the file's not present on my Mac. But it's not relevant whether the file's on the Mac, but whether it's present on the windows machine. 

Of course capybara can't check whether the file is present on the windows machine, but at least it should allow the test to continue, and then the expectation should fail. So I added a config parameter (config.remote = true) to inhibit the exception.

I'm kinda surprised nobody's run into this before, so I'm wondering if this is a weird test configuration or if there's some other solution that I'm not seeing. Or maybe people don't test with IE? Anyway, here's a PR with tests for you. To run just the tests relevant to this feature, use the RSpec tag :attach_file => true.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -24,7 +24,7 @@ module Capybara
     attr_accessor :default_selector, :default_wait_time, :ignore_hidden_elements
     attr_accessor :save_and_open_page_path, :wait_on_first_by_default, :automatic_reload, :raise_server_errors, :server_errors
     attr_writer :default_driver, :current_driver, :javascript_driver, :session_name, :server_host
-    attr_accessor :app
+    attr_accessor :app, :remote
 
     ##
     #
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -142,6 +142,8 @@ module Capybara
       #
       # Find a file field on the page and attach a file given its path. The file field can
       # be found via its name, id or label text.
+      # Exception is raised if file does not exist. This can be suppressed for remote testing scenarios
+      # with the config option config.remote => true
       #
       #     page.attach_file(locator, '/path/to/file.png')
       #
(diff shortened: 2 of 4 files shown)


twalpole:
I think a lack of remote testing and a minimal amount of use of attach_file in testing due to its shortcomings (no multiple files, etc - caused by shortcomings in the selenium api) lead to this not being an issue for most people.  I'm wondering whether it may be better to just not check for existence - or issue a warning rather than raising - instead of adding another configuration switch

lazylester:
@twalpole either of your suggestions is ok. If the file's not present then the test should surely fail anyway, my personal preference would be to eliminate the existence check. Do you need another PR?
Thank you for taking the time to respond, and thanks for your work on Capybara.

twalpole:
@lazytester - looking at this some more and doing a bit of reading of the docs for selenium-webdriver -  I think you should actually be keeping your test files on the local machine, and setting the file detector on the remote driver.  By doing that selenium-webdriver is supposed to find the local file and upload it to the remote test machine - if that does actually work then Capybara checking for file presence makes sense.

lazylester:
@twalpole thanks for your response...

"if that does actually work" you say. I'd be happy to verify it if you can point me to the selenium-webdriver docs that describe it. After reading your suggestion, I reviewed the docs and haven't seen anything like that described. Nor do I see any API that suggests that this is possible.

All we really have is the Capybara attach_file interface. The same test should be executable whether interacting with a browser on the local machine or a remote machine. It seems pretty weird to have some kind of conditional inside the test that invokes some driver API calls if the target is remote. Isn't that what your'e suggesting?

twalpole:
Its setup of the driver, not conditional inside the test.   Here is a link to the documentation on it - http://www.rubydoc.info/gems/selenium-webdriver/Selenium/WebDriver/DriverExtensions/UploadsFiles#file_detector%3D-instance_method

Issue #1531, opened 2015-06-05
Select action doesn't respect disabled flag


Vanuan:
Given this:

```
<select name="my-select">
  <option value="0" disabled="disabled">Disabled option</option>
  <option value="1">Enabled option</option>
</select>
```

When I do this:

```
select('Disabled option', :from => 'my-select')
```

Everything is fine, but I expect an exception to be thrown.

twalpole:
We can't throw an exception currently (breaking behavior change) - but a warning does seem reasonable (like happens when attempting to set read only fields).  See PR #1532

Vanuan:
At least it should not be selected. Thank you!

PR #1532, opened 2015-06-05
warn when attempting to select a disabled option - Issue #1531

twalpole:


Vanuan:
The use case is this:
I have a cucumber test that is working fine. Than I add a "disabled" option to some items dynamically. The test still passes, while a user couldn't select an option.

twalpole:
@Vanuan What driver are you using?

Vanuan:
It's phantomjs

Vanuan:
Posted there https://github.com/ariya/phantomjs/issues/13303

Vanuan:
Oh, no it's poltergeist

twalpole:
@Vanuan but is it actually selecting the option, or just ignoring the select --- I would hope it's just ignoring the select.  Capybara will not be throwing an exception for this behavior currently (just issuing a warning - which will also issue for poltergeist), so I would hope poltergeist throw since that would mean different drivers with different behaviors.   If you want your test to fail when trying to select a disabled element then you will need to assert that the value was or was not actually selected.

PR #1536, opened 2015-06-16
Add a :label selector and support for "dynamic" selectors

twalpole:
Add a label selector that will take  text, "for" identifer, or the element whose bound label should be found.  It also adds the idea of "dynamic" selectors where the selector receives the locator and the node its being called on and then performs its own finding of nodes instead of just providing a css or xpath expression.  I'm not fully sold on the name "dynamic" yet but haven't come up with anything better


twalpole:
"dynamic" selector is not required for label selector - can be implemented with a filter - closing

PR #1539, opened 2015-06-17
Escape url in `current_url`.

yiliangt5:
`URL.parse` in `current_url` will raise an error if url contains any
illegal URI characters.


twalpole:
What problem is this supposed to fix?  - and needs tests

PR #1542, opened 2015-07-01
Update matchers documentation

teoljungberg:
- Prefer `check that`, `check if` reads as if it's vague.
- Add `the` to `current node` for specificity.


twalpole:
If is actually grammatically correct for most of these cases -  http://thegrammarexchange.infopop.cc/eve/forums/a/tpc/f/340600179/m/9821045711

Issue #1540, opened 2015-06-27
#raise_server_error! doesn't raise an exception when multiple sessions are created


yuki24:
Let's say multiple sessions are created while running the test suite and the app raises an exception in a certain situation. Each session makes a request to the same end point and the app will raise an exception. In this scenario, the first session can capture the exception, but the second one can't.

This is because [each session creates a server object](https://github.com/jnicklas/capybara/blob/5b6d65feaef0bbed661db607f9970a90730e1d2b/lib/capybara/session.rb#L66), but [only the first one creates a server thread](https://github.com/jnicklas/capybara/blob/5b6d65feaef0bbed661db607f9970a90730e1d2b/lib/capybara/server.rb#L66). And since the second session talks to the server thread created by the first session, the exception will be rescued by the first server object (more sepcifically, [a middleware object](https://github.com/jnicklas/capybara/blob/5b6d65feaef0bbed661db607f9970a90730e1d2b/lib/capybara/server.rb#L20-L21)).

The following test replicates the issue:

```ruby
# bug_report.rb
require 'capybara'
require 'selenium-webdriver'
require 'minitest/autorun'

class BugTest < Minitest::Test
  Capybara.current_driver = :selenium
  Capybara.app = ->(env) {
    raise "RuntimeError from the app"
  }

  def test_raise_server_error!
    session = Capybara.current_session
    session.visit("/")

    assert_raises RuntimeError, "RuntimeError from the app" do
      session.raise_server_error!
    end

    Capybara.using_session :another_session do
      session = Capybara.current_session
      session.visit("/")

      assert_raises RuntimeError, "RuntimeError from the app" do
        session.raise_server_error!
      end
    end
  end
end
```

And the result:

```
$ ruby bug_report.rb 
Run options: --seed 40314

# Running:

F

Finished in 0.752366s, 1.3291 runs/s, 2.6583 assertions/s.

  1) Failure:
BugTest#test_raise_server_error! [bug_report.rb:23]:
RuntimeError from the app.
RuntimeError expected but nothing was raised.

1 runs, 2 assertions, 1 failures, 0 errors, 0 skips
```

twalpole:
Yes - this is an issue -- not sure what the best solution is though, or whether the current behavior is good enough.  I don't think starting a new session to run a separate test on the same app is a normal use case.  Running a new copy of the app for each session is not a good solution either.

twalpole:
[...]

PR #1545, opened 2015-07-07
Wait for requests to complete on the server in Session#reset!

jhawthorn:
It's possible when using a JS driver for a request to continue being processed on the server after calling `Capybara.reset_sessions!`. Though the browser requests are stopped, any requests the server had already started processing will continue.

This causes issues when using capybara with DatabaseCleaner, as the the cleaner might run at the same time as the request, which will likely cause the request to error.

This PR adds a thread-safe counter to Capybara::Server which keeps track of the number of requests currently being processed by `Capybara::Server::Middleware`. It exposes public methods `Server#pending_requests?`, which is truthy when there are requests still being processed, and `Server#wait_for_pending_requests`, which will wait up to 60 seconds for requests to finish.

### DIFF
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -89,7 +89,7 @@ module Capybara
     end
 
     def wait_for_pending_requests
-      Timeout.timeout(60) { @server_thread.join(0.01) while pending_requests? }
+      Timeout.timeout(60) { sleep(0.01) while pending_requests? }
     rescue Timeout::Error
       raise "Requests did not finish in 60 seconds"
     end
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -4,22 +4,43 @@ require 'rack'
 
 module Capybara
   class Server
+    class Counter
+      attr_reader :value
+
+      def initialize(value=0)
+        @value = value
+        @mutex = Mutex.new
+      end
+
+      def increment
+        @mutex.synchronize { @value += 1 }
+      end
+
+      def decrement
+        @mutex.synchronize { @value -= 1 }
+      end
+    end
+
     class Middleware
       attr_accessor :error
 
-      def initialize(app)
+      def initialize(app, counter=Counter.new)
         @app = app
+        @counter = counter
       end
 
       def call(env)
         if env["PATH_INFO"] == "/__identify__"
           [200, {}, [@app.object_id.to_s]]
         else
+          @counter.increment
           begin
             @app.call(env)
           rescue *Capybara.server_errors => e
             @error = e unless @error
             raise e
+          ensure
+            @counter.decrement
           end
         end
       end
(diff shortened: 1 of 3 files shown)


twalpole:
Why are you attempting to join the server thread while waiting for the requests to end?  Joining would only succeed if the server thread was ending.  I also need to think about how/if this relates to testing with multiple sessions.

jhawthorn:
I used `join` just because it's what `boot` was using. I will change it to a sleep.

twalpole:
This is definitely still on the table and I'm considering it for 2.6 -- I haven't spent enough time to think through whether it has any issues with the fact that multiple sessions often share the same Capybara::Server and Middleware

jhawthorn:
@twalpole thanks for the update. Is there anything I could do to help move this forward?

lib/capybara/server.rb
```diff
@@ -4,22 +4,43 @@
 
 module Capybara
   class Server
+    class Counter
+      attr_reader :value
```
twalpole: We don't really need the value we just need a zero? method

lib/capybara/server.rb
```diff
@@ -4,22 +4,43 @@
...
+      def decrement
+        @mutex.synchronize { @value -= 1 }
+      end
+    end
+
     class Middleware
       attr_accessor :error
 
-      def initialize(app)
+      def initialize(app, counter=Counter.new)
```
twalpole: do we really need to accept a counter here? since it's always passed a new counter in usage couldnt we just create a new counter inside initialize

lib/capybara/server.rb
```diff
@@ -35,7 +56,8 @@ def ports
 
     def initialize(app, port=Capybara.server_port, host=Capybara.server_host)
       @app = app
-      @middleware = Middleware.new(@app)
+      @counter = Counter.new
```
twalpole: see above

lib/capybara/server.rb
```diff
@@ -62,6 +84,16 @@ def responsive?
       return false
     end
 
+    def pending_requests?
+      @counter.value > 0
```
twalpole: see comment about value in Counter class

lib/capybara/session.rb
```diff
@@ -111,6 +112,14 @@ def reset!
 
     ##
     #
+    # Wait for any pending server requests to finish
+    #
+    def wait_for_server_requests
```
twalpole: Does this need to part of the public API?

twalpole:
@jhawthorn ok -- I've added a few comments to the current code, but there is a larger concern.  I believe this suffers from the same problem as Issue #1540 .  Namely that the methods on session don't really do what one would think when multiple sessions using the same app are created.  This may or may not be of concern I haven't really decided if it's actually a real problem or just an issue that occurs in artificial conditions.  I think it would actually be more of an issue for this situation than the #1540 though.  I don't think we want to be duplicating the app object for each session, but a quick thought at a fix would be to have each session create a new server thread with it's own middleware instance on it's own port but share the app object, which would allow tracking requests on a per session basis.  That could lead to threading problems though and probably requires more thought

twalpole:
The per_session_server branch may be required to have this code work with multiple sessions correctly, more comments about it over on Issue #1540

jhawthorn:
@twalpole sorry for the delay.

I've addressed your feedback and rebased against the latest master.

As far as I can tell this works fine with the changes in the `per_session_server` branch (different servers would count their requests separately). The code is very similar to that which handles errors, so I would expect them to behave largely the same.

twalpole:
Closing this PR as it was merged into PR #1637

PR #1550, opened 2015-07-20
Update README.md

GBH:
it's `default_wait_time`

### DIFF
--- a/README.md
+++ b/README.md
@@ -723,7 +723,7 @@ looking for that content for a brief time. You can adjust how long this period
 is (the default is 2 seconds):
 
 ```ruby
-Capybara.default_max_wait_time = 5
+Capybara.default_wait_time = 5
 ```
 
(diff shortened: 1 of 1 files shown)


abotalov:
This is README.md of `master` branch. It's not released yet.
See https://github.com/jnicklas/capybara/pull/1502

twalpole:
The readme is for the upcoming version 2.5 -- where it has changed to default_max_wait_time to attempt to clear up confusion about what exactly the setting does.  default_wait_time is being deprecated.

GBH:
Ah, bah! That makes sense. 

Problem is that one wouldn't know. Even the site: http://jnicklas.github.io/capybara/ tells you to install the gem and points to README on the master branch.

I'm seen on some other repos notices like "This is README for unreleased version, click here for current". This comes up from time to time.

PR #1552, opened 2015-07-21
Ruby warnings

amatsuda:
This PR eliminates some Ruby :warning:s when run with `RUBYOPT="-w"`.

Similar to #599, #897.


twalpole:
thanks

PR #1556, opened 2015-07-29
Update readme with correct method name #default_wait_time

mattduboismatt:


twalpole:
The README is for master where the correct method name is default_max_wait_time

abotalov:
Dup of https://github.com/jnicklas/capybara/pull/1550

PR #1558, opened 2015-08-09
Implement Capybara::Selenium::Node#path

soutaro:
This PR is to implement`Capybara::Selenium::Node#path`. It now returns a XPath which points to the node, instead of raising `NotSupportedByDriverError`.

### DIFF
--- a/lib/capybara/selenium/node.rb
+++ b/lib/capybara/selenium/node.rb
@@ -133,6 +133,34 @@ class Capybara::Selenium::Node < Capybara::Driver::Node
     native == other.native
   end
 
+  def path
+    path = parents
+    path.unshift self
+
+    result = []
+    while node = path.shift
+      parent = path.first
+
+      if parent
+        siblings = parent.find_xpath(node.tag_name)
+        if siblings.size == 1
+          result.unshift node.tag_name
+        else
+          index = siblings.index(node)
+          result.unshift "#{node.tag_name}[#{index+1}]"
+        end
+      else
+        result.unshift node.tag_name
+      end
+    end
+
+    '/' + result.join('/')
+  end
+
+  def parents
+    find_xpath('ancestor::*').reverse
+  end
+
 private
 
   # a reference to the select node if this is an option node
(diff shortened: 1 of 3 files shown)


lib/capybara/selenium/node.rb
```diff
@@ -133,6 +133,34 @@ def ==(other)
...
+        end
+      else
+        result.unshift node.tag_name
+      end
+    end
+
+    '/' + result.join('/')
+  end
+
+  def parents
```
twalpole: shouldn't be public

spec/selenium_spec.rb
```diff
@@ -100,6 +100,22 @@ module TestSessions
         expect(@session).to have_selector(:css, '.change_event_triggered', :match => :one)
       end
     end
+
+    describe "#path" do
+      before :each do
+        @session.visit('/with_js')
```
twalpole: does this need JS in anyway? if not it shouldn't use the /with_js test page -- also since #page is part of the node api this should be moved to the specs used by all drivers rather than just the selenium driver

spec/selenium_spec.rb
```diff
@@ -100,6 +100,22 @@ module TestSessions
...
+      end
+
+      it "returns xpath" do
+        element = @session.find(:css, '#with_focus_event')
+        expect(element.path).to eq('/html/body/p[5]/input')
```
twalpole: Not every driver may produce exactly the same xpath - so its probably better to get the #path result, find the element using that xpath and then make sure they're the same element

twalpole:
Thanks for the PR -- I made a couple of comments, with those fixes I think it looks good --- I'll give it a more in depth review later today

spec/selenium_spec.rb
soutaro: Does not need JS. Should I add something like xpath.erb?

spec/selenium_spec.rb
soutaro: I'm not very sure but the spec looks a spec of Selenium driver. Do you mean I should not write XPath here anyway?

soutaro:
Thanks! I will fix for your comments today.

PR #1565, opened 2015-08-20
Update name of wait time config variable

ramhoj:
`default_max_wait_time` doesn't seem to exist any more. I'm guessing `default_wait_time` is the new name?


twalpole:
Dup of #1550

PR #1566, opened 2015-08-21
Update history to prepare for 2.5 release

twalpole:


twalpole:
@jferris @jonleighton @mhoran  @route  It's been a while since the last Capybara release and some of the features I wanted to add have been delayed (proxy support for selenium driver).  My current plan is to release the current set of updates as 2.5 in the next few days/week, and then do a 2.6 with the other planned updates in the next couple of months.   I don't think there is much in 2.5 that will affect your drivers other than Node#send_keys and that there is now a test for Node#path behavior.  If you have any concerns with the 2.5 release please let me know

mhoran:
Thanks for the heads up. We've not yet implemented `send_keys` in capybara-webkit, though I do have a work in progress. Also, we'd like to cut our own version 2.0 to drop some deprecated features, so it probably makes sense to coordinate the release.

History.md
```diff
@@ -1,8 +1,26 @@
 # Version 2.5.0
 Release date: unreleased
 
-### Added
+###Fixed
+* Error message now raised correctly when invalid options passwed to 'have_text'/'have_content' [Tom Walpole]
```
abotalov: "passed"

twalpole:
@mhoran I'm pretty much ready to do a 2.5 release, would it be of any benefit to do an rc1 release first?

mhoran:
I'd say go ahead and release 2.5. I think the only issue we have at the moment is `send_keys`, which shouldn't be too difficult as we've already looked at it. Unfortunately I've yet to finish up the implementation, but just about everything else looks good. Hopefully we can get a compatible version out the door shortly after the release, but I don't want to hold up the release in the meantime.

jferris:
We can push an interim release which allows Capybara 2.5 but raises an unsupported error for `send_keys`. That's the usual thing to do when a driver doesn't support a feature, right?

twalpole:
Yeah, and as long as you're deriving from Capybara::Driver::Node that should happen automatically

route:
Thank you @twalpole for information, I'll try to run tests and release new version soon!

PR #1570, opened 2015-08-27
Visit using merge

twalpole:
This reimplement Session#visit usuing URI#merge.  It was initially aimed at fixing Issue #1348, but I decided that #1348 wasn't really a valid use of #visit.

### DIFF
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -206,30 +206,30 @@ module Capybara
     #
     # @param [#to_s] url     The URL to navigate to. The parameter will be cast to a String.
     #
-    def visit(url)
+    def visit(visit_uri)
       raise_server_error!
-
-      url = url.to_s
       @touched = true
 
-      url_relative = URI.parse(url).scheme.nil?
+      visit_uri = URI.parse(visit_uri.to_s)
 
-      if url_relative && Capybara.app_host
-        url = Capybara.app_host + url
-        url_relative = false
+      uri_base = if @server
+        visit_uri.port = @server.port if Capybara.always_include_port && (visit_uri.port == visit_uri.default_port)
+        URI.parse(Capybara.app_host || "http://#{@server.host}:#{@server.port}")
+      else
+        Capybara.app_host && URI.parse(Capybara.app_host)
       end
 
-      if @server
-        url = "http://#{@server.host}:#{@server.port}" + url if url_relative
-
-        if Capybara.always_include_port
-          uri = URI.parse(url)
-          uri.port = @server.port if uri.port == uri.default_port
-          url = uri.to_s
-        end
+      # TODO - this is only for compatability with previous 2.x behavior that concatenated
+      # Capybara.app_host and a "relative" path - Consider removing in 3.0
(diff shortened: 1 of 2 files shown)


lib/capybara/spec/session/visit_spec.rb
```diff
@@ -17,6 +17,17 @@
...
   end
 
+  # TODO: This test passes on racktest and selenium - although I think all it's really testing
+  # is that scheme defaults to http if not specified as anything else
+  # it "should fetch a response when absolute URI doesn't have a scheme", focus: true do
+  #   # Preparation
+  #   @session.visit('/')
+  #   root_uri = URI.parse(@session.current_url)
+  #
+  #   @session.visit("//#{root_uri.host}:#{root_uri.port}/foo")
```
abotalov: Do you know a use case for this feature?
@dstosik wanted to use it to load relative page links but that's not what `visit` is intended to do.

lib/capybara/spec/session/visit_spec.rb
```diff
@@ -17,6 +17,17 @@
     expect(@session).to have_content('Another World')
   end
 
+  # TODO: This test passes on racktest and selenium - although I think all it's really testing
+  # is that scheme defaults to http if not specified as anything else
```
abotalov: I think scheme defaults to the scheme of `app_host`

lib/capybara/session.rb
```diff
@@ -206,30 +206,28 @@ def current_url
...
 
-      if @server
-        url = "http://#{@server.host}:#{@server.port}" + url if url_relative
-
-        if Capybara.always_include_port
-          uri = URI.parse(url)
-          uri.port = @server.port if uri.port == uri.default_port
-          url = uri.to_s
-        end
+      # TODO - this is only for compatability with previous 2.x behavior that concatenated
```
abotalov: I think this might be useful for apps deployed to some path instead of a root.

Also it might be useful for single page apps where fragment changes depending on the page (so user can set app_host to `http://www.example.com/app` and then do `visit '#inbox'`, `visit '#starred'`)

lib/capybara/spec/session/visit_spec.rb
twalpole: Exactly -- thats not what visit does, but really all that test is checking is that scheme defaults to 'http' -- which it does because Capybara.app_host or the server uri in visit is set to a scheme of http.  If app_host was not set and no server was started then it would fail if a scheme wasn't specified - or potentially default to a scheme of :file  (as per Jonas' comment on @dstosik issue)

lib/capybara/spec/session/visit_spec.rb
twalpole: Yep - as I said in the other comment - if app_host is set it'll default to it's scheme, if not it'll default to http since thats set for the server url in #visit, and if not then it'll just end up defaulting to whatever the browser defaults to (probably file:)  --  I should probably just delete this test

lib/capybara/session.rb
twalpole: Good point, there may be a use case for keeping the current behavior.

PR #1574, opened 2015-09-18
Support tristate filter in selector and use for disabled

twalpole:


twalpole:
Closed in favor of #1575

PR #1575, opened 2015-09-19
Changing :disabled option to be tristate

jillianrosile:
* Allowing `true`, `false`, or `:all` for all selectors with the `:disabled` option
* Adding `:skip_if` option to `Filter` to skip the filter (i.e. return `true`) if this value is passed

I added the `:skip_if` option to `Filter` because so many selectors had this `:disabled` option and adding the logic to each seemed too repetitive. If there's a preferred alternate approach or a better name than `:skip_if`, let me know.

### DIFF
--- a/lib/capybara/node/finders.rb
+++ b/lib/capybara/node/finders.rb
@@ -56,7 +56,10 @@ module Capybara
       #
       # @option options [Boolean] checked  Match checked field?
       # @option options [Boolean] unchecked   Match unchecked field?
-      # @option options [Boolean] disabled (false)  Match disabled field?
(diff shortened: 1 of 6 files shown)


jillianrosile:
And right after I submit this I notice that @twalpole submitted the same thing earlier today. Oops. Well, here's another idea of how to approach it.

twalpole:
@jillianrosile Thanks for this -- great idea on the :skip_if option.  I made one change so the skip_if value doesn't need to be specified in the valid_values - merged in to master

jillianrosile:
Thanks! Looked at your follow-up commit and I like it. Makes sense to not have to repeat the parameter in both `:valid_values` and `:skip_if`.

PR #1578, opened 2015-10-02
Rack 2: Inline the version of normalize_params that used to exist in Rack::Utils

tjgrathwell:
The explicit dependency on this Rack method breaks when upgrading to Rack 2 because Rack::Utils.normalize_params has been moved since this commit https://github.com/rack/rack/commit/7e7a3890449b5cf5b86929c79373506e5f1909fb

There's [a very similar version in Rack::QueryParser](https://github.com/rack/rack/blob/master/lib/rack/query_parser.rb#L78), but it has
the downside of trying to make a hash of Rack::QueryParser instances
instead of a hash of hashes.

### DIFF
--- a/lib/capybara/rack_test/form.rb
+++ b/lib/capybara/rack_test/form.rb
@@ -23,7 +23,7 @@ class Capybara::RackTest::Form < Capybara::RackTest::Node
       xpath=xpath.union(x.anywhere(*form_element_types).where(x.attr(:form) == native[:id])) if native[:id]
       xpath.where(~x.attr(:disabled))
     end.to_s
-    
+
     native.xpath(form_elements_xpath).map do |field|
       case field.name
       when 'input'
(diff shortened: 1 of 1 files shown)


twalpole:
hmmm -- I don't really like the idea of copying the whole method over, and would prefer if it was possible to do using functionality provided by rack 2 - however I understand that may not be possible.  I need to think about this for a bit

thejamespinto:
:+1: Thank you @tjgrathwell this bug has been here for months.

I decided to search and found your PR because I wanted to put my [WIP Rails Authenticator](https://travis-ci.org/hello-gem/hello) back testing with rails:master Now my specs work again :heart:  https://travis-ci.org/hello-gem/hello/builds/85068627

I had also tried your first approach, @twalpole I can validate that it does not work because it brings instances of internal classes, maybe a cleaner solution will show up in the recent future, but right now reverting the old code seems like the easiest choice. :cry: 

@tjgrathwell thanks again! :+1:

twalpole:
This is merged into master and I added rack 2 and sinatra 2.2.0-alpha to the beta-versions test for travis

PR #1582, opened 2015-10-06
Adjust has_select filters to allow invisible options in some cases

### LINKED ISSUE #1579: have_select with filters does not respect :visible option
`page.have_select('Label', visible: false, selected: 'selected option')`

The select will be found (even though it is hidden), but the selected option will not be found because it (and its text) are hidden.

I would have expected the filter to match even if the option is hidden (if `visible: false` is specified).

Fixing it seems like it would require a change to the filter system in general.

 As far as I can tell, the filter system itself does not pass the underlying options hash (containing `visible: false`) to the filters when checking for a match, so there is no way for the filter to make determinations based on `visible: false`.

Questions:

1. Would you agree that the filter _should_ match on a hidden option, provided `visible: false` is 
specified?
2. Would you be open to a PR to fix this?
3. Passing the underlying options hash to `Filter#matches?` seems like one viable solution, although perhaps heavy-handed.  Any alternative suggestions?

naw:
This fixes #1579

- selected filter always allows invisible options

- options filter allows invisible options if the select
  itself is invisible

- with_options filter allows invisible options if the select
  itself is invisible

### DIFF
--- a/lib/capybara/selector.rb
+++ b/lib/capybara/selector.rb
@@ -226,12 +226,22 @@ Capybara.add_selector(:select) do
   label "select box"
   xpath { |locator| XPath::HTML.select(locator) }
   filter(:options) do |node, options|
-    actual = node.all(:xpath, './/option').map { |option| option.text }
+    if node.visible?
+      actual = node.all(:xpath, './/option').map { |option| option.text }
+    else
+      actual = node.all(:xpath, './/option', visible: false).map { |option| option.text(:all) }
+    end
     options.sort == actual.sort
   end
-  filter(:with_options) { |node, options| options.all? { |option| node.first(:option, option, minimum: 0) } }
+  filter(:with_options) do |node, options|
+    finder_settings = { minimum: 0 }
+    if !node.visible?
+      finder_settings[:visible] = false
+    end
+    options.all? { |option| node.first(:option, option, finder_settings) }
+  end
   filter(:selected) do |node, selected|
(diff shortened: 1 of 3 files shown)


twalpole:
I think it needs a test for a visible select with a non-visible but selected option - otherwise looks good to me

naw:
My original HTML for the selected spec was actually like that, but then I noticed that selenium considers an invisible selected option as visible.  That makes sense --- if you load a form with a pre-selected invisible option into Firefox or Chrome, the option is indeed visible.

Originally, I had a test to verify a visible select with an invisible option would _not_ pass the options filter, but the selenium version of the test failed because selenium treated the invisible option as if it were visible. The RackTest and Chrome version of that test passed.

I can definitely adjust the spec per your suggestion, but in selenium I believe the spec would pass even without the code changes .... in other words, it would pass for the wrong reason(s) and not be testing the correct code path.  Perhaps that's fine, as long as the other drivers _do_ test the correct code path?

Any additional thoughts on this before I move forward?

twalpole:
We're saying that a selected option is checked against selected whether it's visible or not, so if firefox happens to ignore display: none on an option that is selected I think it's fine -- other drivers may not and the test will be verifying that the option is included in the selected filter

naw:
Added a commit to address your suggestion.  I can squash it if it looks OK to you.

twalpole:
looks good -- go ahead and squash and I'll merge it in

naw:
Squashed. Thanks for your quick attention to this.

twalpole:
thanks for the work!

PR #1595, opened 2015-11-02
Improve failure message when text is present but invisible.

alexch:
When testing a page with JS that hides and shows 
certain elements, it's useful to know whether a 
test failure means that an element has failed to 
render, or has rendered but become invisible (or 
failed to become visible). 

For instance, some jQuery plugins will take 
existing elements and hide them, adding some other
elements onto the page that use the originals as
storage. If your tests are written to use the 
now-hidden elements they will mysteriously start
failing. 

This commit only changes the behavior for failed
tests (rspec and assert) and only for text
queries. It might be nice to extend this to the 
other finders.

### DIFF
--- a/lib/capybara/queries/text_query.rb
+++ b/lib/capybara/queries/text_query.rb
@@ -14,7 +14,8 @@ module Capybara
       end
 
       def resolve_for(node)
-        @actual_text = Capybara::Helpers.normalize_whitespace(node.text(@type))
+        @node = node
+        @actual_text = text(node, @type)
         @count = @actual_text.scan(@search_regexp).size
       end
 
@@ -31,6 +32,16 @@ module Capybara
           message << " but found #{@count} #{Capybara::Helpers.declension('time', 'times', @count)}"
         end
         message << " in #{@actual_text.inspect}"
+
+        if @node and visible?
+          invisible_text = text(@node, :all)
+          invisible_count = invisible_text.scan(@search_regexp).size - @count
+          if invisible_count > 0
+            message << ". (However, it was found #{invisible_count} time#{'s' if invisible_count != 1} in invisible text.)"
+          end
+        end
+
+        message
       end
 
       def negative_failure_message
@@ -42,6 +53,15 @@ module Capybara
       def valid_keys
         COUNT_KEYS + [:wait]
       end
+
+      def visible?
+        @type == :visible or (@type.nil? and Capybara.ignore_hidden_elements)
+      end
+
+      def text(node, query_type)
+        Capybara::Helpers.normalize_whitespace(node.text(query_type))
+      end
+
     end
   end
 end
--- a/lib/capybara/queries/text_query.rb
+++ b/lib/capybara/queries/text_query.rb
@@ -19,7 +19,7 @@ module Capybara
         @count = @actual_text.scan(@search_regexp).size
       end
 
-      def failure_message
+      def failure_message(check_invisible = true)
         description =
           if @expected_text.is_a?(Regexp)
             "text matching #{@expected_text.inspect}"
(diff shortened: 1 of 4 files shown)


harlantwood:
+1

lib/capybara/queries/text_query.rb
```diff
@@ -31,6 +32,16 @@ def failure_message
           message << " but found #{@count} #{Capybara::Helpers.declension('time', 'times', @count)}"
         end
         message << " in #{@actual_text.inspect}"
+
+        if @node and visible?
```
twalpole: I don't think this message doesn't really make sense to appear in the negative failure message error

twalpole:
I've added a couple of comments, but overall I'm iffy on this.   Yes it would probably be helpful for some people to have this type of error message, but it's also that kind of things that will encourage people to just pass visible: false into their text query to get tests passing -- which is not something that should be encouraged.

lib/capybara/queries/text_query.rb
```diff
@@ -42,6 +53,15 @@ def negative_failure_message
       def valid_keys
         COUNT_KEYS + [:wait]
       end
+
+      def visible?
+        @type == :visible or (@type.nil? and Capybara.ignore_hidden_elements)
```
twalpole: this would need to take Capybara.visible_text_only into account to match the behavior of Node#text

alexch:
I'm not convinced it would encourage people to just pass :all into their text queries just to get their tests to pass. If some people don't care about their tests being accurate then there's not much a reasonable error message can do for them :-). And it's not suggesting any action, just pointing out the (possibly new or changed) reality and giving a clue -- which I would have appreciated when working with some buttons that toggle invisible checkboxes on a codebase I was new on last week.

lib/capybara/queries/text_query.rb
alexch: I was (and remain) very confused by all the different visibility settings Capy has. When I wrote this line I was looking at the docs for Element#text which only mentions `ignore_hidden_elements` but you're right, `visible_text_only` is looked at too (https://github.com/jnicklas/capybara/blob/master/lib/capybara/node/element.rb#L56).

lib/capybara/queries/text_query.rb
alexch: I started off with the total but decided the message was clearer by subtracting the visible hits from the total hits to make it clear that the invisible count was different. Got a better way of phrasing it?

I agree about not mentioning invisible hits in the negative case, I guess... It's hard to think through all the cases.

twalpole: well saying a certain number was found in invisible text isn't really right in all cases

    <div>hello he<span>llo he</span>llo</div>

if spans are hidden is it really right to say 1 was found in the invisible text? I think a better way would be to use the total and say something like "However, it was found 3 times when invisible text is included"
If spans are hidden then visible will find 2 instances of 'hello'

lib/capybara/queries/text_query.rb
```diff
@@ -14,11 +14,12 @@ def initialize(*args)
...
 
       def resolve_for(node)
-        @actual_text = Capybara::Helpers.normalize_whitespace(node.text(@type))
+        @node = node
+        @actual_text = text(node, @type)
         @count = @actual_text.scan(@search_regexp).size
       end
 
-      def failure_message
+      def failure_message(check_invisible = true)
```
twalpole: failure_message and negative_failure_message is an API required by RSpec - I don't think adding method parameters is a good idea in case RSpec every decides to update their API and pass some.  I think it would be better to factor out the message generation pieces into private methods that can then be called from failure_message and negative_failure_message

alexch: ok

lib/capybara/spec/session/assert_text.rb
```diff
@@ -165,6 +173,14 @@
     end.to raise_error(Capybara::ExpectationNotMet, 'expected not to find text "Some of this text is hidden!" in "Some of this text is hidden!"')
   end
 
+  it "should raise error if :all given and text is invisible." do
```
alexch: whoops, copy-and-paste error in the test title

lib/capybara/queries/text_query.rb
```diff
@@ -31,17 +42,31 @@ def failure_message
...
-      end
 
-      def negative_failure_message
-        failure_message.sub(/(to find)/, 'not \1')
-      end
+        if @node and visible? and check_invisible
+          invisible_text = text(@node, :all)
+          invisible_count = invisible_text.scan(@search_regexp).size
+          if invisible_count != @count
+            message << ". (However, it was found #{invisible_count} time#{'s' if invisible_count != 1} including invisible text.)"
```
twalpole: Please use the declension helper for pluralizing here

twalpole:
Just the one small change I commented on above, and then it looks good to me.  when you've made that change please squash the commits, and then I'll merge it in

twalpole:
This was squashed and merged as 51c83ed5a93f7a6f36f27019927d2ee7c4e2d599

PR #1596, opened 2015-11-03
asset_host now overrides href instead of ignoring it

Nakilon:
The page I'm testing has `<base>` starting with `/` tag, so local htm dump can't find CSS, etc. because the scheme is `file://`. So I would like to force the correct base in my tests.

I guess the original author https://github.com/jnicklas/capybara/commit/e0591fef5c9b4de9d56f71608043c86aed6a8fa4 made this function ignoring it the base tag already exists because he didn't want to bother with regex. So:
1) I've made it override href attribute if it already exists
2) I've changed the `[^<]` to `[^>]` in regex because I assume it's a bug
3) I've added `:focus` ability to rspec, becuse the `spec_helper.rb` requires all the tests, that makes impossible to run the one specific test
4) note you that different drivers reorder html attributes, that makes me unable to fully test the regex in rspec,, but I'm sure it's fine

### DIFF
--- a/lib/capybara/helpers.rb
+++ b/lib/capybara/helpers.rb
@@ -40,10 +40,13 @@ module Capybara
     # @return [String]         The modified HTML code
     #
     def inject_asset_host(html)
-      if Capybara.asset_host && Nokogiri::HTML(html).css("base").empty?
-        match = html.match(/<head[^<]*?>/)
-        if match
-          return html.clone.insert match.end(0), "<base href='#{Capybara.asset_host}' />"
+      if Capybara.asset_host
+        if Nokogiri::HTML(html).css("base").empty?
+          unless Nokogiri::HTML(html).css("head").empty?
+            html.sub!(/<head[^>]*?>/, "\\0<base href='#{Capybara.asset_host}' />")
+          end
+        else
+          html.sub!(/(<base[^>]*?href\s*=\s*(['"])).*?\2/, "\\1#{Capybara.asset_host}\\2")
         end
       end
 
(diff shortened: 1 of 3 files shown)


Nakilon:
Test `"executes successfully even if the page is missing a <head>"` appears to work not fully correctly
with selenium driver because Firefox adds tags on his own: `"<html xmlns=\"http://www.w3.org/1999/xhtml\"><head></head><body>Bar</body></html>"`
I spotted it when for some reason `$& + "` instead of `"\\0` didn't work despite the http://stackoverflow.com/a/32409118/322020

twalpole:
Thanks for the efforts,  however this isn't something I'm going to be merging into Capybara in the near future for a few reasons.  

1) the decision to not modify the base tag if it already existed was made in the discussion when this feature was added (not just because someone didn't want to write a regex) - https://github.com/jnicklas/capybara/pull/958  -  Also your current replacement regex has at least one error since attributes don't have to be quoted unless they contain certain characters
2) As discussed in the above thread the [^<] was for malformed html, may not really be required, but also isn't hurting anything
3) That is specifically not in current code because things like people leaving :focus tags in their code means the automated tests don't fully run (as happened with your first PR).  If you want to test specific tests just use `rspec -t focus`  - it's also not really relevant to the existing PR so we would normally prefer that as a separate PR (or at least a separate commit)
4) I can't test it but "I'm sure it's fine" really isn't something a project wants to hear

Nakilon:
4) Existing regexes are not fully tested for the same reason. Why then they are in the project?

twalpole:
Huh?  The project has been built over quite a bit of time and there very well may be pieces that aren't fully tested. That doesn't mean just because someone says "I'm sure it's fine" we should assume it is,especially since in this case its definitely not fine (see issue 1).  Next time you may be better off opening an issue about new functionality to discuss whether it is something that project would entertain before putting in the effort to implement it.  

If you see other regexs in the project which you consider wrong, please feel free to submit PRs with tests that show the problems.

PR #1597, opened 2015-11-03
Docs: Another example of `Capybara.string`

jaredbeck:
I read the existing documentation, but I was still having trouble with a single element (I was missing the `first('a')` part), so I thought another example would help someone else.


jaredbeck:
Thanks Thomas

PR #1599, opened 2015-11-06
Temporarily use different options for a driver.

thehenster:
When testing an action that redirects offsite it is useful to be able to not follow redirects just temporarily when it is enough to assert that a redirection has occured to the correct location.

Currently when using the Capybara::RackTest::Driver an attempt to redirect to an external site results in a request with the same path to the app currently under test. Most of the time this will cause a test to fail as a request is made to a non-existent path.

I know the ethos behind capybara is that it shouldn't do things users can't do with browsers. This is a bending of the rules to improve usability.. so just throwing it out there.. Happy to add some more test cases if you think it might be useful.

### DIFF
--- a/lib/capybara/rack_test/driver.rb
+++ b/lib/capybara/rack_test/driver.rb
@@ -90,6 +90,16 @@ class Capybara::RackTest::Driver < Capybara::Driver::Base
     !@browser.nil?
   end
 
+  def with_options(override_options)
+    original_options = @options
+    @options = @options.merge(override_options)
+
+    yield(self)
+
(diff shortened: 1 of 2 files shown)


twalpole:
Thanks for the effort but I'm not going to include for a couple of reasons
1. Capybara really is the wrong tool to be using if you're checking result status codes, as evidenced by the fact you have to go to driver specific methods to access them
2. if you want a driver that doesn't redirect you can just register a :racktest_no_redirect driver and use that for the specific tests you don't want redirection on.

PR #1612, opened 2015-11-23
Security update for nokogiri

Kosmas:
https://github.com/sparklemotion/nokogiri/blob/v1.6.6.x/CHANGELOG.rdoc#1664--2015-11-19


twalpole:
We're not going to bump the minimum supported version of nokogiri in the capybara gem. This is something that should be handled in the project using capybara and is only relevant if the person is using the vendored versions of libxml2 and libxslt.  The current spec of >= 1.3.3 does not prevent use of the newest release

PR #1624, opened 2015-12-15
Add more RSpec aliases

Bartuz:
Introducing `xfeature`, `ffeature` and `fscenario`

### DIFF
--- a/lib/capybara/rspec/features.rb
+++ b/lib/capybara/rspec/features.rb
@@ -9,6 +9,8 @@ if RSpec::Core::Version::STRING.to_f >= 3.0
 
   RSpec.configure do |config|
     config.alias_example_group_to :feature, :capybara_feature => true, :type => :feature
+    config.alias_example_group_to :xfeature, :capybara_feature => true, :type => :feature, :skip => "Temporarily disabled with xfeature"
+    config.alias_example_group_to :ffeature, :capybara_feature => true, :type => :feature, :focus => true
     config.alias_example_to :scenario
     config.alias_example_to :xscenario, :skip => "Temporarily disabled with xscenario"
     config.alias_example_to :fscenario, :focus => true
--- a/spec/rspec/features_spec.rb
+++ b/spec/rspec/features_spec.rb
@@ -70,13 +70,25 @@ feature "given and given! aliases to let and let!" do
   end
 end
 
-feature "if xscenario aliases to pending then" do
-  xscenario "this test should be 'temporarily disabled with xscenario'" do
-  end
-end
-
 feature "Capybara's feature DSL with driver", :driver => :culerity do
   scenario "switches driver" do
     expect(Capybara.current_driver).to eq(:culerity)
+  end
+end
+
+if RSpec::Core::Version::STRING.to_f >= 3.0
+  xfeature "if xfeature aliases to pending then" do
+    scenario "this should be 'temporarily disabled with xfeature'" do; end
+    scenario "this also should be 'temporarily disabled with xfeature'" do; end
+  end
+
+  ffeature "if ffeature aliases focused tag then" do
+    scenario "scenario inside this feature has metatag focus tag" do |example|
+      expect(example.metadata[:focus]).to eq true
+    end
+
+    scenario "other scenarios also has metatag focus tag " do |example|
+      expect(example.metadata[:focus]).to eq true
+    end
   end
 end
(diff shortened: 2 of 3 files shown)


lib/capybara/rspec/features.rb
```diff
@@ -21,9 +23,12 @@ def self.included(base)
           alias :background :before
           alias :scenario :it
           alias :xscenario :xit
+          alias :fscenario :fit
```
Bartuz: does RSpec 2.X has `fit` or `fdescribe` aliases? 

http://www.relishapp.com/rspec/rspec-core/v/2-14/docs/configuration/alias-example-to

Bartuz:
How does the test output look like:
test_output

Bartuz:
Only Ruby 1.9.3 build (the second one? there are two builds on 1.9.3) fails. https://travis-ci.org/jnicklas/capybara/jobs/96986194
with message: 
```
NoMethodError: undefined method `exclude_pattern=' for #<RSpec::Core::RakeTask:0x000000015f71b0`
```

Any assistance?

twalpole:
As a guess RSpec 2.2 doesn't support exclude_pattern=   -- don't break out these tests into a separate task and just test for the metadata in the test rather than that true == true.

spec/rspec/features_focus_tag_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'spec_helper'
+require 'capybara/rspec'
+
+RSpec.configuration.before(:each, { file_path: "./spec/rspec/features_spec.rb" } ) do
+  @in_filtered_hook = true
+end
+
+ffeature "if ffeature alises focused tag then" do
```
twalpole: spelling - aliases

Bartuz: :crying_cat_face:

Bartuz:
@twalpole I have merged these tasks and refactored focus tag specs

lib/capybara/rspec/features.rb
```diff
@@ -21,9 +23,12 @@ def self.included(base)
...
           alias :given :let
           alias :given! :let!
           alias :feature :describe
+          alias :xfeature :xdescribe
```
twalpole: The base version of RSpec we currently claim support for (2.2) doesn't support xdescribe -- don't add any of these changes for RSpec 2 - capybara will no longer support it in the next major release anyway

spec/rspec/features_spec.rb
```diff
@@ -70,8 +70,18 @@ def current_example(context)
   end
 end
 
-feature "if xscenario aliases to pending then" do
-  xscenario "this test should be 'temporarily disabled with xscenario'" do
+xfeature "if xfeature aliases to pending then" do
```
twalpole: This needs a condition so the test is not run for RSpec < 3  - probably the other additions too

Bartuz:
rbx-2 uses rspec-core 3.4.1 and fails after removing rspec 2 aliases. need to investigate

spec/rspec/scenarios_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+require 'capybara/rspec'
+
+RSpec.configuration.before(:each, { file_path: "./spec/rspec/features_spec.rb" } ) do
+  @in_filtered_hook = true
+end
+
+if RSpec::Core::Version::STRING.to_f >= 3.0 || RUBY_ENGINE == 'rbx'
```
Bartuz: @twalpole I don't know if it's good idea to put if statement for `rbx`. I'm unfamiliar with it at all. Do you think it's acceptable?

twalpole: I have no idea why 'rbx' would cause a failure here, and really don't want to skip testing on it without understanding exactly why it's failing

spec/rspec/scenarios_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+require 'capybara/rspec'
+
+RSpec.configuration.before(:each, { file_path: "./spec/rspec/features_spec.rb" } ) do
```
twalpole: This isn't needed in this file is it?

Bartuz: Nice catch! Actually `@in_filtered_hook = true` was needed for rbnx tests! They pass now :)

twalpole:
Ok looks like the tests pass, please squash this down to one commit

Bartuz:
Should I force push to `bartuz/master` or create new PR?

twalpole:
Just squash and force push

Bartuz:
@twalpole thanks you assistance and patience! :)

twalpole:
Thanks a lot!

PR #1632, opened 2015-12-30
Replace deprecated google code Selenium docs with correct GitHub docs

Nitrodist:
The Selenium project has moved from code.google.com to github.com, so
this commit modifies the link to the correct wiki page.

There are a few more mentions of code.google.com where two issues are
linked, but the migration to github.com has not included the historical
issues, so those links cannot be updated. The links themselves, though,
do still continue to work.


twalpole:
Thanks

PR #1639, opened 2016-01-20
Set cookie to track session requests and errors when sharing server b…

twalpole:
…etween sessions.  When using multiple sessions with the same app Capybara, by default, reuses the server.  This means when errors are raised in a session, or requests are made from the browser (ajax, etc) we have no way of knowing which session they are associated with, and therefore errors get raised in only the first session.  A setting Capybara.reuse_server was recently merged into Capybara, but that has the downside of using another port, more memory usage, etc.  This PR adds another method of tracking the session by adding a cookie to the browser and using that to track the session id.  Unfortunately with selenium we can't add a cookie to a domain unless the browser is already on a page on that domain (other drivers can, so it may be worth moving this functionality into a method that can be overridden).  To get around that I added a query parameter to every call to #visit which is stripped out by the middleware, used to identify the session, and then set as a cookie in the browser with the response so further requests should pass it back.  Another thought was to have the browser visit a page in the middleware (like **identify**) before every #visit to set the cookie.  Any thoughts anyone reading this has would be great.

### DIFF
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -5,6 +5,16 @@ require 'rack'
 module Capybara
   class Server
     class Middleware
+      class Session
+        attr_accessor :error
+        attr_reader :counter, :id
+
+        def initialize(id)
+          @counter = Counter.new
+          @id = id
+        end
+      end
+
       class Counter
         attr_reader :value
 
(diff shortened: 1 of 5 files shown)


lib/capybara/server.rb
```diff
@@ -5,6 +5,16 @@
 module Capybara
   class Server
     class Middleware
```
abotalov: Should `# @api private` be added for Capybara::Server::Middleware?

twalpole: Probably - I'm just not convinced I like the whole solution - would much prefer if selenium allowed us to preset a cookie before the first visit

PR #1642, opened 2016-01-22
Add `#has_status_code?` in Session Matchers

syndbg:
Just a syntactic sugar method that would make RSpec expectations a little better to read.

Usage in RSpec would be:

```Ruby
expect(page).to have_status_code(200)
```

Compared to the current available:

```Ruby
expect(page.status_code).to eq(200)
```

@twalpole

### DIFF
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -52,7 +52,7 @@ module Capybara
       :windows, :open_new_window, :switch_to_window, :within_window, :window_opened_by,
       :save_page, :save_and_open_page, :save_screenshot,
       :save_and_open_screenshot, :reset_session!, :response_headers,
-      :status_code, :current_scope,
+      :status_code, :has_status_code?, :current_scope,
       :assert_current_path, :assert_no_current_path, :has_current_path?, :has_no_current_path?
     ] + DOCUMENT_METHODS
     MODAL_METHODS = [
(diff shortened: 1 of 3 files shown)


syndbg:
Hmm... the Travis Selenium driver doesn't support `#status_code`. 

What's the preferred approach here when you want to test driver-dependant logic?

twalpole:
@syndbg The reason Capybara doesn't already have a have_status_code matcher is because it doesn't really make a lot of sense to check in feature tests (rather in controller tests where capybara isn't used).  

There are two reasons for this:

1. Capybara attempts to emulate a user as much as possible and users don't really see response codes.
2. Browsers automatically follow redirects, etc - so you can't really check for a large portion of the codes you'd want to be testing for.

Instead, in feature tests you should be testing for content that appears on pages, which indirectly is testing the status codes, i.e. if the correct content appears you know the response was a 2xx 

To answer your question though, driver feature specific tests are handled through :requires metadata on the tests as seen at https://github.com/jnicklas/capybara/blob/2.6_stable/lib/capybara/spec/session/response_code.rb#L2

On a separate note, the matchers in Capybara generally employ a waiting behavior and provide support for RSpec as well as minitest, etc.  A good example of how something like this would be implemented is the have_title (or have_current_path) matcher, so if you have other matcher contributions you'd like to make in the future please follow those as examples, and probably ask on the mailing list mentioned in the README whether they would be considered before putting in the work.


Thanks

PR #1651, opened 2016-02-03
Add match selector to match a given element against selectors

twalpole:


lib/capybara/node/matchers.rb
```diff
@@ -131,6 +152,45 @@ def assert_no_selector(*args)
...
+      #
+      # It also accepts all options that {Capybara::Node::Finders#all} accepts,
+      # such as :text and :visible.
+      #
+      #     node.assert_match_selector('li', :text => 'Horse', :visible => true)
+      #
+      # @param (see Capybara::Node::Finders#all)
+      # @raise [Capybara::ExpectationNotMet]      If the selector does not match
+      #
+      def assert_match_selector(*args)
```
abotalov: Do you think that this method name is descriptive enough to differentiate it from others that check descendants of an element instead of an element itself? What about including `self` into method name? e.g. `assert_self_matches_selector`

twalpole: I've been thinking about that too, what do you think of `assert_is_selector` ?. I don't want it to get too long

twalpole: hmm -- although is_selector would imply be_selector for the RSpec matchers which would be overloading on top of the RSpec default predicate selector -- not sure if thats a good idea

doits: since `be`/`is` is already taken, I'd vote for `matches` to keep it short.

Issue #1659, opened 2016-02-25
page.attach_file options aren't listed in documentation


alextaylor000:
The documentation for `Capybara::Node::Actions#attach_file` doesn't specify the options that can be passed to it - for example, passing `:match => :first` to get the first match in the case of ambiguous selectors.

I'm happy to submit a PR updating the relevant docs, I just wanted to confirm that the valid options are whatever is valid for `Class: Capybara::Query`, i.e.:

```
VALID_KEYS = [:text, :visible, :between, :count, :maximum, :minimum, :exact, :match, :wait]
VALID_MATCH = [:first, :smart, :prefer_exact, :one]
```

Thanks :)

twalpole:
The allowed options would be the ones allowed by #find plus any extras added by the :file_field selector however some don't really make any sense - :text for instance since a file field doesn't have text content, the count ones (count, maximum, minimum, between) don't really make any sense in this context - so I guess the full list of options that may make some sense is :match, :exact, :wait, :disabled, :multiple

alextaylor000:
Great, thanks @twalpole! PR is cookin'.

twalpole:
Closed through PR #1660

PR #1660, opened 2016-02-26
Add documention for Capybara::Node::Actions#attach_file options hash

alextaylor000:
This PR adds some missing documentation for the `attach_file` method, which takes an options hash but didn't previously document what parameters it accepted. Please let me know if the descriptions make sense.

This is a fix for #1659

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -150,6 +150,11 @@ module Capybara
       # @param [String] locator       Which field to attach the file to
       # @param [String] path          The path of the file that will be attached, or an array of paths
       #
+      # @option options [Symbol] match     The matching strategy to use (:one, :first, :prefer_exact, :smart)
+      # @option options [Boolean] exact    Match the exact locator name or accept a partial match
+      # @option options [Fixnum] wait      If using a Javascript driver, number of seconds during which the element will be searched for.
+      # @option options [Boolean] multiple Match field which allows multiple file selection
+      #
       def attach_file(locator, path, options={})
         Array(path).each do |p|
           raise Capybara::FileNotFound, "cannot attach file, #{p} does not exist" unless File.exist?(p.to_s)
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -150,10 +150,10 @@ module Capybara
       # @param [String] locator       Which field to attach the file to
       # @param [String] path          The path of the file that will be attached, or an array of paths
       #
-      # @option options [Symbol] match (:smart)     The matching strategy to use (:one, :first, :prefer_exact, :smart).
-      # @option options [Boolean] exact (true)    Match the exact label name/contents or accept a partial match.
+      # @option options [Symbol] match (Capybara.match)     The matching strategy to use (:one, :first, :prefer_exact, :smart).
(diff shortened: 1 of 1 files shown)


lib/capybara/node/actions.rb
```diff
@@ -150,6 +150,11 @@ def unselect(value, options={})
       # @param [String] locator       Which field to attach the file to
       # @param [String] path          The path of the file that will be attached, or an array of paths
       #
+      # @option options [Symbol] match     The matching strategy to use (:one, :first, :prefer_exact, :smart)
```
twalpole: Should add the default value

lib/capybara/node/actions.rb
```diff
@@ -150,6 +150,11 @@ def unselect(value, options={})
...
+      # @option options [Boolean] exact    Match the exact locator name or accept a partial match
```
twalpole: exact only affects searches for text content of elements, so in this case it would only have an effect on searching label element contents

lib/capybara/node/actions.rb
```diff
@@ -150,6 +150,11 @@ def unselect(value, options={})
...
+      # @option options [Fixnum] wait      If using a Javascript driver, number of seconds during which the element will be searched for.
```
twalpole: It's the maximum number of seconds

twalpole:
I've added a couple of comments to clarify things --  Also when adding documentation only PRs - if you add [ci skip] to end of your commit message it will prevent travis from wasting time doing builds

alextaylor000:
@twalpole, thanks for the feedback (and the tip about `[ci skip]`). I've updated the PR based on your feedback.

One thing I wasn't entirely sure about was what the default value for `multiple` is. From what I understand about [how the Selector is loaded](https://github.com/jnicklas/capybara/blob/master/lib/capybara/selector.rb#L283), it appears to be `true`, but let me know if I've got this wrong.

twalpole:
@alextaylor000 From https://github.com/jnicklas/capybara/blob/master/lib/capybara/selector.rb#L287 you can see the `multiple` option doesn't have a default, so it's ignored unless passed in.  A `default` option would be passed in if there was a default value (like the `disabled` filter above) --  `boolean: true` is just a syntactic sugar instead of passing `valid_values: [true,false]` to the filter

lib/capybara/node/actions.rb
```diff
@@ -150,6 +150,11 @@ def unselect(value, options={})
       # @param [String] locator       Which field to attach the file to
       # @param [String] path          The path of the file that will be attached, or an array of paths
       #
+      # @option options [Symbol] match (:smart)     The matching strategy to use (:one, :first, :prefer_exact, :smart).
```
twalpole: Default here is actually  Capybara.match

lib/capybara/node/actions.rb
```diff
@@ -150,6 +150,11 @@ def unselect(value, options={})
...
+      # @option options [Boolean] exact (true)    Match the exact label name/contents or accept a partial match.
```
twalpole: Capybara.exact is the default

alextaylor000:
Thanks for the notes, I think I was looking too far into `file` to figure out those default values. Updated the PR.

twalpole:
Thanks!

PR #1665, opened 2016-03-10
added session#last and specs, update history.md

Tmee:
I added Session#last which returns the last node found matching the given criteria.

### DIFF
--- a/History.md
+++ b/History.md
@@ -10,6 +10,8 @@ Release date: unreleased
 * :multiple filter added to relevant selectors [Thomas Walpole]
 * Provided server registrations for :webrick and :puma. Capybara.server = :puma for testing with Rails 5 [Thomas Walpole]
 * Deprecate passing a block to Capybara::server user Capybara::register_server instead [Thomas Walpole]
+* Added Session#last. This acts similar to Session#first, except it returns the last node matching the given criteria [Tim Mee]
+
 
 #Version 2.6.2
 Relase date: 2016-01-27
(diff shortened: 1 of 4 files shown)


Tmee:
Hi @jnicklas, any thoughts on bringing this into the next release?

twalpole:
@Tmee jnicklas doesn't really work on Capybara anymore - I'm the person doing most development nowadays.  My concern is whether `#last` justifies expanding the API .  Often while testing one wants to access the first element (or at least an element if any matches exist), I'm not sure how often last would be used, and it's already easy enough to just do `all(...).last`

twalpole:
#first and #all  (and #last if added) also have the major downside of being non-reloadable elements, which can lead to all sorts of flaky test issues.

Tmee:
ah, sorry about the @mention jnicklas.  Ok, I noticed I was typing `all(...).last` a bunch and thought it would be nice to have a `last(...)` similar to the awesome `first(...)` finder.  But no worries, thanks for getting back to me!

twalpole:
If you don't mind explaining - what type of testing are you doing that you're calling `all(..).last` all the time?

Tmee:
Just UI automation specs with selenium.  There have been a hand full of times, similar to the amount I use #first, that I've wanted to check the last element in table or in a list where I have to write out `all(...).last`.

twalpole:
ok -- in most cases you're better off using last-child/last-of-type css selectors with `#find` if/when possible - People really tend to overuse #all/first and it causes a large number of the issues people have with Capybara being 'flaky'  due to it's lack of waiting behavior, and the result elements being non-reloadable

Tmee:
ok, thanks for the info!

PR #1667, opened 2016-03-18
Default to localhost, not 127.0.0.1, if the system prefers IPv6

sodabrew:
In eventmachine/eventmachine#703, we found that Capybara.server_host defaults to `127.0.0.1`, while Thin defaults to `localhost`. With the new EventMachine 1.2.0 gem, we're using the system's `getaddrinfo` to return both IPv4 and IPv6 addresses in whichever order the local operating system prefers. This allows Thin to bind to `::1` by default, but Capybara's default of `127.0.0.1` fails to connect.

By allowing Capybara to also follow the local operating system preference, both client and server will bind and talk on the same local IP address.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -295,7 +295,7 @@ module Capybara
     # @return [String]    The IP address bound by default server
     #
     def server_host
-      @server_host || '127.0.0.1'
+      @server_host || 'localhost'
     end
 
     ##
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -65,7 +65,7 @@ module Capybara
       @server_thread = nil # suppress warnings
       @host, @port = host, port
       @port ||= Capybara::Server.ports[Capybara.reuse_server ? @app.object_id : @middleware.object_id]
-      @port ||= find_available_port
+      @port ||= find_available_port(host)
     end
 
     def reset_error!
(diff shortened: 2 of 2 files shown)


twalpole:
https://github.com/jnicklas/capybara/commit/89d00c4f0801a1d00f57c4bedb81c23b888e18df   - host (which over time became the server_host value) was specifically set to 127.0.0.1 many years ago  to "prevents some issues" .  I have no clue what those issues were, or if they would still be issues.  Other than that this makes sense to me - lets see if @jnicklas has any recollection of what the issues were? :)

jnicklas:
I think that it was maybe some kind of CORS issue? I can't quite remember. Also it's possible that `localhost` is not actually set, if it's not included in the host file. On the other hand from what I understand `localhost` should be resolved by DNS if it's not set in the local configuration.

I do want to point out thought that this is a breaking change. A lot of people have written a lot of test suites which assume that Capybara binds on 127.0.0.1. If this is changed it's going to annoy a lot of people.

twalpole:
@sodabrew  As @jnicklas pointed out, this would be a breaking change - so I'm not going to merge it.  I did however cherry pick the second commit which makes find_available_port check for a port on the correct interface.  For your team to maintain the same behavior it should just be a matter of you setting `Capybara.server_host = 'localhost'` now

sodabrew:
Yes sounds reasonable, thanks! And I've at least made a little paper trail so someone with this error and the updated EventMachine should be able to find this config workaround easily on their favorite search engine :)

PR #1668, opened 2016-03-24
move xpath gem html definitions into capybara

twalpole:
Move the html element selector definitions from the xpath gem into the capybara selectors and modify them so nil locators devolve into just finding element types


jnicklas:
:tada:

PR #1671, opened 2016-03-30
Leave it up to the driver to make sure that resetting is synchronous

### LINKED ISSUE #1670: Race condition in open alerts with Selenium
So this issue is totally my fault :( I'm experiencing an intermittent failure in an application where an alert appears just after a browser has reset the session, since it's in an unload handler. Here's how Capybara currently implements reset:

``` ruby
    def reset!
      if @touched
        driver.reset!
        assert_no_selector :xpath, "/html/body/*" if driver.browser_initialized?
        @touched = false
      end
      @server.wait_for_pending_requests if @server
      raise_server_error!
    end
```

In my case `assert_no_selector` is raising an unhandled alert error.

The reason `assert_no_selector` is called here in the first place is to make sure that resetting the session is synchronous. Instead of relying on the driver to make sure that it is, I thought that I could cheat and let Capybara handle this for the drivers. It seems that that was a mistake. That already became apparent in the need for the `browser_initialized?` hack. Of course we could handle this annoying case here as well, but there's no really way to rescue a Selenium-specific error here that isn't totally gross.

IMO we should cut out losses on this one and simply force the driver to make sure that restting is synchronous.

jnicklas:
I think that this was a futile attempt to resolve problems with restting sessions which should be much better handled by the addition of `wait_for_pending_requests`.

Closes #1670

### DIFF
--- a/lib/capybara/driver/base.rb
+++ b/lib/capybara/driver/base.rb
@@ -91,8 +91,8 @@ class Capybara::Driver::Base
   def no_such_window_error
     raise Capybara::NotSupportedByDriverError, 'Capybara::Driver::Base#no_such_window_error'
   end
-  
-  
+
+
   ##
   #
   # Execute the block, and then accept the modal opened.
@@ -133,9 +133,5 @@ class Capybara::Driver::Base
 
   def needs_server?
     false
-  end
-
-  def browser_initialized?
-    true
   end
 end
--- a/lib/capybara/rack_test/driver.rb
+++ b/lib/capybara/rack_test/driver.rb
@@ -66,11 +66,11 @@ class Capybara::RackTest::Driver < Capybara::Driver::Base
   def find_xpath(selector)
     browser.find(:xpath, selector)
   end
-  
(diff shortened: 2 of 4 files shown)


twalpole:
The problem with wait_for_pending_requests is that it's not 100% reliable when a server is shared between sessions

jnicklas:
Interesting! Why is that?

Regardless I'm not actually convinced that the assert removed in this PR actually does *anything*, at least for Selenium. Other drivers might rely on it to some extent, I'm not sure, but I'm pretty sure that `nagivate.to` is synchronous for Selenium, so this really *shouldn't* make any difference.

twalpole:
I misspoke on this, the issue with wait_for_pending_requests, when the server is shared, is that the middleware has no way to determine which session the requests are coming from, so reseting one session will wait for both sessions requests to end which, as far as reset being synchronous is concerned, isn't really an issue.

twalpole:
Included in PR#1672

PR #1672, opened 2016-04-04
Selenium driver reset synchronous

twalpole:
Make sure `#reset` for the selenium driver is synchronous and handles modals that may appear.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ env:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
 
 addons:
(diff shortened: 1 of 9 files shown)


jnicklas:
This is amazing! Thank you so much for putting so much effort into this.

twalpole:
@jnicklas you're welcome -- I tried it with capybara-webkit and poltergeist and the removal of the assert_no_selector in Session#reset! didn't appear to affect them, so hopefully it doesn't --- I'm sure we'll hear if it does :)

PR #1673, opened 2016-04-04
Warn when extra parameters passed to SelectorQuery

twalpole:
Warn when unused parameters are passed to SelectorQuery


twalpole:
Fixes issue #1395

PR #1675, opened 2016-04-05
Explicit README heading anchors, to work on both GitHub and RubyDoc.info

afoster:
The anchor links in the table of contents at http://www.rubydoc.info/github/jnicklas/capybara/master don't work, because `yard`'s auto anchors differ from the content. But GitHub's work. So I added explicit heading anchors so they should work on both sites.


twalpole:
Thanks!

PR #1678, opened 2016-04-06
Fix "bald contenteditable" bug

chadoh:
A "bald contenteditable" is one added to an element like this:

    <div contenteditable>

Rather than this:

    <div contenteditable="true">

The latter is currently supported by Capybara, but the former is not.

### DIFF
--- a/lib/capybara/spec/session/node_spec.rb
+++ b/lib/capybara/spec/session/node_spec.rb
@@ -126,12 +126,19 @@ Capybara::SpecHelper.spec "node" do
       expect(@session.find(:css,'#blank_content_editable').text).to eq('WYSIWYG')
     end
 
+    it 'should allow me to change the contents of a contenteditable without =true', :requires => [:js] do
+      @session.visit('/with_js')
+      @session.find(:css,'#sleek_content_editable').set('WYSIWYG')
+      expect(@session.find(:css,'#existing_content_editable_child').text).to eq('WYSIWYG')
+    end
+
     it 'should allow me to change the contents of a contenteditable elements child', :requires => [:js] do
       pending "Selenium doesn't like editing nested contents"
       @session.visit('/with_js')
       @session.find(:css,'#existing_content_editable_child').set('WYSIWYG')
       expect(@session.find(:css,'#existing_content_editable_child').text).to eq('WYSIWYG')
     end
+
   end
 
   describe "#tag_name" do
--- a/lib/capybara/spec/views/with_js.erb
+++ b/lib/capybara/spec/views/with_js.erb
@@ -40,6 +40,7 @@
     <p>
       <div contenteditable='true' id='existing_content_editable'>Editable content</div>
       <div contenteditable='true' id='blank_content_editable' style='height: 1em;'></div>
+      <div contenteditable id='sleek_content_editable'>Editable content</div>
       <div contenteditable='true' style='height: 1em;'>
         <div id='existing_content_editable_child' style='height: 1em;'>Content</div>
       </div>


chadoh:
So far, all I've added is a test to expose the problem. However, I have not had luck running this test locally! Is there a way that I can run a single test locally? I tried to do this in several different ways, such as the trusty `rspec lib/capybara/spec/session/node_spec.rb:129`, but everything I tried (other than the hella-lengthy & noisy `rake`) gave me this error:

```
/Users/chadoh/Code/capybara/lib/capybara/spec/session/node_spec.rb:3:in `<top (required)>': uninitialized constant Capybara (NameError)
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1361:in `load'
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1361:in `block in load_spec_files'
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1359:in `each'
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1359:in `load_spec_files'
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/lib/rspec/core/runner.rb:106:in `setup'
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/lib/rspec/core/runner.rb:92:in `run'
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/lib/rspec/core/runner.rb:78:in `run'
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/lib/rspec/core/runner.rb:45:in `invoke'
        from /Users/chadoh/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-3.4.4/exe/rspec:4:in `<top (required)>'
        from /Users/chadoh/.rbenv/versions/2.2.3/bin/rspec:23:in `load'
        from /Users/chadoh/.rbenv/versions/2.2.3/bin/rspec:23:in `<main>'
```

twalpole:
@chadoh add a tag to the test you want to run separately 

    it 'should allow me to change the contents of a contenteditable without =true', :requires => [:js], my_test: true do

then you can do

    rspec -t my_test

to run just that test - remove the tag before committing

twalpole:
BTW -- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable  -- contenteditable is an enumerated attribute - not a boolean one.  As such 

    <div contenteditable>

is not valid - and it must be assigned a value of "true", "", or "false"

twalpole:
@chadoh - Here is the link to the HTML spec about the contenteditable attribute - https://www.w3.org/TR/2008/WD-html5-20080610/editing.html - Not specifying any value leads to "the inherit state, which is the missing value default (and the invalid value default)."  - so it is not the same as contenteditable="true" and the element should only be editable if it has an editable parent - I think this PR is probably invalid

chadoh:
Ah, good catch @twalpole! Thanks!

Issue #1463, opened 2015-01-07
Add :type option to show which element it is (+ related improvements)


abotalov:
I see the following downsides in some aspects of Capybara:

1. `find` always returns instance of `Element`, not a subclass of it. Therefore some methods that make sense only for some kinds of elements aren't provided (e.g. Select#selected_option, FileField#set, Image#loaded?, etc.)
2. :field, :radio_button, etc. are:
  * hard to debug if ElementNotFound occurs. Writer of the code may have meant id/name/placeholder/etc. but it's not easy to find if that element really exists on the DOM
  * difficult to remember which attributes are supported and which are not (is alt attribute supported?)
But people use them as:
  * `have_field('name')` looks better than `have_css("*[name='name']")`
  * there is no good alternative if you want to use selector-specific options.  
    `expect(page).to have_field('field_id', with: 'something')`  
    `expect(page).to have_select('label_text', with_options: ['42', '43'])`

Proposed solution:
1. Add `:name`, `:value`, `:label_text`, `:href`, etc. selector kinds out-of-the-box to make it easier to find elements in a deterministic way without necessity to write longer css/xpath selectors
2. Add `:type` option that specifies which kind of element it is. Examples:

```ruby
have_selector(:id, 'field_id', type: :field) # additional options specific to type: :field are available
have_selector(:field, 'field_id') # type: :field is implicit if selector kind is :field
find(:id, 'field_id', type: :field) # returns a subclass of Capybara::Node::Element, e.g. Capybara::Node::Element::Field
```

New methods `Capybara.register_element_type` and `Capybara.register_selector_kind` will be added. `Capybara.add_selector` will be deprecated. It will be possible to register both element type and selector kind using `add_selector` (some methods will be delegated to registering element type and some others to registering selector kind).

User would be able to define a subclass of `Capybara::Node::Element` and its relation to one of registered element types (some subclasses will be defined as part of Capybara).

`:type` won't be used to check element's tag but only to return element of a more specific class and to support type-specific options (like `:checked`, `:disabled`, `:href`, etc that are defined by `register_element_type`).

This change looks backwards compatible to me. 

WDYT?

twalpole:
Sorry for the delay on this - personal issues have kept me away for a while.
At first glance this sounds interesting, although I think it is two different issues and if implemented should probably be done as two different PRs.  A couple of issue
1. Its not clear to me why add_selector would be deprecated, or what register_selector_kind would be for. 
[...]

PR #1679, opened 2016-04-07
Filter sets

twalpole:
Implementation of reusable sets of filters for selectors - This came out of Issue #1463 .

I'm not convinced this is hugely useful, although for projects that build a large number of custom selectors it may be. Thinking about it.


twalpole:
I've decided I like this for sharing filters between selectors - merging

Issue #1696, opened 2016-05-09
Session reset times out if alert already present when trying to close the page


jnicklas:
#1672 introduced behaviour which tries to make sure that resetting the session is going to close any alerts from a `beforeunload` handler. Unfortunately, if such an alert is already open, there's a race condition where Capybara never navigates to about:blank and thus gets stuck in a loop trying to check for no content, which never happens. This causes the session reset to time out with an error like:

```
/Users/jonasnicklas/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/capybara-2.7.0/lib/capybara/selenium/driver.rb:114:in `reset!': Timed out waiting for Selenium session reset (Capybara::ExpectationNotMet)
	from /Users/jonasnicklas/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/capybara-2.7.0/lib/capybara/session.rb:109:in `reset!'
	from test.rb:10:in `<main>'
```

I have a repro [here](https://gist.github.com/jnicklas/a2ada84190f2cc9fdb95acec6a235dde). I'll try to submit a fix.

twalpole:
This issue is caused by selenium defaulting to dismissing unexpected alerts, which causes the navigation to about:blank to abort.  We catch this and attempt to accept the alert, but since the alert has already been dismissed seleniums alertTimeout (which I don't see a way to override but might be missing something) takes over and selenium waits for 2 seconds before telling us the modal has already gone away - cycle repeats until we timeout.  I've created PR #1698 which changes the default to ignoring (not dismissing or accepting) unexpected alerts so we can accept it.

twalpole:
Fixed via PR #1698

dunban1:
This error seems to still manifest when using the FF profile setup.

I reproduced it using:
- ruby-2.3.0 

gems:
- capybara (3.0.2) 
- rspec (3.6.0) 
- selenium-webdriver (3.11.0) 

driver:
- geckodriver 0.20.1 


```
Capybara.register_driver :selenium do |app|
    options = Selenium::WebDriver::Firefox::Options.new
    options.profile = 'default'
    desired_caps = Selenium::WebDriver::Remote::Capabilities.firefox(
      {
        marionette: true,
        accept_insecure_certs: true,
      }
    )
    
    Capybara::Selenium::Driver.new(
      app,
      browser: :firefox,
      options: options,
      desired_capabilities: desired_caps
    )
  end
  Capybara.current_driver = :selenium
[...]

PR #1697, opened 2016-05-09
Handle open modals on session reset

jnicklas:
Unfortunately the current way of doing this is prone to race conditions. See #1696. Selenium does not provide any API to check if a modal is open, and simply trying to close it and catching the error is very, very slow on Firefox. One quirk though is that Selenium returns the text of any open modal as the return value of `navigate.to`. Why? I have no idea. Is this even an official/intentional API? I don't know. (Maybe @jarib knows?). So it turns out that the fastest way of safely navigating away from a page and catching any alerts caused by this is by checking that the return value of `navigate.to` is neither `""` nor `nil` (the behaviour seems to differ between selenium drivers). I've checked that this works with Chrome, Firefox and IE via the remote driver at least.

Honestly I'm not sure if this is the right way of solving this, but I'm kind of out of other ideas.

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -92,39 +92,35 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
   def reset!
     # Use instance variable directly so we avoid starting the browser just to reset the session
     if @browser
-      navigated = false
       start_time = Capybara::Helpers.monotonic_time
-      begin
-        if !navigated
-          # Only trigger a navigation if we haven't done it already, otherwise it
-          # can trigger an endless series of unload modals
-          begin
-            @browser.manage.delete_all_cookies
-          rescue Selenium::WebDriver::Error::UnhandledError
-            # delete_all_cookies fails when we've previously gone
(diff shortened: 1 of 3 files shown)


twalpole:
Pretty sure selenium used to raise an error if you tried to navigate away while a modal was visible, so this must be a recent behavior change - maybe they replaced the error with returning the text? Seems like a crazy API change, but I'll see if theres any mention in recent selenium commits

twalpole:
I think the root cause of this is that selenium (with firefox at least) is dismissing the unhandled alert itself and no longer raising the error it used to (99% sure it used to).  Dismissing the alert rather than accepting it appears to prevent the location change so everything fails.  @jnicklas  - could you try setting the desired capability "unexpectedAlertBehaviour" (note English spelling of behaviour with a u) to "ignore" and see if your issue goes away, it fixes things in my tests and may be a preferable default since we are managing the alerts ourselves

twalpole:
@jnicklas  Please look at PR #1698 and give your thoughts on that as opposed to this one.

twalpole:
Closing in favor of PR #1698

PR #1698, opened 2016-05-09
default selenium to raising an error when there are unexpected alerts

twalpole:
Alternate to PR #1697  -  This changes selenium-webdriver to ignore unexpected alerts (raise an error) rather than the default of dismissing them which allows us to accept the alert from before unload and therefore allow unload to continue.

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -27,6 +27,7 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
   def initialize(app, options={})
     begin
       require 'selenium-webdriver'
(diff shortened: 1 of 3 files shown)


jnicklas:
It doesn't seem particularly nice to have to change the default value, but I guess modifying the `options` could be tricky since someone can specify their own capabilities?

jnicklas:
Works brilliantly though! I think behaviour wise this is far better than my PR.

twalpole:
@jnicklas - Updating the defaults seemed like the easiest way to make sure it was set all the time (unless overridden by the users options) - there really doesn't seem to be another place in the selenium API to do it nicely.

PR #1711, opened 2016-06-13
Add accessor for returning all sessions

jnicklas:
This is occasionally useful.

For example in a recent application I'm using SauceLabs, and I'd like to iterate through all session in order to close them in a different way.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -324,6 +324,16 @@ module Capybara
 
     ##
     #
+    # All sessions that Capybara knows about, this does not include manually started sessions
+    #
+    # @return [Array<Capybara::Session>]     All known sessions
+    #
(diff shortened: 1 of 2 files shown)


jnicklas:
It's potentially somewhat confusing, since it might well contain sessions not used in the current test, or nto contain a manually started session, so I'm not sure if this is really a fantastic idea. Thought I'd propose it anyway, since I've been finding myself calling `send(:session_pool)` a few times now.

twalpole:
I'm not really sure this is a great idea either.

twalpole:
I don't think I want to add this at the moment (for the reasons you brought up), and it's easy enough for people to work around if they really need the access.

jnicklas:
👍

Issue #1365, opened 2014-07-29
Proper API for working with frames


abotalov:
Hi,

I'd like to contribute a new API for working with frames (it can be done for e.g. 2.5):
- `Session#switch_to_frame(:top_level)`
- `Session#switch_to_frame(:parent)` (switch_to.parent_frame will be added in Selenium-Webdriver 2.43)
- `Session#switch_to_frame(element)` (element is Capybara::Node::Element)
- `Session#within_frame(element)`

Old `within_frame(frame_handle)` where "frame_handle" is passed to the driver (and in reality can be not only handle but also something else) will be deprecated like it was done with `within_window`.

Please approve.

jtrupiano:
Did this new API ever get merged in?  I need the #switch_to_frame methods.

route:
@jtrupiano I don't think so. Ended up here as proposal

jtrupiano:
@route thanks for the reply.  A question for you (and the lurkers) regarding the following situation.  I am directly working with Session instances (no naked calls to #within_frame, etc).  I have scenarios where I want to observe the same behavior when a request comes directly from the browser vs. when the request is loaded into an iframe on another page.  So in my test code, I want a "handle" of sorts to the top level Session-like object.  In other words, something like this...

    # return an object that behaves like Session that is scoped to an embedded iframe (if necessary)
    def get_root(session)
      if session.has_css?('iframe#MyFrame')
        return session.get_iframe('iframe#MyFrame') # should behave like a Session
      else
        return session
      end
    end
    
    root = get_root(session)
    root.fill_in 'Name', 'John'
    root.submit # etc

This is admittedly stripped down for illustrative purposes, but I want to re-use all of that code after the call to #get_root as if it were unaware that it were happening within an iframe.  So what's the recommended way to get a pointer/handle to the iframe that either is or behaves like a Session?

route:
I'm a little out of context but why you cannot use `within_frame` even if you manage session yourself? So basically it pushes an iframe to the stack and when block ends pops it.

jtrupiano:
I don't understand how to do that.  #within_frame doesn't yield a root-level "Session-like" object for me to work with (or does it??).

    session = Capybara::Session.new(Capybara.current_driver)
    session.visit "http://something.com"
    session.within_frame('iframe#MyFrame') do 
      # how would I invoke anything on that iframe from here?
      fill_in 'Name', 'John' # this will throw an error because I'm managing session myself
    end

[...]

PR #1722, opened 2016-07-15
new frame api for drivers

twalpole:
This changes the expected frame api for drivers from `within_frame` to `switch_to_frame` accepting either the frame elment, :parent, or :top to switch to the given frame, the parent frame, or the outermost frame.   

@jferris @mhoran @jonleighton @route  Please give your thoughts?  I'd like to make this change in Capybara 2.8, however if there are reasons this couldn't be implemented relatively quickly in any drivers I may have to reconsider.  Note this only changes the driver API not the user `within_frame` API which will continue to accept index, id/name, or an element.  Once the drivers have implemented the new method we can possibly look at improving the user facing API as discussed in Issue #1365 in a future Capybara release


jferris:
I don't think this should be a problem in capybara-webkit.

twalpole:
Ok - doesn't look like it's an issue for poltergeist either - merging

PR #1734, opened 2016-08-05
Optimize Capybara::Result for use with #all[idx]

twalpole:
This lazily evaluates the results from `#all` when possible.  Not sure if this is a good idea or not since it potentially delays the evaluation of visibility

### DIFF
--- a/lib/capybara/result.rb
+++ b/lib/capybara/result.rb
@@ -25,27 +25,74 @@ module Capybara
 
     def initialize(elements, query)
       @elements = elements
-      @result = elements.select { |node| query.matches_filters?(node) }
-      @rest = @elements - @result
+      @result_cache = []
+      @results_enum = lazy_select_elements { |node| query.matches_filters?(node) }
       @query = query
     end
 
-    def_delegators :@result, :each, :[], :at, :size, :count, :length,
-                   :first, :last, :values_at, :empty?, :inspect, :sample, :index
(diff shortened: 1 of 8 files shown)


lib/capybara/result.rb
```diff
@@ -25,27 +25,77 @@ class Result
...
+
+    def each(&block)
+      @result_cache.each(&block)
+      begin
+        loop do
+          next_result = @results_enum.next
+          @result_cache << next_result
+          block.call(next_result)
+        end
+      rescue StopIteration
```
abotalov: Could you please clarify when StopIteration may occur? Documentation tells that this exception is rescued internally by `loop`.

twalpole: Good catch -- fixing

lib/capybara/result.rb
```diff
@@ -25,27 +25,74 @@ class Result
...
+      @result_cache.each(&block)
+      loop do
+        next_result = @results_enum.next
+        @result_cache << next_result
+        block.call(next_result)
+      end
+      self
+    end
+
+    def [](idx)
```
abotalov: `Array#[]` supports also `[start, length] ` and `[range]` forms.
Do you think it would be better to take `*args` and call `full_results` if first item of args is not Integer >= 0 (for compatibility)?

twalpole: It should to maintain compatibility  -- it's a shame most people assume Result is actually an Array

PR #1744, opened 2016-08-23
clarify Firefox note on readme

titusfortner:
The current statement in the README is technically incorrect.

This is the code that should work to use Selenium 3.0 with the legacy Firefox driver.
```ruby
Capybara.register_driver :legacy_firefox do |app|
  Capybara::Selenium::Driver.new(
    app,
    browser: :firefox,
    desired_capabilities: Selenium::WebDriver::Remote::Capabilities.firefox(marionette: false)
  )
end
```

If this PR is too wordy, maybe the information and above code could be put somewhere else and referenced by link in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,13 @@ through an external gem.
 **Need help?** Ask on the mailing list (please do not open an issue on
 GitHub): http://groups.google.com/group/ruby-capybara
 
-**Note: Firefox 48** If you're using Firefox with selenium-webdrvier, stay on Firefox 47.0.1 and selenium-webdriver 2.53.4.  Firefox 48 requires geckodriver and is not supported without selenium-webdriver 3 which currently has multiple issues and is feature incomplete
+**Note: Firefox 48** If you're using Firefox with selenium-webdriver, 
+it is recommended that you use [Firefox ESR](https://www.mozilla.org/en-US/firefox/organizations/)
+or Firefox 47.0.1. Firefox 48+ requires [geckodriver](https://github.com/mozilla/geckodriver/releases). 
+To [use geckodriver in Selenium 2](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette/WebDriver#Ruby_3), 
+you can register a driver with a desired capability of `marionette: true`. This is set to true by
+default in Selenium 3, so to use the legacy Firefox driver, register a driver with a desired capability of `marionette: false`.
+Geckodriver still has multiple issues and is feature incomplete.
 
 ## Table of contents
 


twalpole:
While this is technically correct - Firefox 48 with selenium-webdriver 2.53.4 and marionette: true is basically unusable with Capybara due to multiple missing/broken features. Rather than deal with all the bug reports that will get filed we are going to stick with the current statement.  Once selenium-webdriver 3 releases we'll consider merging this in with a caveat depending on what functionality works/can be made to work at that point.

titusfortner:
> selenium-webdriver 3 beta w/ FF 47.0.1 and the legacy driver adds no benefit
Except that it does provide benefit because things have been updated that are not backported to the 2.x. But I get your point that it is beta and that you don't want to confuse people unnecessarily.

I would still urge:
1. Recommending ESR which doesn't auto-update and will get security patches rather than 47.0.1. This is what big companies like Salesforce are doing.

2. Adjusting the wording so it doesn't sound like it is Se 3 that is feature incomplete. :)
"...selenium-webdriver 3 which currently has multiple issues and is feature incomplete"

twalpole:
@titusfortner Unless I missed a recent commit, Selenium 3 (selenium-webdriver 3.x) is currently feature incomplete (when used with marionette) -  There is no way to request the current value or value of any other property of a Selenium::WebDriver::Element -  In Selenium 2.x requesting the value of an "attribute" of a Selenium::WebDriver::Element would give you the property if it existed and was a simple type (and boolean attributes were converted to true/false) otherwise it returned the attribute value.  In Selenium 3 all we have access to on Selenium::WebDriver::Element is the attribute value - that makes it unusable with Capybara.

titusfortner:
You are conflating everything upstream of Capybara as "Selenium" which is inaccurate. Selenium is the code that gets sent to binary executables, and those executables are somewhat inconsistent in the results they provide (pretty much all of them except for chromedriver & IEDriver at this point). Selenium itself is not feature incomplete, and as far as Ruby is concerned there are minimal functional differences between 2.x and 3.x except for what is the default firefox driver. There is nothing to "fix" in the Selenium code; it is up to the browser vendors to improve their products at this point, which they are actively doing.

Yes, I have a dozen or so bugs filed for Marionette, and you've already discussed the 3 most important ones - Actions Classes, Select Lists and Attribute Values, and I agree geckodriver won't be very usable without all 3 of these being addressed.

twalpole:
@titusfortner I'm not conflating everything as Selenium, I fully understand the lines between them  and I'm up to date on the spec changes going on in Webdriver.  W3CBridge provides an #element_value method which isn't exposed through Selenium::WebDriver::Element (implemented via the W3C getProperty API which geckodriver supports)  - and I believe I saw commits in other parts of selenium (non-ruby)  recently implement shims to emulate the 2.x behavior using the W3C getAttribute and getProperty methods which geckodriver has implemeted.  It would seem to me it's  currently selenium-webdriver that is deficient on that functionality?

titusfortner:
Ah yes, sorry; sure enough the Ruby bindings were behind. That's been fixed in master. Thanks.

For now everything is still going through the `Element#attribute` method, so it should be all backwards compatible. I'll take a look at the #element_value method, I have no idea why that one was never exposed in the Element class.

twalpole:
@titusfortner Is a nightly build of the selenium-webdriver gem generated anywhere that can be downloaded for our automated tests?  I was using it directly from the selenium github repo but since the switch to using the atom for getAttribute thats no longer possible.  I would like to be able to notify Capybara users as soon as our tests pass with it, which should be soon with the switch to the atom and FF fixing the select behavior in 49

titusfortner:
Yeah, I'm not super happy with that, either. I was hoping that once we deprecated the legacy Firefox & Safari drivers we wouldn't need to do any build process at all for the Ruby code, but I guess it makes sense that where JWP drivers use the atoms, w3c should not.

I have a local Jenkins I use to run Watir specs on Selenium master. I shallow clone Selenium and run: 

`./go //rb:gem:build`

Then set in the gemfile:

`gem 'selenium-webdriver', :path => '../selenium/build/rb'`

twalpole:
@titusfortner Thanks - I'll give something like that a try - on a side note I'm slightly shocked by the size of the atom produced - 37k over the wire every time the value of an "attribute" is requested seems a tad excessive although it does keep it down to one request.

titusfortner:
Huh, that's interesting. The one that we've had stored in the Watir project for years is only 4k. There might be a compile issue, I'll check with Simon & Jason.

twalpole:
@titusfortner Ok - building the current selenium-webdriver and testing with FF 49 gets Capybara down to 13 test failures - mostly sending keys to prompt modals (keySequence.join is not a function, missing mouse move action, sending "special" keys (:space & :backspace - ignored, shifted keys (the shift is not released), and extra key events with code of 0 being generated) that all used to work correctly.  The missing actions are definitely a geckodriver issue and I'm guessing the key issues probably are too.

twalpole:
@titusfortner Actually it appears the shifted keys issue is a selenium-webdriver issue.  Current versions of the webdriver spec appear to have removed the NULL resetting key modifiers that used to be in the spec [spec](https://www.w3.org/TR/2015/WD-webdriver-20150803/#sendkeys) .  It would appear selenium-webdriver needs to either stop accepting accepting nested arrays  ( send_keys([:shift, 'a'], "bcdef")  => Abcdef ) or split out any such passed arrays and send them to w3c compliant drivers separately - https://github.com/SeleniumHQ/selenium/blob/master/rb/lib/selenium/webdriver/common/keys.rb#L114

titusfortner:
@twalpole - 3.0.0.beta3 is now released with the latest updates
Simon suggested he'd follow up on this tomorrow as far as what the w3c spec should do for the send keys.

Speaking of Capybara tests. How do I run a single selenium test, I couldn't find documentation on it? I miss the ability to right click in Rubymine and run a single spec. :)

twalpole:
@titusfortner Because of the way most of the tests are included in the project so the other driver authors can run them against their own drivers the easiest way to run a single test is to tag it and then use the -t flag with rspec

    it "should accept the prompt if the message matches", focus: true do
      @session.accept_prompt 'Prompt opened', with: 'matched' do
        @session.click_link('Open prompt')
      end
      expect(@session).to have_xpath("//a[@id='open-prompt' and @response='matched']")
    end

    rspec -t focus

titusfortner:
@twalpole, Thanks! Yeah the Watir project currently uses a submodule for running watirspecs with different projects. Submodules are kind of a pain, though, so we are considering the option of something similar to what Capybara does. It seems to be substituting one type of pain for another, though. :)

titusfortner:
@twalpole - fwiw, getAttribute is down to 7KB in beta3.1 :)

twalpole:
@titusfortner that's much better :)

twalpole:
Closing in favor of the updated message that's now in the README

PR #1751, opened 2016-09-06
Linkify older Firefox releases in README

vassilevsky:
It was not obvious at all where to get them. Mozilla.com forces the latest version upon users.


twalpole:
thanks

PR #1763, opened 2016-09-26
support custom filter block in finders/matchers/assertions

twalpole:
Allow to pass a block that will be used to filter results once all other filters have passed.


abotalov:
Out of curiosity - why did you decide that it would be better to leave a warning rather than implement the same behaviour for `has_css?`, etc.?

twalpole:
@abotalov It was an intermediate step while I thought about whether it made sense to have on all the specialized matchers/finders

lib/capybara/node/matchers.rb
```diff
@@ -543,12 +433,123 @@ def has_table?(locator=nil, options={})
...
+          result = query.resolve_for(self.query_scope)
+          unless result.include? self
+            raise Capybara::ExpectationNotMet, "Item does not match the provided selector"
+          end
+        end
+        return true
+      end
+
+      def assert_not_matches_selector(*args, &optional_filter_block)
+        query = Capybara::Queries::MatchQuery.new(*args)
```
BrianHawley: `&optional_filter_block` isn't being passed through here.

lib/capybara/node/matchers.rb
```diff
@@ -543,12 +433,123 @@ def has_table?(locator=nil, options={})
...
+
+      ##
+      #
+      # Checks if the current node matches given CSS selector
+      #
+      # @param [String] css The CSS selector to match against the current code
+      # @return [Boolean]
+      #
+      def matches_css?(css, options={}, &optional_filter_block)
+        matches_selector?(:css, css, options)
```
BrianHawley: `&optional_filter_block` isn't being passed through here.

lib/capybara/node/matchers.rb
twalpole: Good catch - fixed through 06a8aa7c050af992ccf145354d609338a3f12fcd - thanks!

lib/capybara/node/matchers.rb
twalpole: Good catch - fixed through 06a8aa7 - thanks!

PR #1765, opened 2016-09-30
Select button element by descendant image alt text

elliterate:
As with links, button elements may consistent entirely of an image. In such cases, one may identify the button using the value of the descendant image's alt attribute (as a screen reader would).

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,8 @@
+#Edge
+
+### Changed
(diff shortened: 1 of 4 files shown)


twalpole:
@elliterate you need to use different text for your added button so it doesn't cause ambiguous issues with the other tests. Other than that this looks good.

twalpole:
Thanks!

PR #1772, opened 2016-10-10
Define the return type for actions where not otherwise defined

twalpole:
This defines the return type for node actions not otherwise defined as being the element being interacted with.  This allows for chaining some actions on elements, or checking their state, etc.

    element.click.send_keys(....)
    element.click.selected?


twalpole:
The return types of these methods were undefined before do defining them in a minor version release should be fine.

Issue #1773, opened 2016-10-13
Quitting a dead remote session raises a fatal exception


dylanlive:
## Meta
Capybara Version: 2.10.1 
Driver Information (and browser if relevant): selenium-webdriver 2.53.4, RSpec 3.4.0, Selenium Grid v.2.53.1

## Summary
When a remote Selenium Grid session is disconnected unexpectedly, it's difficult to recover.

A `Selenium::WebDriver::Error::UnknownError` exception is raised [upon exit](https://github.com/jnicklas/capybara/blob/master/lib/capybara/selenium/driver.rb#L26), even after clearing the session_pool and creating a new instance of the driver (full exception below).

A new driver instance can be created, but after the suite is finished, the old driver instance still tries to exit.

## Expected Behavior
When the test suite is finished running, at exit, it should not create an exception if the remote connection is already gone.

## Actual Behavior
An exception is triggered.

This prevents the user from recovering by creating a new driver, as the old driver raises an exception at exit.

## Steps to reproduce
1) Configure a remote driver to connect to Selenium Grid
2) Add this snippet to a test suite's spec_helper.rb so that the session_pool is cleared, and the example is retried (must be using rspec-retry to use run_with_retry...):
```
config.around(:each) do |example|
  example.run
  if example.exception
    Capybara.send(:session_pool).clear # Remove old session to recreate driver.
    example.run_with_retry # Re-run the example
  end
end
```

3) Let the test suite start running
4) Close the remote browser manually, or shut that node down
 - An exception will be raised, but the code snippet above will clear the session pool and reconnect to the remote

5) Don't interfere when it reconnects. Examples will pass, but at the very end of the suite, when exiting, this is raised:

```
Finished in 50.3 seconds (files took 0.42497 seconds to load)
6 examples, 0 failures

/Users/dylan/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/remote/response.rb:70:in `assert_ok': Error communicating with the remote browser. It may have died. (Selenium::WebDriver::Error::UnknownError)
Build info: version: '2.53.1', revision: 'a36b8b1', time: '2016-06-30 17:37:03'
System info: host: 'intentionally_removed', ip: 'intentionally_removed', os.name: 'Linux', os.arch: 'amd64', os.version: '3.13.0-92-generic', java.version: '1.8.0_91'
Driver info: driver.version: EventFiringWebDriver
[...]

PR #1774, opened 2016-10-13
Catch error for a dead remote connection upon quit

dylanlive:
Potential solution for #1773.

Upon quitting the browser, this catches a `Selenium::WebDriver::Error::UnknownError` exception, which is raised when there is a server-side issue communicating with the browser.

If the remote browser session is terminated for a server-side reason, Capybara will be able to assign `@browser` to nil upon quitting.

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -242,6 +242,8 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
     @browser.quit if @browser
   rescue Errno::ECONNREFUSED
     # Browser must have already gone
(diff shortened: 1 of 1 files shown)


twalpole:
As mentioned in Issue #1773 I'm reticent since I don't want to end up hiding actual errors people may need to know about -- Was there a message with the error you're receiving that could be used to narrow the errors that get ignored?

twalpole:
@dylanlive  Please try the selenium_quit_error branch/PR #1775 and see if that solves the issue you're having  -  Closing this in favor of the more limited error ignoring (assuming it works)

PR #1775, opened 2016-10-13
ignore another error while quitting selenium-webdriver

twalpole:
PR #1774 But checking for a specific message in the exception to limit the number of errors ignored

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,9 @@
+#Edge
+Release data: unreleased
+
+### Fixed
+* Ignore specific error when qutting selenium driver instance - Issue #1773 [Dylan Reichstadt\, Thomas Walpole]
+
 #2.10.1
 Release date: 2016-10-08
 
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -242,6 +242,9 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
     @browser.quit if @browser
   rescue Errno::ECONNREFUSED
     # Browser must have already gone
+  rescue Selenium::WebDriver::Error::UnknownError => e
+    raise unless e.message =~ /Error communicating with the remote browser/
+    # probably already gone
   ensure
     @browser = nil
   end


twalpole:
Potential fix for Issue #1773 - waiting for feedback from @dylanlive before merging

dylanlive:
@twalpole Tested and verified this will catch the case of a user closing the browser on Grid.

Unfortunately I don't believe it will catch other Grid connection issues, such as an SO_TIMEOUT. SO_TIMEOUT's also fall under `Selenium::WebDriver::Error::UnknownError`, but have a different exception message.

The easy solution would be to add another expected message to catch, but I fear the maintenance of string matching... what are your thoughts?

Exception:
```
[remote server] org.openqa.grid.internal.ActiveTestSessions(ActiveTestSessions.java):105:in `getExistingSession': Session [*] was terminated due to SO_TIMEOUT (org.openqa.grid.common.exception.GridException) (Selenium::WebDriver::Error::UnknownError)
```

For brevity of this comment, I made a gist containing the backtrace:
https://gist.github.com/dylanlive/df4645d54cdaa21c170c9fec1e62cc3a

twalpole:
@dylanlive  Ugh -- and this is the problem with UnknownError - I'm not interested in maintaining the strings and really don't want to just ignore all the errors - how often are you getting these errors?

dylanlive:
@twalpole I concur :P

Reasonable argument that people would want to know about it; however, the exception will likely still be raised during test execution (which can be caught in an rspec around block).

If the session was already terminated, does it make sense for quit to raise an exception? Likely the same argument could be used for `Errno::ECONNREFUSED`.

As for my experience with the SO_TIMEOUT's, they're occurring around some of our bandwidth throttling scenarios. Need to spend time researching prevention, but in the interim, was looking at recovery. The at_exit logic to quit is killing recovery efforts.

twalpole:
I've merged in the catching of the general 'Error communicating with the remote browser' error since it probably means the browser is gone and really isn't providing any useful info that could be used to diagnose any issue.   I'm not, as of now, comfortable doing the same with SO_TIMEOUT errors because I feel those would tend to indicate an issue between the client and the grid which should be diagnosed by the user

Issue #1776, opened 2016-10-13
Improve waiting for browser window to resize


pkmiec:
## Meta

This is a followup to https://code.google.com/p/selenium/issues/detail?id=7405.
Capybara Version:
2.10.1
Driver Information (and browser if relevant):
selenium-webdriver 2.53.4 with Firefox 47.0.1 on OS X Sierra
## Expected Behavior

Improve waiting for the window to resize. Here is how it could work for maximize (but similar technique could be applied to `resize_window_to`).

``` ruby
def maximize_window(handle)
  within_given_window(handle) do
    current_window = browser.manage.window
    current_window.maximize

    prev_window_size = nil
    "something similar to synchronize here which waits until Capybara.default_max_wait_time" do
      new_window_size = current_window.size
      return if new_window_size == prev_window_size

      prev_window_size = new_window_size
      sleep 0.05
    end
  end
end
```
## Actual Behavior

The current behavior (see below) waits for 0.1 seconds for the window to maximize. It's not clear whether the `sleep 0.1` is long enough. For example, on my giant monitor on os x, it is not. 

``` ruby
  def maximize_window(handle)
    within_given_window(handle) do
      browser.manage.window.maximize
    end
    sleep 0.1 # work around for https://code.google.com/p/selenium/issues/detail?id=7405
  end
```

twalpole:
@pkmiec The problem is there really is no way to know when maximize has finished.  Your method assumes the size has started changing within .05 seconds and will never spend more than .05 seconds without changing.  Those might be valid assumptions but I'm not convinced of that.  Since you're calling maximize you're obviously interested in some change that occurs when the window size changes, so the proper solution for this is to wait for that change.  What change are you expecting in your page when maximized and is it something you Capybara will wait for?

pkmiec:
You're right there is no way to tell for sure. I was just suggestion something which I think could work than the simple sleep 0.1.

twalpole:
I guess it wouldn't hurt to do something like this, although it should be implemented in Window#maximize not in the driver -  PRs welcome

abotalov:
Another issue with this is that window could already be maximized when `#maximize` is called.

twalpole:
@pkmiec Please give the window_size_change_wait branch a try and see if that would work for you

pkmiec:
@twalpole will do .. thanks

pkmiec:
@twalpole i tested it on my giant monitor and it works.

twalpole:
[...]

PR #1779, opened 2016-10-27
wait for window size changes to stabilize

twalpole:


twalpole:
Waiting for feedback from Issue #1776

Issue #1785, opened 2016-11-04
Broken exceptions with rails views


mluts:
Broken `ActionView::Template::Error` with rails v4.2.1

## Meta
Capybara Version: 2.10.1
Driver: Poltergeist

## Expected Behavior
Should see
```
ERROR["test_redirects_to_user_registration_page", UserInvitationTest, 5.526182861009147]
 test_redirects_to_user_registration_page#UserInvitationTest (5.53s)
ActionView::Template::Error:         ActionView::Template::Error: undefined method `ssn' for #<User:0x0055de54509ca8>
            app/views/users/invitations/edit.html.haml:29:in `block in _app_views_users_invitations_edit_html_haml___1320733798883780792_47206699358840'
            app/views/users/invitations/edit.html.haml:11:in `_app_views_users_invitations_edit_html_haml___1320733798883780792_47206699358840'
```

## Actual Behavior
Instead i see
```
ERROR["test_redirects_to_user_registration_page", UserInvitationTest, 2.2288646230008453]
 test_redirects_to_user_registration_page#UserInvitationTest (2.23s)
ArgumentError:         ArgumentError: wrong number of arguments (1 for 2)
            test/support/integration_tests.rb:28:in `block in <class:IntegrationTest>'
```
## Steps to reproduce

1. Write a rails view
2. Put a `raise "Boom!"` inside a view.

```
-# app/views/users/registrations/new.html.haml
- raise "Boom"
```

3. Write an integration test for this view

```
# test/integration/user/sign_up_test.rb
class IntegrationTest < ActionDispatch;:IntegrationTest
  include Capybara::DSL
  
  setup do
    Capybara.current_driver = :poltergeist
  end
  
   test 'sign-up form works' do
     visit new_user_registration_path
   end
end
```

4. Observe weird exception

```
ERROR["test_user_is_created_after_filling_in_sign-up_form", User::SignUptest, 2.349236459005624]
 test_user_is_created_after_filling_in_sign-up_form#User::SignUptest (2.35s)
ArgumentError:         ArgumentError: wrong number of arguments (1 for 2)
            test/integration/user/sign_up_test.rb:15:in `fill_in_input'
            test/integration/user/sign_up_test.rb:26:in `block (2 levels) in <class:SignUptest>'
            test/integration/user/sign_up_test.rb:25:in `block in <class:SignUptest>'
```

## Quickfix

```diff
[...]

PR #1786, opened 2016-11-08
raise the exception rather than the exception class - Fix Issue #1785

twalpole:


lib/capybara/session.rb
```diff
@@ -125,7 +125,8 @@ def raise_server_error!
         begin
           raise CapybaraError, "Your application server raised an error - It has been raised in your test code because Capybara.raise_server_errors == true"
         rescue CapybaraError
-          raise @server.error.class, @server.error.message, @server.error.backtrace
+          #needed to get the cuase set correctly in JRuby -- otherwise we could just do raise @server.error
```
RKushnir: Typo `cause`

twalpole: Thanks - fixed

PR #1787, opened 2016-11-09
Faster checking of checkbox by label

mrloop:
Speed up interactions with checkbox / radio where input is hidden but label is not by
* not waiting to find checkbox if not visible
* not waiting while trying to set value on checkout box `cbox.set` if not visible

Before this the default behaviour was to wait until the time out before checking if could interact via the label.

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -115,18 +115,17 @@ module Capybara
         locator, options = nil, locator if locator.is_a? Hash
         allow_label_click = options.delete(:allow_label_click) { Capybara.automatic_label_click }
 
-        begin
-          radio = find(:radio_button, locator, options)
-          radio.set(true)
-        rescue => e
-          raise unless allow_label_click && catch_error?(e)
+        radio = find(:radio_button, locator, options.merge({visible: (allow_label_click ? :all : true)}))
+
+        if !radio.visible? && allow_label_click
           begin
-            radio ||= find(:radio_button, locator, options.merge({wait: 0, visible: :all}))
             label = find(:label, for: radio, wait: 0, visible: true)
             label.click unless radio.checked?
           rescue
-            raise e
+            find(:radio_button, locator, options.merge({wait: 0})).set(true)
           end
+        else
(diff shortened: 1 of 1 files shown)


twalpole:
This speedup would be nice to have however in the current form it will attempt to call set on a non-visible element if allow_label_click is false (default) which is not acceptable.  Also it may be better to change this so all the logic is inside a synchronize block so the whole thing gets retried rather than worrying about the wait time on each separate call.

mrloop:
Thanks for the quick review @twalpole 

>in the current form it will attempt to call set on a non-visible element if allow_label_click is false (default) which is not acceptable.

Where `allow_label_click` is false the implemention in this PR will raise at [line 156](https://github.com/mrloop/capybara/blob/d8d57df8bef11ddf674e3a6f9f5a89b6fd576acb/lib/capybara/node/actions.rb#L156)
`cbox = find(:checkbox, locator, options.merge({visible: (allow_label_click ? :all : true)}))`

This scenario is tested for [here](https://github.com/mrloop/capybara/blob/d8d57df8bef11ddf674e3a6f9f5a89b6fd576acb/lib/capybara/spec/session/check_spec.rb#L160) , maybe the ternary operator doesn't help readability, also not sure if it's in keeping with the project style. Let me know if it's acceptable in it's current form or if it needs changed.

twalpole:
You are correct, my mistake.  So really the only behaviors that would be changed are the the error message that is raised if the wait time is mentioned in the error message and what happens if the user passes a `visible` option in.  An alternate implementation that would provide similar results is PR #1789 but without changing the current behavior at all.

mrloop:
@twalpole great I'll try out #1789 with test suite tomorrow. (UTC+00:00)

twalpole:
@mrloop Thanks -- let me know if it behaves basically equivalent to this PR and then I'll have to decide which way to go.

mrloop:
@twalpole #1789 working as expected. I'll close this PR in favour of yours. Thanks for coming up with better solution and I now know capybara internals a little bit better :)

PR #1795, opened 2016-11-17
Add `:last` match strategy

booch:
I ran into a situation where I've got multiple matching nodes for a query, and always need to pick the last one. I added this as a patch to my tests, and figured I'd clean it up for upstream, as others might find it useful.

### DIFF
--- a/History.md
+++ b/History.md
@@ -5,6 +5,7 @@ Release date: unreleased
 * Options for clearing session/local storage on reset added to the Selenium driver
 * Window size changes wait for the size to stabilize
 * Defined return value for most actions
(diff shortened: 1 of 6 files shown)


twalpole:
Thanks for the effort, however I don't really want to add any more matching strategies currently - they're confusing enough for people already.  For ones own use in a project rather than patching in a last strategy I think I'd instead write a method to include in the features - something like

    def last(*args, **options, , &optional_filter_block)
        options = { minimum: 1 }.merge(options) # force waiting behavior
        all(*args, **options, &optional_filter_block).last
    end

PR #1798, opened 2016-11-22
Ignore connection error when quitting

psyho:
It looks like the message for marionette is different than for the old version. Because of this, when Firefox shuts down too quickly, the "unknown error" thrown by the driver will make the entire RSpec test suite fail.

The issue can be reproduced using:
* capybara 2.10.1
* selenium-webdriver 3.0.1
* geckodriver 0.11.1
* firefox 49.0.1
* Ubuntu 14.04 (tested on both Ubuntu and Mac OS, but the issue only appears on Ubuntu).

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -254,12 +254,15 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
     message
   end
 
+  CONNECTION_ERROR_TEXT_FIREFOX = /Error communicating with the remote browser/
+  CONNECTION_ERROR_TEXT_MARIONETTE = /Failed to decode response from marionette/
+
   def quit
     @browser.quit if @browser
   rescue Errno::ECONNREFUSED
     # Browser must have already gone
   rescue Selenium::WebDriver::Error::UnknownError => e
-    raise unless e.message =~ /Error communicating with the remote browser/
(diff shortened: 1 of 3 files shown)


twalpole:
How is your firefox being shut down before the driver calls quit?  This really seems like something that should be fixed in selenium/geckodriver rather than us continuing to add more and more strings for each different browser selenium supports.

psyho:
@twalpole I haven't debugged how it happens. `quit` is being called by Capybara and it raises an exception. I noticed that the same exception, with a different message is being handled already in Capybara code, so I updated the code to preserve the same behaviour against the latest versions of Firefox/geckodriver. All I can add is that Firefox is shut down without any processes hanging around after the test suite finishes, same goes for the geckodriver process.

twalpole:
The problem is that "Failed to decode response from marionette" doesn't mean the same thing as "can't communicate" to me (one of them comes from a selenium grid, one from geckodriver I think) - and the fact that it's being returned in an `UnknownError` makes it tough to say what it means and whether it should be ignored or not. I also don't want to start adding constants for every message we're going to ignore.   I guess it probably doesn't matter what the message is in an UnknownError during quit we could just warn out the message and continue on.  If you want to change this to just do something like `warn "Ignoring unknown Selenium error: #{e.message}"` and get rid of the tests you added I'd merge that in.

twalpole:
Closing this in favor of PR #1799

PR #1801, opened 2016-11-30
Prevent bad uri errors

mduqueoviedo:
When using the gem [capybara-screenshot](https://github.com/mattheworiordan/capybara-screenshot), sometimes I get a BadURI error, because the current URL of the application has the following form:

`http://someNormalUrl/some_normal_path/some_id##other_stuff` (Note the two hashes)

When there is this type of url capybara throws the error and the screenshot is not saved, which is bad because we want the screenshot to understand what is going on. It still shows the cause of the error and everything else runs normally, but with no screenshot.

This fix prevents the badURI error and allows to save the screenshot. All tests are passing as well, but I am honestly not sure wether this can have more impact on the execution of capybara.

### DIFF
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -168,7 +168,8 @@ module Capybara
     # @return [String] Path of the current page, without any domain information
     #
     def current_path
-      path = URI.parse(current_url).path
+      encoded_url = URI.encode(current_url)
+      path = URI.parse(encoded_url).path
       path if path and not path.empty?
     end
 


altmer:
Yep, this is happening really often. Looking forward to merging of this pull request.

twalpole:
Randomly encoding a URL is not the solution to this issue, and will just end up breaking path marchers etc in people's code.  The proper solution is to figure out why the driver is returning an invalid URL from current_url.  What driver are you using?

mduqueoviedo:
I am using Poltergeist as driver. 

I'd gladly fix this in my code or in the capybara-screenshot gem instead, but the issue raises when calling this method, it is so deep in execution that I had to put the fix here.

twalpole:
it sounds like it's something that should be fixed in Poltergeist so it's current_url method returns a valid URL.  That will need tests for different urls to make sure it never double encodes, etc

twalpole:
Looking at this further - it appears selenium returns the same invalid URL if you have invalid urls in links, etc.  You can't visit those urls through Capybara though because they are technically invalid and should be http://someNormalUrl/some_normal_path/some_id#%23other_stuff to be valid.  Calling encode however turns it into http://someNormalUrl/some_normal_path/some_id%23%23other_stuff which is also not correct.   The easiest solution here may be to use Addressable::URI.parse in this case which uses a much more lenient parsing algorithm  -  Please try the `current_path_parse` branch and see if that fixes your issue

twalpole:
@mduqueoviedo What framework are you using that ends up producing technically invalid URLs?

twalpole:
For completeness the spec that covers this is https://tools.ietf.org/html/rfc3986#section-3.5  where 

fragment         = *( pchar / "/" / "?" )
pchar              = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved     = ALPHA / DIGIT / "-" / "." / "_" / "~"
pct-encoded  = "%" HEXDIG HEXDIG
sub-delims     = "!" / "$" / "&" / "'" / "(" / ")"  / "*" / "+" / "," / ";" / "="

twalpole:
Closing in favor of 95173f04c3f5a8984949e738ebafaf530479257b now merged into master

mduqueoviedo:
Awesome, thanks!

PR #1804, opened 2016-12-14
save_page saves the CSS and images

zedtux:
This commit injects the CSS as internal, and the images as Base64 allowing to have an HTML screenshot looking much better.

Regarding the code, I don't like to pass the session as parameter to the helper methods, but it was even worst to me to have those methods in the `Capybara::Session` class. I was tempted to create a `Capybara::HtmlTransformer` class which would embed this code and called from the `Capybara::Session`.

Can someone (@twalpole maybe?) tells me the way I should implement this better (if needed)?

### DIFF
--- a/lib/capybara/helpers.rb
+++ b/lib/capybara/helpers.rb
@@ -34,6 +34,21 @@ module Capybara
 
     ##
     #
+    # Updates the HTML before writing it to the temp file.
+    #
+    # @param [String] html                HTML code to inject into
+    # @param [Capybara::Session] session  current Capybara session
+    # @return [String]                    The modified HTML code
+    #
+    def finalise_html(html, session)
+      html = inject_asset_host(html)
+      html = inject_css_as_internal(html, session)
+      html = inject_html_images_as_base64(html, session)
+      inject_css_images_as_base64(html, session)
+    end
+
+    ##
+    #
     # Injects a `<base>` tag into the given HTML code, pointing to
     # `Capybara.asset_host`.
     #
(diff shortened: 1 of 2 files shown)


zedtux:
I will check and fix the tests soon.

zedtux:
Also I was thinking of adding an option in order to enabled/disable this feature

twalpole:
Hi - thanks for the efforts, however the point of saving the page is to look at the HTML as your app rendered it, so drastically changing the page isn't really something I consider desirable.  We do provide the `asset_host` setting which can be set so you can load up the saved page and load the assets from a different base url if wanted, and if you really want to see what the page looked like in the browser there is `save_screenshot`.  The functionality you're proposing may be better as a separate gem.

zedtux:
Thank you @twalpole  for your fast reply.


> [...] changing the page isn't really something I consider desirable

This is the reason why in my last comment I suggested to add an option to enable/disable it.

> We do provide the asset_host setting which can be set so you can load up the saved page and load the assets from a different base url ...

This option require to have a running server, where my feature doesn't, which ease the life of developers.

> [...] if you really want to see what the page looked like in the browser there is save_screenshot.

As far as I know, this requires to have a JavaScript driver which is slow. (Correct me if I'm wrong).
Also this prevent developers to inspect the page and see the CSS rules for example.

The added code is really small, and with an option to enable it (disabled by default) would give us the best of both worlds.

twalpole:
Yes - it requires a running server, but that's easy enough to setup and means the page is loading the same way it would in production.
Yes - screenshot requires a driver that supports taking screenshots, and that is only the JS supporting drivers, however with the amount of JS/CSS used in apps today the rack-test driver really is very limited and most people are using one of the JS drivers.

I still think this feature is one that would be best implemented as a separate gem that does something like add a 'save_finalized_page` method to Capybara::Session.

zedtux:
Okay, thank you, I'll check with the `capyabara-screenshot` gem if they accept my code (I have the feeling to be the guy looking for friends 😄, or a kind of Repo-less).

twalpole:
@zedtux Why don't you create your own gem -  name it  capybara-finalize-page or something and release that, rather than trying to get it included into another gem?

zedtux:
Just to avoid people having n amount of gems to be installed for just 20 lines of code. But in the end, if event `capybara-screenshot` doesn't want my code, I will do a gem for sure :)

PR #1807, opened 2016-12-16
Integer Unification for Ruby 2.4.0+

koic:
Ruby 2.4.0 unifies Fixnum and Bignum into Integer.

https://bugs.ruby-lang.org/issues/12005

Just modified `Fixnum` to `Integer`. As a result of that will fix `warning: constant ::Fixnum is deprecated` warning in Ruby 2.4.0+.

Thanks.


twalpole:
LGTM - Thanks

Issue #1248, opened 2014-02-17
Thread safety


jnicklas:
Capybara is not thread safe in that we rely on global configuration which is liable to change at any point. Unfortunately, fixing this is probably going to break APIs in a few subtle ways. Changing a global configuration parameter is not going to affect an already existing session, since we'd need to copy configuration to sessions to prevent concurrent modification, as well as providing an API to change options per session. I think this is consistent with how _most_ people use Capybara, but it's going to break for some people, so I want to do this for 3.0.

timlentse:
Is this ok?
```ruby
5.times do |id|
    workers << Thread.new do
    session[id].visit(url)
   end
end
```
Here `session` is an array of session instance, I have tried, but it turned out to be dead lock. I don't know why?

oystersauce8:
is there a branch or a patch that makes this possible?  When is capybara 3 coming out.

twalpole:
There is no branch or patch currently and not current timeline for Capybara 3

twalpole:
The largest part of this (the per session configuration) is being worked on in the `session-config` branch/PR #1808  if anyone has any feedback/suggestions

twalpole:
Closed via 0270b7a4cf960055a87f93db1322b53204aec350

PR #1808, opened 2016-12-16
Allow for per-session configuration of relevant global options for "thread safety"

twalpole:
As the largest part of Issue #1248 this allows for per session configuration of the relevant global options.  The feature is hidden behind the `Capybara.per_session_configuration` swtich.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ You can read more about the missing features [here](https://github.com/teamcapyb
 - [Beware the XPath // trap](#beware-the-xpath--trap)
 - [Configuring and adding drivers](#configuring-and-adding-drivers)
 - [Gotchas:](#gotchas)
+- ["Threadsafe" mode](#threadsafe)
 - [Development](#development)
 
 ## <a name="key-benefits"></a>Key benefits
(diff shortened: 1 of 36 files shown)


twalpole:
@abotalov Any feedback on this?

twalpole:
@jnicklas Any feedback on this?

ronakjain90:
Hi, I can tell you that this branched fixed one of our production problem with respect to capybara running in thread-safe mode.

ronakjain90:
We were having an issue of Capybara not being thread safe. i.,e we were using multiple threads, and each thread used to access the same global level page configuration giving us faulty results. After switching to this branch and using the Thread Safe config, Capybara is threadsafe now. Also creating individual sessions for each thread gives us flexibility to customise settings for each thread. This PR is awesome and solves a lot of production problems for us.

twalpole:
@ronakjain90 Ok -- so you are actually using Capybara in your app, as opposed to just for running tests, and manually creating sessions, etc.  Is there anything that would make the current API better for you for that use.  Just trying to figure out how much more to do along this line (and priorities) before merging into master, since I don't want to be revamping/deprecating parts of it immediately after release.

ronakjain90:
Yes. Capbara is used as a core module rather than just being used to run test cases. Regarding any suggestion for improvements, I think what we have in this PR is great. I didn't have to do any patching or customisation to make it work. It has all the features necessary to go straight to production. Thank you so much for this Thomas.

Issue #1809, opened 2016-12-18
Non-empty about:blank prevents Capybara from resetting session


wildlyinaccurate:
## Meta
Capybara Version: 2.2.1, 2.11.0
Driver Information (and browser if relevant): selenium-webdriver 3.0.3 using ChromeDriver 2.26.436421 (6c1a3ab469ad86fd49c8d97ede4a6b96a49ca5f6) to drive Chrome 55
Other: Tests are running with Cucumber 2.4.0

## Expected Behavior

Capybara should reset the session after each scenario with no issues.

## Actual Behavior

When JavaScript is disabled, ChromeDriver appends an iframe to the about:blank page's body, which causes Capybara's [empty page check](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selenium/driver.rb#L136) to fail:

```html
<html><head></head><body><iframe name="chromedriver dummy frame" src="about:blank"></iframe></body></html>
```

**2.11.0** Chrome navigates to about:blank, and after 10 seconds Capybara raises `Timed out waiting for Selenium session reset (Capybara::ExpectationNotMet)`

**2.2.1** as soon as Chrome navigates to about:blank, Capybara raises `expected not to find xpath "/html/body/*", found 1 match: "" (Capybara::ExpectationNotMet)`

**2.2.0** works fine because it uses an empty HTML file instead of about:blank

## Steps to reproduce

With the following files, run `bundle install && bundle exec cucumber`

```ruby
# Gemfile
source 'http://rubygems.org'

gem 'cucumber', '~> 2.4.0'
gem 'capybara', '~> 2.11.0'
gem 'selenium-webdriver',  '~> 3.0.0'
```

```ruby
# features/support/env.rb
require 'capybara/cucumber'

Capybara.default_driver = :chrome_no_js
Capybara.register_driver :chrome_no_js do |app|
  Capybara::Selenium::Driver.new(
    app,
    :browser => :chrome,
    :prefs => {
      :profile => {
        :default_content_setting_values => { :javascript => 2 }
      }
    }
  )
end

When('I open BBC News') { visit 'http://www.bbc.com/news' }
```

```cucumber
# features/test.feature
Feature: iframe shenanigans

Scenario: the test
  When I open BBC News
```

## Screenshot

![screen shot 2016-12-18 at 22 13 13](https://cloud.githubusercontent.com/assets/794263/21292819/00c7fbae-c576-11e6-98de-b07db1e3ad42.png)

[...]

PR #1810, opened 2016-12-19
Ignore extra dummy frame on page for SeleniumDriver#reset!

twalpole:
Attempt to fix Issue #1809.  This allows for chrome having an extra iframe on the about:blank page when JS is disabled - not sure why and long term should probably be something fixed by chrome/chromedriver

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,9 @@ Release date: unreleased
 ### Added
 * Session#switch_to_frame for manually handling frame switching - Issue #1365 [Thomas Walpole]
 
+### Fixed
+* SeleniumDriver#reset! allows dummy frame with chrome when JS is disabled - Issue #1809
+
 #Version 2.11.0
 Release date: 2016-12-05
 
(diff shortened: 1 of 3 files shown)


abotalov:
I googled a bit, and it seems this frame might be added by Chromedriver itself - https://chromium.googlesource.com/chromium/src/+/lkcr/chrome/test/chromedriver/chrome/navigation_tracker.cc
But it seems it should be removed quickly. Code contains the following:
```js
document.body.appendChild(frame);
window.setTimeout(function() {
  document.body.removeChild(frame);
}, 0);
```

twalpole:
@abotalov Definitely seems that it's not removed.  This PR doesn't really do anything that will break things so it should be safe to merge, just waiting for a response from the OP of Issue #1809 to make sure if fixed the issue for them.  There's also the question of whether or not there's much use in trying to use Capybara with Chrome when javascript is disabled -- Haven't really tried anything to find out.

twalpole:
This went along with Issue #1809 - Closing due to lack of feedback

Issue #1662, opened 2016-03-05
'execute_script' should accept optional '*args' attribute.


tommyh:
Hi,

First off - I've been using capybara for a few years now and i love it.  So thanks for all the hard work.  :)

Feature request: `execute_script` should accept optional `*args` attribute.

Note: I'm more than happy to work on a pull request for this feature, but wanted to open an issue first to discuss to see if this had already been thought of and not implemented for some reason.

Use case: I am using site_prism as my page object model and sometimes I want to execute javascript on one of the selenium elements which I have in my object model.

So instead of writing:

```ruby
execute_script("$('.login').fooBar();")
```

I could write:

```ruby
execute_script("$(arguments[0]).fooBar();", @home.login_bar.native)
```

(where `@home.login_bar` is my site_prism element)

One reason I can come up with why capybara's execute_script only takes one argument, and selenium webdriver takes 2 arguments (js and elements), is that either of the other drivers (poltergiest, capybara-webkit) only support the one js argument.  But I haven't invested the time to see if it's possible in the other drivers under the covers or they just chose to leave it out in the public ruby api as well.

Right now I am simply doing:

```ruby
page.driver.browser.execute_script("$(arguments[0]).fooBar();", @home.login_bar.native)
```

But it would be more convenient if I didn't have to think about the interface difference.  And it took me a far amount of googling/experimentation to even get that to work, so I think capybara supporting it (even if for supported drivers) would be helpful to others.

Thanks,
Tom

twalpole:
This has been discussed before, and as you surmised the issue is whether or not it's possible to pass element references to the browser in the other available drivers.  It requires elements having a unique id that can be sent back and forth.  When this was previously discussed it wasn't possible in at least one other driver available at the time, not sure if that has changed.   We try not to add functionality that can't be implemented in other drivers because that ends up being just as bad as calling methods through page.driver (from a cross driver test usage perspective) but with the added benefit of not being able to tell it's not cross driver compatible

tommyh:
Gotcha, I've confirmed that the other drivers don't have it in their public api, and I'm not motivated enough to chase down if the unique id requirement has been improved.

Glad you gave me this feedback before I invested time in a pull request.  :)

Now that I think of it, I might try monkey patching this behavior in our project.  If I do that, I'll post that code on this issue if it would help others.

Thanks again,
Tom

tommyh:
[...]

Issue #1297, opened 2014-04-22
A discussion about adding *args option to execute_script


abotalov:
I don't have permissions to reopen #938 so I start a new issue to discuss it further.

It seems that primary reason why that PR was rejected was that it was hard to make options hash working in some drivers.

But there is already similar situation - `options` argument of `save_screenshot`. Different drivers have different possible values of those options:
https://github.com/thoughtbot/capybara-webkit/blob/master/lib/capybara/webkit/driver.rb#L162
https://github.com/teampoltergeist/poltergeist/blob/master/lib/capybara/poltergeist/browser.rb#L179

PR #1812, opened 2016-12-23
Support optional passing of arguments to execute/evaluate_script

twalpole:
Let `execute_script` and `evaluate_script` optionally take arguments that will get passed to the scripts as `arguments`.  Selenium supports passing elements as arguments too.

### DIFF
--- a/lib/capybara/driver/base.rb
+++ b/lib/capybara/driver/base.rb
@@ -28,11 +28,11 @@ class Capybara::Driver::Base
     raise Capybara::NotSupportedByDriverError, 'Capybara::Driver::Base#go_forward'
   end
 
-  def execute_script(script)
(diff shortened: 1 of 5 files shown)


abotalov:
By the way this feature was rejected several times before: https://github.com/teamcapybara/capybara/issues/1662, https://github.com/teamcapybara/capybara/issues/1297, https://github.com/teamcapybara/capybara/pull/938

twalpole:
@abotalov Yes it was, because it was reported to not be possible (or very difficult) in other drivers.  What has changed is that I went ahead and implemented it for Poltergeist - https://github.com/teampoltergeist/poltergeist/pull/838 - so it's not a Selenium specific feature anymore

PR #1813, opened 2016-12-23
Support easier full text matching

twalpole:
This adds support for `exact_text` option on selector queries, and an 'exact' option on title and text queries.  The thought is that an `exact` option always affects the primary function of the query - so the title in title queries, text in text queries, locator in selector queries.  The `exact_text` option on selector queries can be true/false in which case it will affect the `text` option, or it can be a string which is a shortcut for passing that string as the `text` option and also passing `exact_text: true`.   

The one issue of "inconsistency" is that the text query `exact` option defaults to Capybara.exact_text, but this is documented and I don't consider it a huge issue for a user to understand - and makes more sense than it defaulting to Capybara.exact from a usability pov.

Another issue is the `exact` option for title queries currently defaults to false, but will be changed in v3 to default to Capybara.exact

### DIFF
--- a/lib/capybara/node/document_matchers.rb
+++ b/lib/capybara/node/document_matchers.rb
@@ -7,10 +7,11 @@ module Capybara
       #
       # @!macro title_query_params
       #   @overload $0(string, options = {})
-      #     @param string [String]           The string that title should include
+      #     @param string [String]           The string that title should include      #
       #   @overload $0(regexp, options = {})
       #     @param regexp [Regexp]           The regexp that title should match to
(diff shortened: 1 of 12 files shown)


lib/capybara/node/document_matchers.rb
```diff
@@ -7,10 +7,11 @@ module DocumentMatchers
       #
       # @!macro title_query_params
       #   @overload $0(string, options = {})
-      #     @param string [String]           The string that title should include
+      #     @param string [String]           The string that title should include      #
```
abotalov: Is `#` at the end a typo?

twalpole: Looks like it - fixed

PR #1814, opened 2016-12-28
warn when Capybara DSL methods are shadowed

twalpole:
This PR attempts to issue a warning whenever a class that has included Capybara::DSL includes another module that shadows some of the Capybara::DSL methods.  If you are expecting the methods to be shadowed they can be added to the array Capybara.expected_shadowed_dsl_methods to suppress the warning, or set it to nil to disable completely.  Specifically this is an attempt to notify users of the case where RSpec::Matchers  shadows `within` which leads to the block being passed to it being completely ignored but not necessarily raising an error.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -31,6 +31,7 @@ module Capybara
     attr_reader :save_and_open_page_path
     attr_accessor :exact_text
     attr_accessor :app
+    attr_accessor :expected_shadowed_dsl_methods
 
     ##
     #
@@ -504,6 +505,7 @@ Capybara.configure do |config|
   config.automatic_label_click = false
   config.enable_aria_label = false
   config.reuse_server = true
+  config.expected_shadowed_dsl_methods = []
 end
 
 Capybara.register_driver :rack_test do |app|
--- a/lib/capybara/dsl.rb
+++ b/lib/capybara/dsl.rb
@@ -7,7 +7,7 @@ module Capybara
       warn "including Capybara::DSL in the global scope is not recommended!" if base == Object
       base.define_singleton_method :include do |*args|
         if !Capybara.expected_shadowed_dsl_methods.nil?
-          args.select { |arg| !include?(arg) }.each do |arg|
+          args.reject { |arg| include?(arg) || arg.to_s == "Capybara::Minitest::Assertions" }.each do |arg|
             conflicts = (arg.instance_methods & Capybara::Session::DSL_METHODS)- Capybara.expected_shadowed_dsl_methods
             warn "Capybara::DSL methods #{conflicts} are shadowed by #{arg}. "\
                  "If you expected this please add the method names to Capybara.expected_shadowed_dsl_methods and "\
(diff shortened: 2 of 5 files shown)


abotalov:
IMO this alone might be not a quiet good solution to the underlying issue.
User should be able to use both Capybara's and RSpec-Expectations APIs. `all`/`within` have good implementations in both libraries.

Let's see what ways we have:
* User includes RSpec::Matchers after Capybara::DSL. Here we can:
  - Recommend to use `all` and `within` via session object (page)
  - Add `within_element` method as alias for within and replace `all` with `find_all` / `within` with `within_element` in Capybara's docs.
* User includes Capybara::DSL after RSpec::Matchers. Here we can:
  - Recommend to alias `all` / `within` rspec matchers using `alias_matcher`. One may try to add these aliases into RSpec itself.
  - Implement blocks suggested in https://github.com/rspec/rspec-expectations/issues/820 in RSpec-Expectations and recommend to access shadowed rspec methods using such blocks

Regarding your point in https://github.com/rspec/rspec-expectations/pull/960:
`claim RSpec owns the within method name is ludicrous` - [one may say that its RSpec that owns the namespace](https://github.com/teamcapybara/capybara/issues/1396#issuecomment-57104345). I agree that it would have been better if RSpec had chosen names that don't collide with Capybara's methods. But now both libraries own their methods and can't just remove/rename them.

Going forward, I think it might be good [to have build checks for such clashes](https://github.com/teamcapybara/capybara/issues/1396#issuecomment-57098125) so there would be less chance of new colliding methods.

cc @JonRowe

twalpole:
@abotalov The issue this is proposed to solve is really the `within` issue when RSpecs matchers override Capybaras version of the method.  In all the other cases of `all` and `within` the user ends up with some sort of error which allows them to in some way work out that it's calling the wrong method.  With RSpecs `within` overiding Capybaras though it just ends up with the test silently ignoring a block of code

    within(element) do
      expect(page).to have_content('Some test')  # ends up not being run if RSpecs `within` has shadowed Capybara's 
    end

which leads to false passing of tests (the worst possible outcome)

Since RSpecs `within` matcher doesn't take a block I proposed having RSpec flag that as an error which would at least have warned users - but that was not acceptable.  I don't have time to continue proposing other changes to RSpec that may or may not be acceptable, and the maintainers of RSpec don't seem to really consider it to be an issue that needs flagging,  so I'm looking to do anything I can purely in Capybara to at least give users notice that things may not be as expected.  Adding `within_element` is definitely a possibility but, as you said, doesn't solve usage in existing tests.

axlekb:
This is great!  We were just running into false positive `within` problems because we included `RSpec::Matchers` after `Capybara::DSL`.  I didn't spend enough time figuring out why it only seemed to happen sometimes, but having a warning would have definitely saved some headaches!

twalpole:
@axlekb Yeah -- it's still being considered, want to make sure it doesn't potentially create issues for people who aren't having the problem to start with.

twalpole:
Closing in favor of PR #1850

PR #1815, opened 2016-12-29
Add docs about Capybara.save_path and #save_screenshot

strika:
Hope this will save few minutes for other developers. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,6 +672,10 @@ Or have it save and automatically open:
 save_and_open_screenshot
 ```
 
+Screenshots are saved to `Capybara.save_path`, relative to the app directory.
+If you have required `capybara/rails`, `Capybara.save_path` will default to
+`tmp/capybara`.
+
 ## <a name="matching"></a>Matching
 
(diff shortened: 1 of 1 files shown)


twalpole:
This is only true if you've required capybara/rails and not set anything else.

strika:
I see. Any suggestions how a developer can know where a screenshot is going to be saved? (Without digging into the [code](https://github.com/teamcapybara/capybara/blob/1730fd91e7a7e6c02a9492682c595bceb2448648/lib/capybara/session.rb#L821-L829)).

twalpole:
@strika Rather than saying "If the application that you are testing is a Rails app" say something like "If you have required capybara/rails Capybara.save_path will default to ..."

strika:
Thanks. I updated the PR.

Is it a problem that this doesn't cover all cases (e.g. it doesn't mention `Capybara.save_and_open_page_path`)?

twalpole:
@strika save_and_open_page_path is deprecated so it doesn't matter

twalpole:
Thanks

PR #1818, opened 2016-12-30
Add an option for attach_file to allow modifying the elements style s…

twalpole:
…ince they are so often hidden on the page  -  This adds a :style option to the attach_file method which allows for modifying the file inputs CSS before attempting to attach the file.  This is because it's pretty standard for a file input to be hidden in some way for styling reasons, and attach_file requires the input to be visible to attach the file with no other way of interacting with it.  This just prevents people from having to add their own code that modifies style before every attach_file call.  It requires driver support for argument passing of elements to execute_script.

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -230,7 +230,7 @@ module Capybara
       # @option options [String] id             Match fields that match the id attribute
       # @option options [String] name           Match fields that match the name attribute
       # @option options [String, Array<String>] :class    Match links that match the class(es) provided
-      # @option options [Hash] style   A Hash of CSS styles to change before attempting to attach the file
+      # @option options [Hash] style   A Hash of CSS styles to change before attempting to attach the file (may not be supported by all driver)
       #
       # @return [Capybara::Node::Element]  The file field element
       def attach_file(locator, path, options={})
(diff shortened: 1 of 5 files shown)


abotalov:
Do you think one may misunderstand `:style` option and think that it finds an element with a given style? (as majority of other `Capybara::Node::Actions` options are shared with `Finders`)

Do you think that it would be better to rename it to e.g. `set_style`?

twalpole:
@abotalov Yeah - I originally had it as :style_change but didn't like that -- I'm not sure I like `preset_style`, maybe `set_style` -  I'll think about it

strika:
Is the only reason to have `style` option to make the input field visible? If so, is there a way to implement an option to make the element visible (and hide the details from users) without setting style explicitly? For example:

```ruby
attach_file(avatar, file, :make_visible => true)
```

twalpole:
@strika - no there isn't due to the many ways an element can be hidden

strika:
OK. Thanks.

twalpole:
With a night of sleep on this I like the idea put forward by @strika, with a minor extension.  We could have a `make_visible: true` option that defaults to setting something like `opacity: 1, display: block`  - and if that doesn't work for a specific users site they can set it to a hash of settings to change.

twalpole:
See PR #1823

PR #1820, opened 2017-01-02
Add assert_all_of_selectors and assert_none_of_selectors

twalpole:
Add assertions to match all/none of a group locators applied to a specific selector type.

### DIFF
--- a/lib/capybara/node/matchers.rb
+++ b/lib/capybara/node/matchers.rb
@@ -97,6 +97,58 @@ module Capybara
         end
       end
 
+      # Asserts that all of the provided selectors are present on the given page
+      # or descendants of the current node.  If options are provided, the assertion
+      # will check that each locator is present with those options as well (other than :wait).
+      #
+      #   page.assert_all_of_selectors(:custom, 'Tom', 'Joe', visible: all)
+      #   page.assert_all_of_selectors(:css, '#my_div', 'a.not_clicked')
+      #
+      # It accepts all options that {Capybara::Node::Finders#all} accepts,
+      # such as :text and :visible.
+      #
+      # The :wait option applies to all of the selectors as a group, so all of the locators must be present
+      # within :wait (Defaults to Capybara.default_max_wait_time) seconds.
+      #
+      # @overload assert_all_of_selectors([kind = Capybara.default_selector], *locators, options = {})
+      #
+      def assert_all_of_selectors(*args, &optional_filter_block)
+        options = if args.last.is_a?(Hash) then args.pop.dup else {} end
+        selector = if args.first.is_a?(Symbol) then args.shift else Capybara.default_selector end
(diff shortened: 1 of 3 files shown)


lib/capybara/node/matchers.rb
```diff
@@ -97,6 +97,58 @@ def assert_selector(*args, &optional_filter_block)
...
+      #
+      # @overload assert_none_of_selectors([kind = Capybara.default_selector], *locators, options = {})
+      #
+      def assert_none_of_selectors(*args, &optional_filter_block)
+        options = if args.last.is_a?(Hash) then args.pop.dup else {} end
+        selector = if args.first.is_a?(Symbol) then args.shift else Capybara.default_selector end
+        wait = options.fetch(:wait, Capybara.default_max_wait_time)
+        synchronize(wait) do
+          args.each do |locator|
+            assert_no_selector(selector, locator, options, &optional_filter_block)
```
simi: I was thinking about "DRYzing" those methods, since this is only different line, but I wasn't able to find out nice way how to do that. :(

Maybe something like: :question: 

```ruby
def assert_none_of_selectors
  _for_all_selectors(args) do |selector, locator, options|
    assert_no_selector(selector, locator, options, &optional_filter_block)
  end
end

*That will probably need to use original rails implementation wait per assert.*

@twalpole feel free to ping me if you think this is way to go. I can provide needed changes.

Comment (twalpole) [inline lib/capybara/node/matchers.rb]: @simi The methods are small enough that I'm okay with the repetition in the current version.  Capybara 3.0 will require ruby 2.2+ which means that a lot of the parameter extraction code can go away with the better argument options- and I much prefer the wait being shared with the group since that's what most people will actually want.

Comment (simi) [inline lib/capybara/node/matchers.rb]: OK, that makes sense.

I had also one more idea:

```ruby
def assert_selectors(assertion, *args, &optional_filter_block)
  options = if args.last.is_a?(Hash) then args.pop.dup else {} end
  selector = if args.first.is_a?(Symbol) then args.shift else Capybara.default_selector end
  wait = options.fetch(:wait, Capybara.default_max_wait_time)
  synchronize(wait) do
    args.each do |locator|
      send(assertion, selector, locator, options, &optional_filter_block)
    end
  end
end

def assert_none_of_selectors(*args, &optional_filter_block)
  assert_selectors(:assert_not_selector, args, &optional_filter_block)
end
```

But you're right. Keyword arguments will solve this in better way.

abotalov:
What is the benefit of using these methods instead of just:
```ruby
page.assert_selector(:css, '#my_div')
page.assert_selector(:css, 'a.not_clicked')
```
?

twalpole:
@abotalov `page.assert_all_of_selectors(:css, '#my_div', 'a.not_clicked') ` is slightly less typing and reads fine - see the rails PR linked above

PR #1822, opened 2017-01-03
Fix name of one selector spec

abotalov:
Behaviour was changed in https://github.com/teamcapybara/capybara/commit/b13e4d3887bfbfd9bcd765c7292554a95b52f549


twalpole:
Thanks

PR #1823, opened 2017-01-03
Change :style option for attach_file to :make_visible

twalpole:
This changes the :style option for `attach_file` to :make_visible which takes true or a hash of style options to change.  If `true` is specified it currently applies opacity:1, display: 'block'.   Still considering how to handle errors/the changes not making the field visible, currently it warns and then raises an error about not finding the element -- probably should just raise an error about the style changes not making the element visible.

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -230,7 +230,7 @@ module Capybara
       # @option options [String] id             Match fields that match the id attribute
       # @option options [String] name           Match fields that match the name attribute
       # @option options [String, Array<String>] :class    Match links that match the class(es) provided
-      # @option options [Hash] style   A Hash of CSS styles to change before attempting to attach the file (may not be supported by all driver)
+      # @option options [true, Hash] make_visible   A Hash of CSS styles to change before attempting to attach the file, if `true` { opacity: 1, display: 'block', visibility: 'visible' } is used (may not be supported by all drivers)
       #
       # @return [Capybara::Node::Element]  The file field element
       def attach_file(locator, path, options={})
@@ -239,11 +239,18 @@ module Capybara
           raise Capybara::FileNotFound, "cannot attach file, #{p} does not exist" unless File.exist?(p.to_s)
         end
         # Allow user to update the CSS style of the file input since they are so often hidden on a page
-        if style = options.delete(:style)
+        if style = options.delete(:make_visible)
+          style = { opacity: 1, display: 'block', visibility: 'visible' } if style == true
           ff = find(:file_field, locator, options.merge({visible: :all}))
           _update_style(ff, style)
+          if ff.visible?
+            ff.set(path)
+          else
+            raise ExpectationNotMet, "The style changes in :make_visible did not make the file input visible"
+          end
+        else
+          find(:file_field, locator, options).set(path)
         end
-        find(:file_field, locator, options).set(path)
       end
 
     private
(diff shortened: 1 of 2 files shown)


lib/capybara/node/actions.rb
```diff
@@ -230,7 +230,7 @@ def unselect(value, options={})
       # @option options [String] id             Match fields that match the id attribute
       # @option options [String] name           Match fields that match the name attribute
       # @option options [String, Array<String>] :class    Match links that match the class(es) provided
-      # @option options [Hash] style   A Hash of CSS styles to change before attempting to attach the file (may not be supported by all driver)
+      # @option options [true, Hash] make_visible   A Hash of CSS styles to change before attempting to attach the file, if `true` { opacity: 1, display: 'block' } is used (may not be supported by all driver)
```
abotalov: all driver -> all drivers
?

lib/capybara/node/actions.rb
```diff
@@ -239,9 +239,11 @@ def attach_file(locator, path, options={})
           raise Capybara::FileNotFound, "cannot attach file, #{p} does not exist" unless File.exist?(p.to_s)
         end
         # Allow user to update the CSS style of the file input since they are so often hidden on a page
-        if style = options.delete(:style)
+        if style = options.delete(:make_visible)
+          style = { opacity: 1, display: 'block' } if style === true
```
abotalov: Did you use case equality / subsumption operator purposely?

twalpole: No I did not - result of swapping between multiple languages  - fixing - thanks

twalpole:
@abotalov Can you think of any other style that would make sense for the default (true) option to trigger.  It could do things like make the position absolute and move the element to the top of the page, etc - but that might be going too far and better suited for people to specify as custom changes if needed.  I believe the majority of cases (of hidden file inputs) I see currently are just the field with opacity set to 0 and then overlayed over whatever element is being styled as the upload button.

abotalov:
By the way, a file input can have `opacity: 0`, `display: block` and be interactable in Selenium.

twalpole:
@abotalov Which version of selenium?

abotalov:
The following spec passed in Chrome with selenium-webdriver 3.0.5 on this branch:
```ruby
@session.attach_file("hidden_file", __FILE__, visible: false)
```

twalpole:
Interesting, I would have sworn that didn't work previously.  It definitely violates the webdriver definition of intractability since opacity: 0 is specifically mentioned

twalpole:
This may not be needed (in the current form) with the current versions of drivers since they appear to allow setting file inputs when hidden (using visible: false option) - need to figure out what exactly the CSS settings were for people whom that doesn't work for

twalpole:
@abotalov Ok -- that examples does fail when testing Selenium with Firefox without marionette (<= 2.47), but it looks like geckodriver and chromedriver support attaching a file to non visible file inputs

lib/capybara/node/actions.rb
```diff
@@ -230,7 +230,7 @@ def unselect(value, options={})
       # @option options [String] id             Match fields that match the id attribute
       # @option options [String] name           Match fields that match the name attribute
       # @option options [String, Array<String>] :class    Match links that match the class(es) provided
-      # @option options [Hash] style   A Hash of CSS styles to change before attempting to attach the file (may not be supported by all driver)
+      # @option options [true, Hash] make_visible   A Hash of CSS styles to change before attempting to attach the file, if `true` { opacity: 1, display: 'block' } is used (may not be supported by all drivers)
```
abotalov: `visibility: 'visible'` was added

PR #1826, opened 2017-01-09
Work around issue in Ruby 2.4.0 

twalpole:
Ruby 2.4.0 has an issue where using Forwardable delegators can produce random errors

https://bugs.ruby-lang.org/issues/13107?next_issue_id=13106&prev_issue_id=13111

This works around if for the methods we have seen errors on from rack-test.  The other option is to just wait for 2.4.1 to release where the issue has been fixed.

### DIFF
--- a/lib/capybara/rack_test/driver.rb
+++ b/lib/capybara/rack_test/driver.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'rack/test'
+require 'capybara/rack_test/patch.rb'
 require 'rack/utils'
 require 'mime/types'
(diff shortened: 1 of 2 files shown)


abotalov:
2.4.0 may cause a number of issues in various gems. So, perhaps, it doesn't make much sense to do patches for them at gem level.

twalpole:
@abotalov  Yeah - I think I'm just going to mark 2.4.0 as allowed failures on travis and ignore it for now - 2.4.0 is basically unusable if anything included into a project uses delegators from forwardable.

PR #1827, opened 2017-01-13
Do not raise path exceptions on nil current_url

dylanlive:
When we experimented moving our suite to BrowserStack, `page.current_url` was often returning nil.

Calling `have_current_path` and `page.current_path` would raise:
`NoMethodError: undefined method 'path' for nil:NilClass`, with no intelligent waiting for BrowserStack to finally return a valid url.

This PR modified to only call those methods if the parsed URI is not nil. If it's nil, @actual_path remains nil.

### DIFF
--- a/lib/capybara/queries/current_path_query.rb
+++ b/lib/capybara/queries/current_path_query.rb
@@ -17,10 +17,12 @@ module Capybara
         @actual_path = if options[:url]
           session.current_url
         else
+          uri = ::Addressable::URI.parse(session.current_url)
+
           if options[:only_path]
-            ::Addressable::URI.parse(session.current_url).path
+            uri.path unless uri.nil? # Ensure the parsed url isn't nil.
           else
-            ::Addressable::URI.parse(session.current_url).request_uri
+            uri.request_uri unless uri.nil? # Ensure the parsed url isn't nil.
           end
         end
 
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -173,6 +173,10 @@ module Capybara
     def current_path
       # Addressable parsing is more lenient than URI
       uri = Addressable::URI.parse(current_url)
+
+      # If current_url ends up being nil, won't be able to call .path on a NilClass.
+      return nil if uri.nil?
+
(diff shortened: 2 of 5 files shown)


dylanlive:
Hmm, I was seeing this locally too. `Capybara::Session DSL` tests don't seem to let me stub `current_url` to return nil. Any ideas?

twalpole:
Because it's to from the driver , which runs in a different thread.  It's the drivers responsibility not to return nil

dylanlive:
Per BrowserStack, apparently it's intentional.
```

dylanlive:
Per BrowserStack, apparently it's intentional for Selenium to return nil.

They disclosed this to me in an email:
> We were able to confirm that this is the default behaviour of Selenium when a webpage is currently in loading state. Basically, Selenium responds /provides a particular value to a getCurrentUrl() method according to the current status of the webpage. If the webpage is still loading it returns a null value, whereas, if the webpage has loaded completely it returns a valid response.

I have been running this patch locally over the last few weeks, and verified Capybara's wait feature properly works (when BrowserStack is finally providing a legitimate url).

abotalov:
Could you also, perhaps, file an issue in https://github.com/w3c/webdriver?
Currently the specification doesn't seem to say anything about what behaviour should be in this case.

twalpole:
Interesting, I've never seen Selenium return nil fro current_url when used with anything else, but this code doesn't complicate anything so I'm fine to add this.  

The tests are failing on the DSL portion (They are run against  DSL, rack_test and selenium) because during DSL testing @session isn't actually a Capybara::Session instance.  You should be able to mock using 

    allow_any_instance_of(Capybara::Session).to receive(:current_url) { nil }

and have it work.   Once you've fixed that please squash the commits and I'll go ahead and merge.

Thanks!

dylanlive:
@twalpole Yeah I hadn't either until we experimented moving our suite to BrowserStack. I've noticed a lot of latency in general. Regardless, I figured it wouldn't hurt to prevent this issue from raising an exception in Capybara.

Thanks for the suggestion! I've updated the mock, squashed commits, and updated.

@abotalov I'll take a look into this.

PR #1830, opened 2017-01-27
Support returning an element from evaluate_script with selenium

twalpole:


abotalov:
Does it support unwrapping? (https://github.com/SeleniumHQ/selenium/blob/selenium-3.0.1/rb/spec/integration/selenium/webdriver/driver_spec.rb#L182)
I think support for it is also needed as behaviour is surprising otherwise.

twalpole:
@abotalov No - it supports returning AN element -- It's not going to run through objects to figure out whether or not they have elements in them

twalpole:
@abotalov Actually - I've implemented full unwrapping for selenium - Just need to check it can work with Poltergeist and Capybara too

PR #1835, opened 2017-02-16
Workaround JRuby lazy enumerators bug below version 9.2.0.0

boris-petrov:
https://github.com/jruby/jruby/issues/4212 - this bug is still unfixed and they've suggested a fix will be in 9.2.0.0. Because of that, https://github.com/teamcapybara/capybara/commit/f7e537006af5cf882bc3df4317d2974bf392c974 breaks Capybara 2.12.0 for JRuby above 9.1.5.0. This is our case - we use JRuby 9.1.7.0 and Capybara 2.12.0 doesn't work.

### DIFF
--- a/lib/capybara/result.rb
+++ b/lib/capybara/result.rb
@@ -28,11 +28,11 @@ module Capybara
       @result_cache = []
       @results_enum = lazy_select_elements { |node| query.matches_filters?(node) }
       @query = query
-      # JRuby < 9.1.6.0 has an issue with eagerly finding next in lazy enumerators which
+      # JRuby < 9.2.0.0 has an issue with eagerly finding next in lazy enumerators which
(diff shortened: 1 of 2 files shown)


twalpole:
Hmmm -- I could have sworn this was fixed in 9.1.6.0, did they revert?  Also it's passing the tests we have for it.

twalpole:
Actually it passes the tests locally, but does appear to be causing hangs on travis -- I'm just going to remove lazy results evaluation from JRuby totally (without version check) and we'll worry about re-enabling it later.  I'll release a 2.12.1 later today (California) once all the tests pass.

twalpole:
Fixed via https://github.com/teamcapybara/capybara/commit/1b65bf243c506ede3a14d57faa985616a92e505d and 2.12.1 released

PR #1841, opened 2017-03-06
Add parity between rspec support and minitest

twalpole:


twalpole:
See https://github.com/rails/rails/issues/28254

PR #1850, opened 2017-03-28
Proxy all/within to the correct version based on parameters passed

twalpole:
Call either the Capybara or RSpec versions of all/within based on the parameter types passed in.  Potentially obviates the need for PR #1814, and just does what the user expects to happen.   Risks are if RSpec changes the acceptable parameters for their all/within but they don't seem like matchers that could really accept anything else.


twalpole:
@abotalov What do you think about something like this for the all/within issue?

abotalov:
The solution seems good to me.

But this doesn't fully help in case user doesn't require `capybara/rspec` or `capybara/cucumber` (because the user should know that they should include Capybara::RSpecMatcherProxies manually).

I think it would be better to be more intrusive and to include this module automatically when a user invokes either `include Capybara::DSL` or `include RSpec::Matchers` if the other module is already included.

twalpole:
@abotalov My thought was to make it possible to not include the proxies in case they cause issues,  however you may be right.

twalpole:
@abotalov  I've changed it to include the proxies whenever Capybara::RSpecMatchers is included as long as Capybara::DSL has already been included, and then monkey patched RSpec to raise an error when the parameters aren't valid to output a message about including RSpec::Matchers before Capybara::DSL - I think that should take care of the current situations

abotalov:
The clash is between `Capybara::DSL` and `RSpec::Matchers`. So I think the new module should be included not when `Capybara::RSpecMatchers` is included but when either `Capybara::DSL` or `RSpec::Matchers` is included.

E.g.:
```ruby
module Capybara
  module DSL
    def self.included(mod)
      if defined?(::RSpec::Matchers) && mod.include?(::RSpec::Matchers)
        require 'matcher_proxies'
        mod.send(:include, ::Capybara::RSpecMatcherProxies)
      end
    end
  end
end

if defined?(::RSpec::Matchers)
  module RSpec
    module Matchers
      def self.included(mod)
        if mod.include?(::Capybara::DSL)
          require 'matcher_proxies'
          mod.send(:include, ::Capybara::RSpecMatcherProxies)
        end
      end
    end
  end
end
```

twalpole:
I was trying to stay away from RSpec::Matchers included in case they decide to implement it, but I will consider it.

lib/capybara/rspec/matchers.rb
```diff
@@ -268,3 +268,27 @@ def become_closed(options = {})
     end
   end
 end
+
+if defined?(::RSpec::Matchers)
```
abotalov: Currently this code is executed only if user requires `capybara/rspec/matchers`.
But the user may just include `Capybara::DSL` and `RSpec::Matchers` into the same scope without requiring this capybara file.

twalpole: Yep - it's not done yet - working through piece by piece

abotalov:
> I was trying to stay away from RSpec::Matchers::included in case they decide to implement it, but I will consider it.

I think it would be better to keep proxy-related logic in a single space. IMO either Capybara should know which clashes it has with RSpec or RSpec should know about clashes.
Otherwise (IMO) proxy would have to be supported in two repositories.

twalpole:
@abotalov - I was referring to the `included` method -- not that RSpec would implement the proxies

README.md
```diff
@@ -241,6 +241,10 @@ RSpec.describe "todos/show.html.erb", type: :view do
 end
 ```
 
+**Note: When you require 'capybara/rspec' proxy methods are installed to work around name collisions between Capybara::DSL methods
+  `all`/`within` and the identically named built-in RSpec matchers. If you opt not to require 'capybara/rspec' (because you wanted to customize the RSpec
+  configurationm etc.), you can install the proxy methods by requiring 'capybara/rspec/matcher_proxies' after requiring RSpec and 'capybara/dsl'**
```
abotalov: `configuration`

PR #1853, opened 2017-04-01
Add browser local/session storage option

twalpole:
Add a setting to clear localStorage and sessionStorage on reset if Capybara is running the server for the session.


abotalov:
Do you want to keep Selenium-specific option (https://github.com/teamcapybara/capybara/pull/1781) as well?

twalpole:
I was thinking about deprecating them, but hadn't fully decided - The issue with those options is that they can't clear the storage in all conditions (if storage is saved in an onunload handler) so they're not perfect.  I'm also trying to decide whether the new version should support specifying to clear the session or local storage - or whether it should just be both as it is now.  Thoughts?

PR #1863, opened 2017-05-18
Add missing bracket to a selector's description when a class is specified

ElMassimo:
There is a missing square bracket at the beginning of the description message for any selector if a class is specified, this pull request adds it.

### Example
```ruby
find('.item', text: 'content', class: 'css')
```
results in:
```
expected to find ".item" with text "content" with classes css] but there were no matches
```
and after the fix:
```
expected to find ".item" with text "content" with classes [css] but there were no matches
```


twalpole:
Thanks for this, could you change it to add the missing [ at the other end rather than removing the ending one please?

twalpole:
@ElMassimo Thanks! -- If you could squash those two commits down to one, I'll go ahead and merge this.

twalpole:
@ElMassimo Thanks for this - as soon as the tests pass I'll merge this in to master

PR #1868, opened 2017-05-26
Fix Ruby warnings that appeared in 2.14.0

matthewd:
Previous warning-avoidance on similar config settings has used `attr_reader` / `attr_writer`... but here as the attributes need to be listed in the constant, it seems simplest to just remove the unwanted accessor before replacing it.


twalpole:
Thanks

Issue #1865, opened 2017-05-23
has_select?() should support with_selected argument for symmetry with with_options


dwt:
## Meta
Capybara Version: 2.10.1

## Expected Behavior
It should be easy to query wether a multi-select <select>-box has a specific selection as part of it's multi select.

## Actual Behavior
Actually that's surprisingly difficult.

## Steps to reproduce
Try to find out wether a particular <select> has a specific selection as part of it's selection. Then nearly despair once you find out that `with_options:` is actually supported, but `with_selected:` or something similar is not.

twalpole:
Valid point, do you want to submit.a PR?  Also note that in your own code you can always add a new filter to any of the built-in sectors using `Capybara.modify_selector`

dwt:
@twalpole I'd rather not submit a pr, as my ruby is not quite at production strength.

Oh, and thanks for the heads up with the modify_selector, I didn't know about that!

twalpole:
@dwt Should be a relatively easy to thing to implement based off the current implementation of :select ed and :with_options.  If you have a chance it would be a good PR for a beginner rubyist to start with, If not I'll probably get around to it in a week or two.

zatapathique:
Hey @twalpole, we are having the same issue in one of our projects so I'm willing to implement this feature :)

twalpole:
@zatapathique Great

twalpole:
@zatapathique Any update on this?

zatapathique:
@twalpole I'm planning to implement this at the weekend.

twalpole:
Closed via PR #1873

PR #1873, opened 2017-06-04
Add :with_selected filter to :select selector

zatapathique:
Add :with_selected filter connected with [this](https://github.com/teamcapybara/capybara/issues/1865) issue.

### DIFF
--- a/lib/capybara/selector.rb
+++ b/lib/capybara/selector.rb
@@ -352,6 +352,7 @@ end
 # @filter [Array<String>] :options Exact match options
 # @filter [Array<String>] :with_options Partial match options
 # @filter [String, Array<String>] :selected Match the selection(s)
+# @filter [String, Array<String>] :with_selected Partial match the selection(s)
 #
 Capybara.add_selector(:select) do
   label "select box"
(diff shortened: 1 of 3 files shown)


lib/capybara/selector.rb
```diff
@@ -380,15 +381,21 @@
...
 
   filter(:selected) do |node, selected|
-    actual = node.all(:xpath, './/option', visible: false).select { |option| option.selected? }.map { |option| option.text(:all) }
-    [selected].flatten.sort == actual.sort
+    actual = node.all(:xpath, './/option', visible: false).select(&:selected?).map { |option| option.text(:all) }
+    Array(selected).sort == actual.sort
+  end
+
+  filter(:with_selected) do |node, selected|
+    actual = node.all(:xpath, './/option', visible: false).select(&:selected?).map { |option| option.text(:all) }
```
zatapathique: Im not too happy about this line repeating in :select and :with_selected filters, though I didn't want to extract this to method/lambda to not break the very clear structure of this file.

twalpole: I don't have an issue with lines repeating in this file.

lib/capybara/selector.rb
```diff
@@ -380,15 +381,21 @@
   end
 
   filter(:selected) do |node, selected|
-    actual = node.all(:xpath, './/option', visible: false).select { |option| option.selected? }.map { |option| option.text(:all) }
-    [selected].flatten.sort == actual.sort
+    actual = node.all(:xpath, './/option', visible: false).select(&:selected?).map { |option| option.text(:all) }
+    Array(selected).sort == actual.sort
```
twalpole: Array(selected) and [selected].flatten are not identical since the latter will flatten multiple nested arrays - In this instance it probably doesn't make any difference though

twalpole:
@zatapathique Please squash this down to one commit

lib/capybara/selector.rb
zatapathique: That's true, but as you said, it doesn't make a difference in this case since, based on do documentation and specs I read, nesting options is not something that's deliberately supported.

zatapathique:
@twalpole squashed :)

twalpole:
Thanks for this!

PR #1876, opened 2017-06-09
Ensure the cursor points to the end of form before sending backspaces

shinnn:
### Problem

In the latest stable Google Chrome (v59.0.3071.86), when an `input` element has `autofocus` attribute, the mouse cursor doesn't points its end but its *head*.

image

This behavior causes a problem in the backspace-based value replacement introduced in https://github.com/teamcapybara/capybara/commit/ed2071b480e3849b64754be48da82803ee210b31 – no matter how many times Capybara presses the backspace key, it cannot clear the form value since the backspace key can only deletes values before the current cursor position.

Here is the example test code to reproduce the problem:

`run.rb`

```ruby
require "minitest/autorun"
require 'capybara/minitest'

Capybara.app = Rack::File.new File.dirname __FILE__

Capybara.register_driver :chrome do |app|
  Capybara::Selenium::Driver.new(app, browser: :chrome)
end

Capybara.current_driver = :chrome
Capybara.javascript_driver = :chrome

class Test < Minitest::Test
  include Capybara::DSL

  def test_input
    visit '/test.html'
    find('input').set('bar', clear: :backspace)
    assert_equal 'bar', find('input').value
  end
end
```

`test.html`

```html
<!DOCTYPE html>
<html>
<body>
  <input autofocus value="foo">
</body>
</html>
```

```
$ ruby run.rb
Run options: --seed 27403

# Running:

F

Finished in 1.661836s, 0.6017 runs/s, 0.6017 assertions/s.

  1) Failure:
Test#test_input [run.rb:20]:
Expected: "bar"
  Actual: "barfoo"

1 runs, 1 assertions, 1 failures, 0 errors, 0 skips
```

image

### Solution

I added a keydown `↓` before backspaces and ensure the cursor always points the end of the form before clearing values.

### DIFF
--- a/spec/shared_selenium_session.rb
+++ b/spec/shared_selenium_session.rb
@@ -71,6 +71,13 @@ RSpec.shared_examples "Capybara::Session" do |session, mode|
         expect(@session.find(:fillable_field, 'form_first_name').value).to eq('Harry')
       end
 
+      it 'should clear an existing value even if a cursor is autofocused to a field' do
+        @session.visit('/form')
+        @session.fill_in('form_last_name', with: 'Potter',
+                          fill_options: { clear: :backspace} )
+        expect(@session.find(:fillable_field, 'form_last_name').value).to eq('Potter')
+      end
+
       it 'should only trigger onchange once' do
         @session.visit('/with_js')
         @session.fill_in('with_change_event', with: 'some value',
--- a/lib/capybara/selenium/node.rb
+++ b/lib/capybara/selenium/node.rb
@@ -61,7 +61,7 @@ class Capybara::Selenium::Node < Capybara::Driver::Node
         if options[:clear] == :backspace
           # Clear field by sending the correct number of backspace keys.
           backspaces = [:backspace] * self.value.to_s.length
-          native.send_keys(*(backspaces + [value.to_s]))
+          native.send_keys(*([:down] + backspaces + [value.to_s]))
         elsif options[:clear] == :none
           native.send_keys(value.to_s)
         elsif options[:clear].is_a? Array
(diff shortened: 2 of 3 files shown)


spec/shared_selenium_session.rb
```diff
@@ -71,6 +71,13 @@
         expect(@session.find(:fillable_field, 'form_first_name').value).to eq('Harry')
       end
 
+      it 'should clear an existing value even if a cursor is autofocused to a field' do
+        @session.visit('/form')
+        @session.fill_in('form_last_name', with: 'Potter',
+                          fill_options: { clear: :backspace} )
+        expect(@session.find(:fillable_field, 'form_last_name').value).to eq('Potter')
```
shinnn: Without `[:down]`, this test actually fails:

```
1) Capybara::Session with chrome with selenium driver #fill_in with { :clear => :backspace } fill_option should clear an existing value even if a cursor is autofocused to a field
   Failure/Error: expect(@session.find(:fillable_field, 'form_last_name').value).to eq('Potter')
   
     expected: "Potter"
          got: "PotterSmith"
   
     (compared using ==)
   Shared Example Group: "Capybara::Session" called from ./spec/selenium_spec_chrome.rb:40
   # ./spec/shared_selenium_session.rb:78:in `block (4 levels) in <top (required)>'
```

shinnn:
https://travis-ci.org/teamcapybara/capybara/jobs/241092641#L516-L517

The test with Firefox 47 is failing, but I cannot reproduce the failure on my macOS v10.12.4, Firefox v45.4.0.

Maybe `↓` doesn't move the mouse cursor to the end of the form in Linux?

twalpole:
The :backspace clear method is defined as sending the number of backspaces there are characters in the control.  I don't think adding a down arrow to that is a great solution for a change in one browser version.  I think you'd be better off sending something like   `clear: [[:control, 'a'], :backspace]]` in the specific tests where you have issues.

twalpole:
Also, the fact that this doesn't work on most of the travis tests means it definitely can't be included as is.

PR #1882, opened 2017-06-23
Loosen restrictions on where `within_window` can be called from

twalpole:
Theres's no longer any need to prevent a user from doing

     within_window(window1) do
       ... do something ...
       within_window(window2) do
         ... do something in window2
       end
       .... do something else in window1
    end

This PR loosens the restrictions to allow for this (still can't be used inside `within_frame` blocks).  The restriction stays for `switch_to_window` since it can really screw things up if used inside within's

### DIFF
--- a/lib/capybara/spec/session/window/within_window_spec.rb
+++ b/lib/capybara/spec/session/window/within_window_spec.rb
@@ -57,10 +57,10 @@ Capybara::SpecHelper.spec '#within_window', requires: [:windows] do
       expect(@session.send(:scopes)).to eq([nil])
     end
 
-    it "should leave correct scopes after execution in case of error" do
+    it "should leave correct scopes after execution in case of error", requires: [:windows, :frames] do
       window = (@session.windows - [@window]).first
       expect do
-        @session.within 'html' do
+        @session.within_frame 'frameOne' do
           @session.within_window(window) {}
         end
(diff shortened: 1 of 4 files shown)


lib/capybara/spec/session/window/within_window_spec.rb
```diff
@@ -102,6 +102,45 @@
...
+        end
+        expect(@session).to have_css('#divInPopupTwo')
+        expect(@session).not_to have_css('divInPopupOne')
+      end
+      expect(@session).not_to have_css('#divInPopupTwo')
+      expect(@session).not_to have_css('divInPopupOne')
+      expect(@session.title).to eq('With Windows')
+    end
+
+    it "should be able to nest within_window" do
```
abotalov: What is the difference between these two specs?

twalpole: Good question -- looks like a copy paste screwup -- fixing

abotalov:
What use case do you see for this?

twalpole:
It came from a question on SO - https://stackoverflow.com/questions/44703919/with-capybara-how-do-i-switch-to-the-new-window-when-i-have-already-switched-to - which is probably not the best way to structure a test but there really is no technical reason why it shouldn't be allowed.  It would probably make more sense in automation than testing but why limit it.

PR #1884, opened 2017-06-28
Switch to mini_mime gem from mime-types

Fryguy:
[mini_mime](https://github.com/discourse/mini_mime) is a much more performant replacement for the mime-types gem.  It uses the same database as mime-types, but loads it in a much more efficient manner, improving the number of allocations, memory, and speed.  The [mail](https://github.com/mikel/mail) gem, which is a core component of Rails, has already switched to it.  See https://github.com/mikel/mail/pull/1059 for more details on the mail gem changes, as well as the some performance numbers.  Bonus is that mini_mime supports Ruby versions all the way back to 1.8.7, so there's no need for a caveat anymore.

cc @SamSaffron

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Capybara requires Ruby 1.9.3 or later. To install, add this line to your
 gem 'capybara'
 ```
 
-**Note:** If using Ruby < 2.0 you will also need to limit the version of mime-types to < 3.0 and the version of rack to < 2.0
+**Note:** If using Ruby < 2.0 you will also need to limit the version of rack to < 2.0
 
 If the application that you are testing is a Rails app, add this line to your test helper file:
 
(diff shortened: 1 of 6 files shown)


Fryguy:
Performance details (quoting @SamSaffron from the referenced PR)

> The mime types gem (even with the magical columnar store by @jeremyevans) is a memory hog.
> 
> In particular on boot, mime/types/columnar will allocate **109k** objects and end up with a retained count of **31K** objects.
> 
> This bloat leaves objects that need to be marked and swept every major GC and bloats ruby processes.
> 
> To circumvent mini_mime was created.
> 
> It uses the exact same database as the full fledged mime types and is capable of only loading stuff on demand with a practical, safe and bound in-memory cache.
> 
> mini_mime will allocate **398** objects on boot and only retain **62** (due to rubygems inefficiency)
> 
> In table form:
> 
> | boot allocations | boot retained | lookup | lookup uncached |
> | --- | --- | --- | --- |
> | mini_mime | 398 | 62 | 641K/s | 33K/s |
> | mime-types | 109796 | 31165 | 361K/s | 361K/s |
> 
> The performance of mini_mime is pretty good, cached lookups are faster than the mime types gem and uncached lookups are 10x slower. (which is really not a huge issue considering the memory savings)

twalpole:
LGTM - thanks!

PR #1886, opened 2017-07-05
implement `refresh`

twalpole:


lib/capybara/selenium/driver.rb
```diff
@@ -67,6 +67,13 @@ def visit(path)
     browser.navigate.to(path)
   end
 
+  def refresh
+    accept_modal(wait: 0.1) do
```
elliterate: It looks like you intend for the `wait` to make its way into the `options` hash, for use in `find_modal`. But instead it's swallowed by the unused `_type` parameter. You probably want something like this:

```ruby
accept_modal(nil, wait: 0.1) do
  # ...
end
```

twalpole: Thanks for catching that -  fixed

PR #1896, opened 2017-07-31
Silence puma output

jonleighton:
Without this, I see the following output when I run my tests:

Puma starting in single mode...
* Version 3.8.2 (ruby 2.4.0-p0), codename: Sassy Salamander
* Environment: test
* Listening on tcp://127.0.0.1:37739
Use Ctrl-C to stop

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -448,7 +448,7 @@ end
 
 Capybara.register_server :puma do |app, port, host|
   require 'rack/handler/puma'
-  Rack::Handler::Puma.run(app, Host: host, Port: port, Threads: "0:4", workers: 0, daemon: false)
+  Rack::Handler::Puma.run(app, Host: host, Port: port, Threads: "0:4", workers: 0, daemon: false, Silent: true)
 end
 
 Capybara.configure do |config|


twalpole:
Jon - the problem I have with this is that the output provides useful info when trying to diagnose Capybara issues with new users.   Rails added a configuration option to their puma server registration to take care of this (although they don't yet force puma into single mode (in process) which I believe is going to create some debugging headaches for users).  Maybe adding a second puma registration, `:puma_silent`, or something would be worthwhile.

twalpole:
Or something along the lines of https://github.com/teamcapybara/capybara/commit/fa47c84413d15f624c902f6d10efdaafd66fa3d8 so the user can specify 

    Capybara.server = :puma, { Silence: true }

if the passing of the Silence option is desired

jonleighton:
That seems like a decent solution to me 👍 Alternatively, perhaps we could send the output to the Rails log file so it can still be found but isn't so in your face.

twalpole:
@jonleighton After a very quick look at the Puma options available, I don't immediately see how one would go about sending the output to the Rails log file (definitely doesn't mean it's not possible).  If you know how and would like to submit a PR that overrides the default :puma server registration in `lib/capybara/rails.rb` to send the output to the Rails log file I'd be good with that.

twalpole:
Closing in favor of PR #1897

PR #1897, opened 2017-08-02
Allow registered servers to receive options

twalpole:
Option to PR# 1896 that leaves the default outputting useful info.  Experience users can then opt to silence Puma when wanted without having to register their own server block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,13 @@ Capybara.app = MyRackApp
 ```
 
 If you need to test JavaScript, or if your app interacts with (or is located at)
-a remote URL, you'll need to [use a different driver](#drivers).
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -106,7 +106,13 @@ Capybara.app = MyRackApp
 ```
 
 If you need to test JavaScript, or if your app interacts with (or is located at)
-a remote URL, you'll need to [use a different driver](#drivers).
+a remote URL, you'll need to [use a different driver](#drivers).  If using Rails 5.0+, but not using the Rails system tests from 5.1, you'll probably also
+want to swap the "server" used to launch your app to Puma in order to match Rails defaults.
+
+```ruby
+Capybara.server = :puma # Until you're setup is working
```
jonleighton: Small grammar point: "you're" = "you are", it should be "your"

twalpole: Thanks for catching that - fixed

PR #1900, opened 2017-08-08
Don't wait for element during inspection

ineverov:
In the current implementation, if you inspecting stale element, Capybara will wait for `default_max_wait_time` before inspecting it.
In proposed patch, this behavior is changed to immediately return obsolete element

If you have the stale collection of element this wait is multiplied by the number of elements in the collection.

### DIFF
--- a/lib/capybara/node/element.rb
+++ b/lib/capybara/node/element.rb
@@ -371,9 +371,9 @@ module Capybara
       end
 
       def inspect
-        %(#<Capybara::Node::Element tag="#{tag_name}" path="#{path}">)
+        %(#<Capybara::Node::Element tag="#{base.tag_name}" path="#{base.path}">)
       rescue NotSupportedByDriverError
-        %(#<Capybara::Node::Element tag="#{tag_name}">)
+        %(#<Capybara::Node::Element tag="#{base.tag_name}">)
       rescue => e
         if session.driver.invalid_element_errors.any? { |et| e.is_a?(et)}
           %(Obsolete #<Capybara::Node::Element>)
--- a/spec/shared_selenium_session.rb
+++ b/spec/shared_selenium_session.rb
@@ -161,7 +161,8 @@ RSpec.shared_examples "Capybara::Session" do |session, mode|
       it "outputs obsolete elements" do
         @session.visit('/form')
         el = @session.find(:button, 'Click me!').click
-        sleep 2
+        expect(@session).to have_no_button('Click me!')
+        expect(el).not_to receive(:synchronize)
         expect(el.inspect).to eq "Obsolete #<Capybara::Node::Element>"
       end
     end


twalpole:
Hmmm, thinking about any negative repercussions of this. Stale elements can become valid again, since they are potentially reloaded, and if changing the test I would much prefer 

     expect(page).to have_no_button('Click me!')

instead of the potentially never-ending while loop

What is your use case where you are calling inspect on a collection of stale elements?

ineverov:
@twalpole fixed test.
If element becomes valid again after reload inspect will return proper string next call. But why we need it for inspection element?

About use case, I was investigating another issue with capybara and also found this problem.
Original issue can be reproduced by this scenario [diff](https://github.com/teamcapybara/capybara/compare/master...ineverov:all_timeout?expand=1)

Imagine you have a set of links on the page. You do `all(:link)` which will return a lazy iterator. Meanwhile, links have changed/reloaded on the page. And after some time you just want to get size/iterate through links. Capybara will evaluate `matches_filter?` for each of stale elements and wait for `wait_time` on checking visibility of stale element [here](https://github.com/teamcapybara/capybara/blob/59750ebfef667cf50804219ebe72690c1db88d8a/lib/capybara/queries/selector_query.rb#L83) because it uses `synchronize` and elements are not `reloadable` if you using `all`. As a result, we wait `wait_time * number_of_links` to return an empty array, but reloaded links exist on the page.
I was trying to play with `allow_reload` a little bit, but taping elements with `.allow_reload!`[here](https://github.com/teamcapybara/capybara/blob/59750ebfef667cf50804219ebe72690c1db88d8a/lib/capybara/node/finders.rb#L269) evaluates lazy iterator. So we need to set `@allow_reload` during element initialization.

So I have a question, why don't we set allow_reload in `all` and `first`, but setting it in all other finders?

twalpole:
`all` and `first` don't have `allow_reload` set because it can lead to strange/confusing/undefined results.  Elements are reloaded when they become obsolete, therefore if we had an array of 4 matched elements [1, 2, 3, 4] and removed the 2nd element from the page, that 2nd element would reload and become what was the 3rd element. However since the 3rd element never became obsolete it wouldn't reload so you'd now have an array of 1, 3, 3, 4 elements which doesn't make a lot of sense.  `first` is because it's currently based on `all` but will be changing in Capybara 3.0 whenever that is finished/released.  `all` and `first` really should be used much less than people use them (lots of terrible "tutorials" and examples out there) and the single element finders preferred whenever feasible.

As for `inspect` retrying there probably isn't necessary, however I'm going to think about it for a bit more just to make sure.

ineverov:
@twalpole thanks for explaining. I appreciate it!

What about an issue I described when you wait `number_of_element * timeout` seconds?

I'm using `all` through `site_prism` gem, which provides nice page objects around capybara.

But here is an example: the page has a search field and list or results. It updates results filtering data by value in the input field. I want to check search is working correctly. And I may have more than 1 result for each input (in general, unknown count > 0). 
What I'm doing now (using capybara DSL)
```ruby
some_waiter_method_catching_stale_exceptions do
   rows = all('tr') # lazy
   expect(rows.size).to be > 0  # evaluate matches_filter? and waiting for number_of_stale_elements * wait_time
   rows.all? do |el|
      el.find('td:nth-child(1)').text == 'search'
   end
end
```
If a page is reloading it will be hanging on `all?` trying to evaluate `visible?` for stale elements. But results may already be updated at the time it finishes and we may reach time out in waiter method.

I, personally, would prefer it failed with StaleElementError or at least not to wait in such cases.

About avoiding `all` usage, it would be good in the perfect world. But how to solve above example without `all`?

What about raising an error if the element is not reloadable [somewhere here](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/node/base.rb#L93) ?

twalpole:
@ineverov If you're checking whether your apps search is working correctly, then you also control the data that should be returned, so you should know what to expect and what to wait to appear/disappear on the page before calling all so you don't end up with a whole array of stale elements.  You should also be using the count options if you care about the size of the results rather than expecting on `size`. Assuming I understand sort of what your method is trying to do, it should probably just be written something like 

    expect(page).to have_css('tr td:nth-child(1)`, text: 'search', minimum: 1)

The real issue is that Capybara doesn't distinguish between errors that can change state if retried and errors that can't, but maybe it should.  That would require a change to the driver API to allow them to identify those types of errors though, so it's not just a minor fix.

ineverov:
@twalpole controlling data is complicated if you creating integration tests with micro service architecture. But it's another discussion.
About distinguishing, it would be nice to have in future. Probably in one of following major releases... 
Thanks for your help and clarifying.

PR #1902, opened 2017-08-13
Fix headless Chrome check not working properly

stanhu:
The call to `Driver#headless_chrome?` would fail since the data in the
memoized `@processed_options` would be cleared by the Selenium WebDriver.
To fix this problem, pass a copy of the Hash to the driver.

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -20,7 +20,10 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
       end
 
       @processed_options = options.reject { |key,_val| SPECIAL_OPTIONS.include?(key) }
-      @browser = Selenium::WebDriver.for(options[:browser], @processed_options)
+      # Selenium's Chrome driver deletes keys from the Hash
+      # (https://github.com/SeleniumHQ/selenium/blob/selenium-3.5.0/rb/lib/selenium/webdriver/chrome/driver.rb#L74),
+      # so make a copy to avoid losing data.
+      @browser = Selenium::WebDriver.for(options[:browser], @processed_options.clone)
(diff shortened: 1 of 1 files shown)


twalpole:
We rely on Selenium WebDriver modifying the contents of `@processed_options` (that's why it's called `@processed_options`) to normalize the parameters passed in.  What are you passing that is not working?

twalpole:
Closing due to the filing of a different PR for apparently the same issue, but with no reason.

stanhu:
This is different from #1903. The problem is this:

1. `Driver#browser` is called.
2. `@processed_options` is memoized and passed to `Selenium::Driver`.
3. The Chrome Selenium driver removes the `desired_capabilities` option.
4. `Driver#accept_modal` is called and calls `Driver#headless_chrome?`.
5. https://github.com/stanhu/capybara/blob/a3408d47fb33d416bf6fd6a292b2975d5c898adf/lib/capybara/selenium/driver.rb#L325 is called, but since the Hash is empty `caps` is returned `nil`.
6. The test fails completely with `nil` because the next line assumes a valid Hash is returned.

stanhu:
Does it make more sense to change https://github.com/stanhu/capybara/blob/a3408d47fb33d416bf6fd6a292b2975d5c898adf/lib/capybara/selenium/driver.rb#L325 to use `options` instead of `@processed_options`?

twalpole:
No because It needs to be the options modified by selenium to normalize them.  The only reason to stay on selenium 2.53.4 is if you need to test with FF <= 47, otherwise it’s obsolete - update to 3.5

stanhu:
Thanks, looks like updating to 3.5 worked. Sorry for the noise.

PR #1903, opened 2017-08-13
Fix detection of headless Chrome

stanhu:
Passing in a symbol for `chromeOptions` does not appear to do the right
thing. `chromeOptions` in String form appears to be the correct format for
`Selenium::WebDriver::Remote::Capabilities`.

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -320,7 +320,7 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
   def headless_chrome?
     if chrome?
       caps = @processed_options[:desired_capabilities]
-      chrome_options = caps[:chrome_options] || caps[:chromeOptions] || {}
+      chrome_options = caps[:chrome_options] || caps[:chromeOptions] || caps['chromeOptions'] || {}
       args = chrome_options['args'] || chrome_options[:args] || []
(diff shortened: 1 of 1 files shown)


twalpole:
This won't be accepted without a reason for it -  What EXACTLY are you passing to `Capybara::Selenium::Driver.new` and what version of selenium-webdriver are you using?

stanhu:
This was on selenium-webdriver 2.53.4. It looks like lots of updates have been made.

These are the options being used:

```
Capybara.javascript_driver = :chrome
Capybara.register_driver :chrome do |app|
  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
    'chromeOptions' => {
      'args' => %w[headless no-sandbox disable-gpu --window-size=1240,1400]
    }
  )

  Capybara::Selenium::Driver
    .new(app, browser: :chrome, desired_capabilities: capabilities)
end
```

This is in line with a number of references:

* http://blog.faraday.io/headless-chromium-with-capybara-and-selenium-webdriver/
* https://stackoverflow.com/q/44591612/1992201

I tried changing the config to `:chrome_options` to no avail. I'll see if the latest gem helps.

twalpole:
2.53.4 is obsolete - update to a newer selenium-webdriver

stanhu:
Looks like the update work. Thanks.

PR #1911, opened 2017-09-07
[fix] selenium set_text passing in empty string with options

gsilis:
If an empty value if passed in with a clear option, it seems that the
caller's intent is for the option to be honored regardless of the value
to be set in the field.

Just passing in a blank value without options would behave as it did before.

Related to this PR/use case: https://github.com/teamcapybara/capybara/pull/1498

### DIFF
--- a/spec/shared_selenium_session.rb
+++ b/spec/shared_selenium_session.rb
@@ -71,6 +71,16 @@ RSpec.shared_examples "Capybara::Session" do |session, mode|
       end
     end
 
+    context '#fill_in_with empty string and no options' do
+      it 'should trigger change when clearing a field' do
+        @session.visit('/with_js')
+        @session.fill_in('with_change_event', with: '')
+        # click outside the field to trigger the change event
+        @session.find(:css, 'body').click
+        expect(@session).to have_selector(:css, '.change_event_triggered', match: :one)
+      end
+    end
+
     context "#fill_in with { :clear => :backspace } fill_option", requires: [:js] do
       it 'should fill in a field, replacing an existing value' do
         @session.visit('/form')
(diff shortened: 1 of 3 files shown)


spec/shared_selenium_session.rb
```diff
@@ -71,7 +71,17 @@
...
+      it 'should trigger change when clearing a field' do
+        @session.visit('/with_js')
+        @session.fill_in('with_change_event', with: '')
+        # click outside the field to trigger the change event
+        @session.find(:css, 'body').click
+        expect(@session).to have_selector(:css, '.change_event_triggered', match: :one)
+      end
+    end
+
+    context "#fill_in with { :clear => :backspace } fill_option", requires: [:js], focus: true do
```
ineverov: remove focus, pls

twalpole: Remove the `focus` metadata

spec/shared_selenium_session.rb
```diff
@@ -88,13 +98,13 @@
         expect(@session.find(:css, '.change_event_triggered', match: :one)).to have_text 'some value'
       end
 
-      it 'should trigger change when clearing field' do
+      it 'should trigger input event field_value.length times' do
```
twalpole: Please add this as a new test and leave the previous test as well

twalpole:
Ok -- this looks good.  The Travis failures are due to issues with Chrome and Travis which I haven't looked into yet - please squash down to one commit and then I'll merge -- Thanks

twalpole:
Also, if you rebase to master when squashing the commits the travis tests should then pass

gsilis:
@twalpole done. Thanks for the help!

twalpole:
Thanks

PR #1917, opened 2017-10-02
Attempt to move element into view when selenium doesn't correctly do it

twalpole:
On `click` when Chrome returns with an error about another element receiving the click this will attempt to move the target element to the middle of the page so it is out from behind fixed headers/footers - and then allow the click to retry.  Unfortunately current versions of Firefox with geckodriver don't return any error in this case, they just click the wrong element and silently move on so there's nothing we can really do for that.


twalpole:
This works with Firefox nightly (as of Oct 2, 2017) if the `moz:webdriverClick`  capability is enabled

PR #1933, opened 2017-11-01
Fix broken README syntax highlighting and add missing

alyssais:
The README contained a couple of markdown syntax errors that broke syntax highlighting on GitHub. I fixed those, then added it to the fenced code blocks that were missing it.


twalpole:
Thanks

PR #1947, opened 2017-12-26
Test and document has_select? for no selected options.

ybakos:
Typically, we use `has_select? 'foo', selected: 'a'` to determine if a select
element 'foo' has an option 'a' with the `selected` attribute. However, we can
also invoke `has_select? 'foo', selected: []` to determine that none of the
options has the `selected` attribute. This is handy, but not documented nor
tested.

Test and document `has_select? 'foo', selected: []`.

### DIFF
--- a/lib/capybara/node/matchers.rb
+++ b/lib/capybara/node/matchers.rb
@@ -423,6 +423,10 @@ module Capybara
       #
       #     page.has_select?('Language', selected: ['English', 'German'])
       #
+      # Or, that no options should have the selected attribute:
+      #
+      #     page.has_select?('City', selected: [])
+      #
       # It's also possible to check if the exact set of options exists for
       # this select box:
       #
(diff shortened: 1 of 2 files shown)


twalpole:
Thanks but this will only work in the rack_test driver because real browsers automatically select the first option if none is selected (hence why the tests are failing)

ybakos:
@twalpole Ah, so is this "feature" really just "undocumented behavior with the rack_test driver" and as such, should remain undocumented and this PR closed?

(The rspec suite passed locally, I overlooked the CI failure.)

twalpole:
@ybakos Correct, it should remain undocumented and this PR closed -- I'm curious what you're running that passed though.  Probably just `rspec`  I guess since I think that will only run the rack_test tests.  if you run `rake` it should run the selenium tests with FF (same as `rake spec_marionette`) and `rake spec_chrome` should run the selenium tests with chrome

ybakos:
@twalpole Yeah I just ran `rspec` hence the pass. Thanks for educating me.

PR #1948, opened 2017-12-26
Fix spelling: non-existant -> non-existent.

ybakos:


twalpole:
Thanks for the this, however could you please get rid of the hyphens between non and existent too - the word is no longer hyphenated in general usage.

ybakos:
@twalpole Good catch, I just force-pushed a new commit here on this PR. Also caught a grammar error (unexistent -> nonexistent).

twalpole:
Thanks!

PR #1950, opened 2017-12-30
Use backslashes for file path for Marionette driver on Windows

hron:
`attach_file` doesn't work on Windows when Marionette driver is in use.
It expects a pathname in backslash notation. For example if you try to
execute `attach_file("D:/Folder/Example.txt")` the driver returns the
following:

```ruby
  1) Capybara::Session selenium #attach_file with normal form should set a file path by id
     Failure/Error: native.send_keys(*path_names)

     Selenium::WebDriver::Error::InvalidArgumentError:
       File not found: D:/Folder/Example.txt
     # WebDriverError@chrome://marionette/content/error.js:235:5
     # InvalidArgumentError@chrome://marionette/content/error.js:362:5
     # interaction.uploadFile@chrome://marionette/content/interaction.js:360:11
     # ./lib/capybara/selenium/node.rb:61:in `set'
     # ./lib/capybara/node/element.rb:108:in `block in set'
     # ./lib/capybara/node/base.rb:85:in `synchronize'
     # ./lib/capybara/node/element.rb:106:in `set'
     # ./lib/capybara/node/actions.rb:256:in `attach_file'
     # ./lib/capybara/session.rb:808:in `block (2 levels) in <class:Session>'
     # ./lib/capybara/spec/session/attach_file_spec.rb:13:in `block (3 levels) in <top (required)>'
```

### DIFF
--- a/lib/capybara/selenium/node.rb
+++ b/lib/capybara/selenium/node.rb
@@ -55,6 +55,8 @@ class Capybara::Selenium::Node < Capybara::Driver::Node
         path_names = value.to_s.empty? ? [] : value
         if driver.chrome?
           native.send_keys(Array(path_names).join("\n"))
+        elsif driver.marionette? && Gem.win_platform?
+          native.send_keys(*Array(path_names).map {|p| p.tr('/', '\\')})
         else
           native.send_keys(*path_names)
         end


twalpole:
Is this a recent change in marionette on windows?  I find it strange this could have been broken for any length of time with no one reporting it.  If it is a recent change in marionette then we would need to still support previous versions of FF/geckodriver too and this is probably better fixed in selenium-webdriver than Capybara.

hron:
I guess no one uses ruby on Windows except me that's why there is no report for this ;] I'm going to continue my research to find out if selenium-webdriver is more proper place to fix this. 

BTW, are you interesting in set up CI to run Capybara's test suite on Windows? I tried to run it locally and there are some failing tests including this one with `attach_file`. If you think it could be useful, I can work on this on a separate branch.

twalpole:
@hron Yeah - I think either selenium or geckodriver should be handling this, and I would think this is a new change in one of them otherwise there would have been other reports before.  

Getting the test suite running on Windows would be interesting although I don't believe Travis supports windows.

hron:
@twalpole I've delved  a bit in selenium-webdriver codebase and it seems the code assumes that using  backslashes in the file path on uploading is capybara responsibility. Check out this test:  https://github.com/SeleniumHQ/selenium/blob/66ca693a8f9578d953d8cb25411b6e78ead12acd/rb/spec/integration/selenium/webdriver/element_spec.rb#L79

I'm not an expert, but I think selenium-webdriver operates more low level API. Capybara calls the same method (`send_keys`?) for both file uploading and text filling. Thus maybe it's better to have this issue fixed in the `attach_file` method of Capybara?

twalpole:
@hron Ok -- but your PR only fixes it for Firefox when running with marionette.   Does it also need changing when using Chrome and legacy FF ??  I'm reticent to change this without actually testing the different combinations on Windows and a few confirmations of the necessity.  Also, if you just call `attach_file("D:\Folder\Example.txt")` does it work for you?

hron:
@twalpole Chrome doesn't need this fix. It works just fine on Windows for path with slashes. I didn't test it on legacy FF yet. 

`attach_file("D:\Folder\Example.txt")` works on my project, but only on Windows, this would break the tests on Linux. So, I've ended up with a workaround that converts slashes if the tests are running on Windows. However I believe this should be fixed on Capybara. Because even this a very straightforward code doesn't works out of the box: `attach_file(File.join('..', 'file.txt'))`. 

I understand your reticent about this change. Thus I propose to setup CI for Windows using [Appveyor](https://www.appveyor.com/). As I mentioned before I am happy to help with it. Just let me know how to approach this better. Should I work on this branch until I fix all failing tests on Windows?

twalpole:
So the fact that it works just fine with Chrome via Selenium makes me not want to fix this in Capybara even more.  That means geckodriver is behaving differently than chromedriver, and without reading the spec I'm not really sure who is correct or not.  Can you post an issue over on the chromedriver project and find out what their perspective on it is?

twalpole:
One other thing to consider here is that when Selenium is run in a remote or grid system the filename may actually be being sent to a system running on a different OS, which is another reason why this really shouldn't be fixed in Capybara.

twalpole:
@hron Did you ever make any progress on running Capybara CI on windows?

hron:
@twalpole Oh. I have forgotten about this. ;) However I'm still interesting in getting this done. It seems I can work on this during the next weekend. Will get back to you with some results on Monday.

Sounds good?

twalpole:
@hron Sounds ok -- I'd like to get a test running against IE before the Capybara 3.0 release just to have some idea what compatibility is like (won't be something that holds up the release though)

PR #1952, opened 2018-01-17
show attempted url when setting invalid app_host or default_host

adsteel:
I've encountered this warning recently in tests that only run on a CI container. It seems like a more informative error message would help folks more generally, too.


adsteel:
I am not sure why the AppVeyor build failed. It does not seem related to this change.

twalpole:
You can ignore appveyor - it's not setup yet.

PR #1956, opened 2018-01-31
Add No Sandbox to Selenium Chrome Headless

metaskills:
In some non-Rails projects I like using Capybara defaults and testing on TravisCI. However, I found this memory issue to be an ever constant problem https://github.com/travis-ci/travis-ci/issues/7150

Can we add `--no-sandbox` to the default options here for the `:selenium_chrome_headless` driver? Are there drawbacks to this?

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -478,5 +478,6 @@ Capybara.register_driver :selenium_chrome_headless do |app|
   browser_options = ::Selenium::WebDriver::Chrome::Options.new
   browser_options.args << '--headless'
   browser_options.args << '--disable-gpu'
+  browser_options.args << '--no-sandbox'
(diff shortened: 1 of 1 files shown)


elliterate:
For what it's worth, sandboxed Chrome recently stopped working altogether in their container-based infrastructure (see: travis-ci/travis-ci#8836). [Per their documentation](https://docs.travis-ci.com/user/chrome), it now requires the use of their sudo-enabled infrastructure.

twalpole:
--no-sandbox is not required on every platform and can be considered a security issue depending on what the browser is being used for.  If the issue is with `travis` then `travis` should fix their container configurations.  We're not going to be adding it to the default configuration at the current moment in time.

metaskills:
Thanks y'all for the feedback and considering it!

PR #1965, opened 2018-02-14
Retry navigation if unload alert disappears

elliterate:
When a Selenium driver interaction raises an unexpected alert error, Firefox 59+ [automatically dismisses the user prompt][1] that caused it. Therefore, if the driver reset triggers an unload alert, we cannot wait to handle it until we receive an unexpected alert error. Instead, we must retry navigation and proactively attempt to accept any alert that appears.

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1416284

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -142,7 +142,20 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
         @browser.switch_to.alert.accept
         sleep 0.25 # allow time for the modal to be handled
       rescue modal_error
-        # The alert is now gone - nothing to do
+        # The alert is now gone
+
+        if navigated
+          # The alert appeared after attempting navigation but disappeared before we could accept it
+
+          # Try to navigate again, anticipating the alert this time
+          begin
+            accept_modal(nil, wait: 0.1) do
+              @browser.navigate.to("about:blank")
+            end
+          rescue Capybara::ModalNotFound
+            # No alert appeared this time
+          end
+        end
       end
       # try cleaning up the browser again
       retry


elliterate:
This fixes the "Timed out waiting for Selenium session reset" errors in the builds using Firefox Beta and Firefox Nightly.

twalpole:
Has this been reported to firefox/geckodriver?  Is it intentional behavior?

elliterate:
Yes and yes. I reported it as mozilla/geckodriver#1171 and was informed that it was deliberately implemented in https://bugzilla.mozilla.org/show_bug.cgi?id=1416284.

elliterate:
BTW, I realized after I submitted the pull request that my implementation wouldn't be sufficient, since it adds a delay to every session reset to handle a rare scenario. I'll keep working on it, but I'm also happy to hear any thoughts you have on how best to handle this.

(Also, as the build shows, it appears to be getting stuck on some test.)

twalpole:
Well that's depressing behavior decision - So you tell the browser to go to another page and FF decides to auto-dismiss any confirm/prompt (rather than accept) and prevent the page change - why would anyone want that as the default?  It also obviously doesn't align it with chromedriver (as stated) or things would still work (since it currently works fine with chromedriver)

twalpole:
@elliterate Yeah -- it would be better to catch the error and then try navigating agin within accept_modal so the sleep isn't always needed.

elliterate:
>So you tell the browser to go to another page and FF decides to auto-dismiss any confirm/prompt (rather than accept) and prevent the page change

It doesn't actually auto-dismiss *on navigation*—only if you do something that triggers an unexpected alert error. But that's exactly what Capybara does when it proceeds to poll for an empty page body.

And because it auto-dismisses the alert at the same time that it raises an unexpected alert error, attempting to accept the alert in the error handler will no longer work. You'll have to re-attempt the navigation, then immediately go looking for an alert to accept.

>Well that's a depressing behavior decision

We're in violent agreement.

But it might get worse. According to the latest draft WebDriver specification, they intend for user prompts triggered by `beforeunload` handlers to [*always* be dismissed][1], which, if my understanding is correct, means that [one would effectively be trapped on such a page][2].

[1]: https://github.com/w3c/webdriver/commit/276a514ed32a3ae73ceb0b6dbb1deb32ed81c80e#diff-ce5a67d95770dc70b1b4250c7a9348f2R5453
[2]: https://github.com/w3c/webdriver/issues/1227

twalpole:
@elliterate Ugh --- Definitely sounds like that would be the case, hopefully someone pays attention to your issue report - I often have the feeling the people writing the spec don't actually ever use the things they're writing the specs for.

elliterate:
While I've updated my branch with what I believe is a reasonable solution, I discovered mozilla/geckodriver#1181 in the process, which likely blocks me from verifying it with passing tests.

twalpole:
@elliterate The beta build still fails, are you saying that's because of the geckodriver issue? If so tis there any benefit at all to merging this at the current time?

elliterate:
>The beta build still fails, are you saying that's because of the geckodriver issue?

That's correct.

>If so tis there any benefit at all to merging this at the current time?

I don't believe we'd see an immediate benefit, no. That issue is preventing this fix from being useful, as it's preventing the `find_xpath('/html/body/*')` from raising an unexpected/unhandled alert error.

twalpole:
Closing in favor of a66ff2eb8cb1abc7a2426c952ff77b1ad234a01f

Issue #1971, opened 2018-02-16
Finders#first returns nil rather than raising exception


ms-ati:
## Meta
Capybara Version:
2.17.0 (latest release), also 2.15.4 

Driver Information (and browser if relevant):
ChromeDriver 2.33 with Chrome 62.0.3202.94-1 on Linux

## Expected Behavior
Per [the API documentation](http://www.rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Finders#first-instance_method) for `Finder#first`: we would expect that, if no matching element can be found before the `Capybara.default_max_wait_time`, that capybara would raise `Capybara::ElementNotFound`.

## Actual Behavior
In practice, it appears that `#first` will return `nil` instead of raising the error. It's not clear if this was previously the *intended* behavior and the intent changed between versions, because while the code docs definitely document the expected error, they also still say "The found element or nil" at https://github.com/teamcapybara/capybara/blob/master/lib/capybara/node/finders.rb#L274 -- so there's some ambiguity about the expected behavior even with in the documented API?

## Steps to reproduce

```ruby
element = first("td", text: "WontBeFound")
#=> nil
```

ms-ati:
Digging deeper, am I correct in my reading that, apart from ambiguity in the doc comment, this was actually fixed in https://github.com/teamcapybara/capybara/commit/d873b83b2aeba0c479bc589a8542072c7b5d69eb#diff-e0a7e0b5db6bbb0355a03b1e6ee1e3b0 on Nov 14 2017?

ms-ati:
...and am I also correct that this fix is now awaiting an actual release version, as [2.18.0](https://github.com/teamcapybara/capybara/blob/2.18.0/lib/capybara/node/finders.rb) doesn't have this fix no the spec driving it?

twalpole:
You're linking to the docs for the master branch, which is the for the upcoming v3 --  If you look at the documentation for the 2.x release versions you'll see the current behavior is as documented - http://www.rubydoc.info/gems/capybara/Capybara/Node/Finders#first-instance_method.  This is one of the things changing for 3.0 and is mentioned in the https://github.com/teamcapybara/capybara/blob/master/History.md for v3.0

ms-ati:
Ah cool, thanks! Well looking forwards to 3.0 then. @twalpole before closing this, are you sure you don't want to fix the ambiguous doc comment that still exists in master ("...or nil")?

twalpole:
[...]

PR #1972, opened 2018-02-16
[comments only] Clarify semantics of Finders#first

ms-ati:
Proposed change is only to documentation comments, to clarify the semantics
around exactly when and how the `Finders#first` method can be changed from
the default behavior (of throwing an exception) to the alternate behavior (of
returning `nil`) when no matching element is found before the max wait time
expires.

cc @twalpole

See original discussion on #1971

### DIFF
--- a/lib/capybara/node/finders.rb
+++ b/lib/capybara/node/finders.rb
@@ -265,14 +265,17 @@ module Capybara
       # Find the first element on the page matching the given selector
       # and options. By default `first` will wait up to `Capybara.default_max_wait_time`
       # seconds for matching elements to appear and then raise an error if no matching
-      # element is found
+      # element is found. Providing the option `minimum: 0` changes this behavior to
+      # instead return `nil` if no matching elements appear in time.
       #
       # @overload first([kind], locator, options)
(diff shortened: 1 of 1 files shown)


twalpole:
Have you confirmed passing `minimum:0` actually works like this? As I mentioned in the other discussion that was my gut reaction to how it would work, but needed to verify.

twalpole:
Also, for non-code changes please include [ci skip] in the commit description

ms-ati:
@twalpole please see https://github.com/teamcapybara/capybara/blob/9d3fd642378d675fc559fd123efabb6141d134c0/lib/capybara/spec/session/first_spec.rb#L18

twalpole:
@ms-ati AH -- I already have a test for that behavior - I really should add for :count, and :between

ms-ati:
@twalpole were you planning to merge this comment change? Or waiting on me for something?

twalpole:
@ms-ati Sorry -- though I commented -- closing in favor of PR #1973 which add more tests to confirm the full behavior of this.

ms-ati:
Ah ok, thanks for solving it then

twalpole:
Merged PR #1973

PR #1977, opened 2018-02-23
Skip unload alert tests for Marionette

elliterate:
Until mozilla/geckodriver#1181 is resolved, Capybara cannot reliably reset a page with an unload alert.

### DIFF
--- a/lib/capybara/spec/session/reset_session_spec.rb
+++ b/lib/capybara/spec/session/reset_session_spec.rb
@@ -40,6 +40,7 @@ Capybara::SpecHelper.spec '#reset_session!' do
   end
 
   it "handles modals during unload", requires: [:modals] do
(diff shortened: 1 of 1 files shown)


elliterate:
@twalpole Given that #1965 is blocked on mozilla/geckodriver#1181, I thought it might be worth skipping the troublesome tests for now. Thoughts?

twalpole:
Pretty sure you'll need to use `skip` rather than `pending` (pending still runs the tests, but expect failure, which I think will leave the browser in an unrecoverable state).  Also, if we can figure out how to determine the FF version I'd prefer to only skip for FF 59 if possible.

twalpole:
Closing in favor of c5bde074371fb52436711b2e3b8e0997a6a685f5

PR #1981, opened 2018-03-05
Derived object should call #initialize method of parent class.

voxik:
Not calling #initialize might result in some weird behavior and unexpected access to uninitialized instance variables of object, such as https://github.com/rack-test/rack-test/issues/214


twalpole:
IIRC this would prevent usage with rack_test < 0.8 for no real gain. Update your Capybara.

Issue #1353, opened 2014-07-02
Unicode spaces shouldn't be normalized


abotalov:
HTML 5.1 (http://www.w3.org/TR/html51/single-page.html#space-character) defines two methods for normalizing spaces:
- "skip whitespace" (that normalizes just 5 characters)
- "skip White_Space characters" (that normalizes all Unicode spaces)

Only the first of those algorithms seems to be actually used in the spec.

XML defines only 4 characters as one that should be normalized - http://www.w3.org/TR/xml/#NT-S

See https://gist.github.com/abotalov/19f6f07b2d9cc64f8a59 (HTML), https://gist.github.com/abotalov/df3630771e21bae504c8 (XML) for my evaluations.

What should be changed? (please approve. I think it can be changed in 2.x as it's a bug)
1. `Capybara::Selenium::Node#visible_text` shouldn't do space normalization as Webdriver already does it - https://dvcs.w3.org/hg/webdriver/raw-file/default/webdriver-spec.html#widl-WebElement-getElementText-DOMString)
2. Algorithm of `normalize_whitespace` itself should be changed to do space normalization similarly to how getElementText does it. Some parts of getElementText's algorithm (e.g. processing CSS whitespace style) aren't worth to be checked so the algorithm will be somewhat different.

Initially `normalize_whitespace` was made to handle so many space characters after https://github.com/jnicklas/capybara/issues/767. But it seems to me that this issue is entirely different (or code has changed too much as currently `normalize_whitespace` has nothing to do with `find_link`). Normalization for `find_link` happens in `xpath` gem and #767 is still not fixed and should be fixed there (this is an attempt - https://github.com/jnicklas/xpath/pull/63, however it's incorrect as it shouldn't treat several nbsp as one)

twalpole:
With 3.0 release coming up, I probably need to look at this -- Does anyone have any examples of where Capybara is doing the wrong thing currently and what the correct results should be?

abotalov:
Well, I didn't check it since writing this issue.
But I think the issue is that when HTML contains "a&nbsp;&nbsp;b", Webdriver returns text "a  b", but Capybara returns "a b".

elliterate:
(For those reading on a mobile device, that's two `&nbsp;` codes in the HTML, two standard space characters returned by WebDriver, and one standard space character returned by Capybara.)

twalpole:
[...]

PR #1982, opened 2018-03-06
Change #text to be closer to what a browser displays

twalpole:
Playing with potential changes for Issue #1353

### DIFF
--- a/lib/capybara/spec/session/ancestor_spec.rb
+++ b/lib/capybara/spec/session/ancestor_spec.rb
@@ -17,7 +17,7 @@ Capybara::SpecHelper.spec '#ancestor' do
 
   it "should find the ancestor element using the given locator and options" do
     el = @session.find(:css, '#child')
-    expect(el.ancestor('//div', text: 'Ancestor Ancestor Ancestor')[:id]).to eq('ancestor3')
+    expect(el.ancestor('//div', text: "Ancestor\nAncestor\nAncestor")[:id]).to eq('ancestor3')
   end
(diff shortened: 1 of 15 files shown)


twalpole:
@abotalov Any comments/feedback on this as it relates to Issue #1353

lib/capybara/spec/session/ancestor_spec.rb
```diff
@@ -17,7 +17,7 @@
 
   it "should find the ancestor element using the given locator and options" do
     el = @session.find(:css, '#child')
-    expect(el.ancestor('//div', text: 'Ancestor Ancestor Ancestor')[:id]).to eq('ancestor3')
+    expect(el.ancestor('//div', text: "Ancestor\nAncestor\nAncestor")[:id]).to eq('ancestor3')
```
peetar: After upgrading our gems, we ended up with a few hundred test failures requiring an update like this. (Yes, we foolishly assumed the whitespace normalization would never change and used this pattern all over our test suite). Anybody have a suggestion on how we might solve this without having to manually update a whole bunch of assertions?

twalpole: @peetar You can use `Capybara.default_normalize_ws = true` - although that will most likely go away in Capybara 4 - so you're going to want to update your tests at some point.

peetar: Thanks for the suggestion, might use it as a temporary workaround

PR #1983, opened 2018-03-06
Fix comment typo in `session.rb`

seangoedecke:
"Symbol" was mis-typed as "Symobl". Now it's "Symbol", like it should be.

No code changes.


twalpole:
thanks

Issue #1984, opened 2018-03-06
`session#reset!` should reset all sessions windows


cyrilchampier:
## Real life problem
In a rails test, I open 2 pages. 
On each page, there is some javascript executing. 
This javascript make an ajax request every seconds to our server.
On teardown, I use `Capybara.reset!`, then clear database.
`Capybara.reset!` will stop javascript executing on current_page, but not on second page
If an ajax request comes during database clear, the test will fail.

## Meta
Capybara Version:
2.17
Driver Information (and browser if relevant):
selenium-webdriver 3.7.0 with Chrome

## Expected Behavior
All opened window should be cleared
The test below only test for html because thats what cause me troubles, but it should also test for storage to be cleared.

## Actual Behavior
Only the current window is cleared

## Steps to reproduce
capybara tests:
```rb
  it "reset multiple windows", requires: [:windows] do
    windows = [@session.current_window, @session.open_new_window]

    windows.each do |window|
      @session.within_window(window) do
        @session.visit('/with_html')
        expect(@session).to have_content('This is a test')
        expect(@session.find('.//h1').text).to include('This is a test')
      end
    end

    @session.reset_session!

    windows.each do |window|
      @session.within_window(window) do
        expect(@session.body).not_to include('This is a test')
        expect(@session).to have_no_selector('.//h1')
      end
    end
  end
```

## Possible solutions
* Modify the `Session#reset!`: but we dont know if underlying driver supports windows
* Add a  `Session#reset_windows!` ?

twalpole:
Currently it's the users responsibility to close extra windows opened during a test when done with them.  If we want to change this behavior it would need to be implemented in each drivers #reset! method.

cyrilchampier:
I would be ok to submit a PR for that, but appart from this one '/lib/capybara/selenium/driver.rb',
where would the other drivers be ?

twalpole:
in their own repos - poltergeist, capybara-WebKit, etc

cyrilchampier:
you seems to contribute to these 3 repositories,
would you approve if I submit one pr on each with this windows loop in their respective implementation of `reset!` ?

twalpole:
@cyrilchampier The more I think about it, the more I think we should just be closing extra windows before we reset the one remaining window.  What are your thoughts on doing that rather than resetting them?

cyrilchampier:
This is something I did not yet completely investigated: whether a simple close would work, or if we need to visit `about:blank` first.
[...]

PR #1986, opened 2018-03-12
Driver reset! should close all but 1 window

twalpole:
This should fix issue #1984 for selenium.  Poltergeist/capybara-webkit/etc will need updates.  A different option would be to move this into Session#reset!


abotalov:
I'd like to clarify if it works when non-first window was switched to when `reset!` is invoked.
I'm not sure if Selenium will switch top-level context without invoking "switch to window".

twalpole:
@abotalov Good point

PR #1989, opened 2018-03-16
Windows CI

hron:
- [x] Fix failing tests on Windows
- [x] Setup Appveyor for testing current test suite
- [x] Add specs for Internet Explorer selenium driver

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,14 @@ environment:
   matrix:
     - RUBY_VERSION: 25
       CAPYBARA_FF: true
+    - RUBY_VERSION: 25
+      CAPYBARA_IE: true
     - RUBY_VERSION: 22
+
+matrix:
+  allow_failures:
+    - CAPYBARA_IE: true
+    - CAPYBARA_EDGE: true
 
 install:
   - set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,26 @@
+version: 1.0.{build}-{branch}
+
+cache:
+  - vendor/bundle
+
+environment:
+  matrix:
+    - RUBY_VERSION: 25
+      CAPYBARA_FF: true
+    - RUBY_VERSION: 22
+
+install:
+  - set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
+  - cinst Firefox GoogleChrome
+  - bundle config --local path vendor/bundle
+  - bundle install
+
+build: off
+
+before_test:
+  - ruby -v
+  - gem -v
+  - bundle -v
+
+test_script:
+  - bundle exec rake travis
--- a/lib/capybara/spec/session/attach_file_spec.rb
+++ b/lib/capybara/spec/session/attach_file_spec.rb
@@ -11,19 +11,19 @@ Capybara::SpecHelper.spec "#attach_file" do
 
   context "with normal form" do
     it "should set a file path by id" do
-      @session.attach_file "form_image", attachment_path(__FILE__)
+      @session.attach_file "form_image", with_os_path_separators(__FILE__)
       @session.click_button('awesome')
       expect(extract_results(@session)['image']).to eq(File.basename(__FILE__))
     end
 
     it "should set a file path by label" do
-      @session.attach_file "Image", attachment_path(__FILE__)
+      @session.attach_file "Image", with_os_path_separators(__FILE__)
       @session.click_button('awesome')
       expect(extract_results(@session)['image']).to eq(File.basename(__FILE__))
     end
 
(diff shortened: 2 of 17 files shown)


appveyor.yml
```diff
@@ -0,0 +1,34 @@
...
+    - RUBY_VERSION: 23
+    - RUBY_VERSION: 23
+      CAPYBARA_FF: true
+    - RUBY_VERSION: 22
+    - RUBY_VERSION: 22
+      CAPYBARA_FF: true
+
+install:
+  - set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
+  - cinst selenium-gecko-driver selenium-chrome-driver Firefox GoogleChrome
```
twalpole: If you change the `webdrivers` gem in gemspec from being conditional on ENV['TRAVIS'] to ENV['CI'] then the geckodriver and chrome-driver install shouldn't be necessary

appveyor.yml
```diff
@@ -0,0 +1,34 @@
...
+  - bundle config --local path vendor/bundle
+  - bundle install
+
+build: off
+
+before_test:
+  - ruby -v
+  - gem -v
+  - bundle -v
+  - chromedriver --version
```
twalpole: logging of driver version in the browser specific specs can be made dependent on ENV['CI'] - then this can go away

lib/capybara/spec/session/attach_file_spec.rb
```diff
@@ -11,33 +11,33 @@
 
   context "with normal form" do
     it "should set a file path by id" do
-      @session.attach_file "form_image", __FILE__
+      @session.attach_file "form_image", attachment_path(__FILE__)
```
twalpole: please change the name of `attachment_path` to mention `os`so it's clear it's modifying the file names in an OS specific manner

appveyor.yml
```diff
@@ -0,0 +1,34 @@
+version: 1.0.{build}-{branch}
+
+environment:
+  matrix:
+    - RUBY_VERSION: 25
+    - RUBY_VERSION: 25
+      CAPYBARA_FF: true
```
twalpole: Really don't need to build on all the ruby versions on Windows (if it's a ruby version specific issue it'll fail on the linux build too)-- just go for minimum (2.2.2)  and latest - also don't really need to do FF on every ruby version, just pick one of them.

twalpole:
So it looks like webdrivers can't download MicrosoftWebDriver.exe (for use with edge) automatically, so that needs to be installed via some other means, and then with IE it appears something weird is going on with windows and modals or something - any clue on that?

hron:
Yep, I'm going to fix the MSWebdriver downloading issue tomorrow and give another try in resolving IE driver tomorrow.

hron:
@twalpole So, here is what we get right now. Firefox and Chrome tests are run on Windows CI. 

Edge test suite works on my local machine. There is a lot of failing tests, but the test suite is useful as a starting point for improving the support for this browser. Unfortunately Edge can't be run on Appveyor because Microsoft don't allow to run this browser on Windows Server.

IE tests don't work and fails in a very strange way. It seems that Capybara can make some calls and then it loses control over the driver. The symptoms are the same for my local machine and CI. 

I suggest to merge this branch in its current state. This will allow testing of Firefox and Chrome on Windows. In the meantime, I'm going to work on IE driver fix separately. As far as I can see from selenium docs there is a lot of pitfalls in getting IE to work properly. 

https://github.com/SeleniumHQ/selenium/wiki/InternetExplorerDriver

Let me know what do you think. Thanks!

twalpole:
The IE/Edge news is bit disappointing (but not all that surprising).  This is still a good step in the right direction though, so if you could please squash it to at most 2 or 3 commits then I'll go ahead and merge -- thanks!

hron:
@twalpole it's done!

twalpole:
@hron Any progress on the IE tests?  If not, it may not make sense to run them at all since every test is failing and it's not really telling us anything about the state of Capybara, just increasing the suite run time on appveyor.  On the Edge front, if there's a different CI provider that provides free runs for OSS projects and allows for Edge to run there's no reason we need to stick on appveyor.

hron:
@twalpole I've managed to run IE tests locally last weekend. There are ~140 failing tests with js and modals filtered out. I still have no clue why all tests fail on appveyor though. Okay, I will take a look at other CIs available for github.

twalpole:
@hron  I got the IE tests running on appveyor by setting the necessary registry keys to keep the connection open - https://ci.appveyor.com/project/twalpole/capybara/build/1.0.204-windows_modals/job/aqiecpe9k4rvaotr

hron:
Great news! For some reason this didn't work for me. Probably I misspelled the registry key. Anyway that's great. The failing tests are the same as I have locally.

PR #1994, opened 2018-03-31
Fix running IE tests on CI

hron:
This pull request incorporates all fixes for ability to run IE tests on CI. There are some failing tests still, but they are not related to CI configuration. I think it makes sense to work on them separately.

I'm also not sure about passing `capybara_skip: skipped_tests` to definition of browser specs. 

There is a fix for `webdrivers` gem for sorting IEDriver versions properly. I guess it's fine to have it from my github profile while the pull requests is not yet accepted.

Anyway let me know what you think!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ gem 'bundler', '~> 1.1'
 gemspec
 
 gem 'xpath', git: 'git://github.com/teamcapybara/xpath.git'
+gem 'webdrivers', git: 'git://github.com/hron/webdrivers', branch: 'iedriver-sorting-fix' if ENV['CI']
 
 group :doc do
   gem 'redcarpet', platforms: :mri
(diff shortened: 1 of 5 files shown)


twalpole:
Why are you skipping the ‘js’ tests? Does the IE driver really not support running JS?

hron:
As far as I can understand IE driver doesn't support JS. Check the IE driver compatibilities: https://github.com/SeleniumHQ/selenium/blob/9ac652d4975e0cea3ea8e0ffe60772c7ef64b513/rb/lib/selenium/webdriver/remote/capabilities.rb#L109

twalpole:
Hmmm, that would make it a pretty pointless driver then, and probably not worth running on App Store at all.  Do all the JS requiring tests fail if not skipped?

hron:
I've just tried to run js tests and it seems that some of them actually works. 

```bash
97 examples, 51 failures, 46 passed
```

twalpole:
Ok  - merging - thanks!

PR #2002, opened 2018-04-06
Docs: Mention breaking change to has_text?

jaredbeck:


twalpole:
The base cause of this is change in behavior in `Node#test` so maybe that should be what's mentioned in this change along with the fact that it affects `has_text?`, `assert_text`, etc...

jaredbeck:
> The base cause of this is change in behavior in Node#text so maybe that should be what's mentioned in this change along with the fact that it affects has_text?, assert_text, etc...

I think UPGRADING.md should focus on the methods people are most likely to use, whichever those may be. I'm happy to accept edits to this PR to that effect.

twalpole:
Closing in favor of 649d7ba4ca8593131d2fc490640b703f0596e214

PR #2004, opened 2018-04-07
Set puma to leak_stack_on_error

swrobel:
Currently puma doesn't return full stacktraces when exceptions happen. I believe this is desirable behavior for test suites, so the browser doesn't just render the [generic](https://github.com/puma/puma/blob/5fb2df9144803f0f80ec1d9fb19d20d433706fef/lib/puma/server.rb#L873), and unhelpful "An unhandled lowlevel error occurred. The application logs may have details."

This could, of course, be made configurable, but I think this is a sane default that I can't imagine anyone would want to change.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -458,6 +458,7 @@ Capybara.register_server :puma do |app, port, host, **options|
     s.binder.parse conf.options[:binds], s.events
     s.min_threads = conf.options[:min_threads]
     s.max_threads = conf.options[:max_threads]
+    s.leak_stack_on_error = true
   end.run.join
 end
 


twalpole:
Capybara catches app errors and re-raises then in the test thread so I don't believe this would make any difference for that.  Additionally `leak_stack_on_error` appears to default to true in Puma::Server so this doesn't appear it would make any difference.  If you can provide an example showing my analysis of this code is wrong I'd be happy to consider the PR.

swrobel:
Not if you set `Capybara.raise_server_errors = false` which I do for system tests so I can look for exceptions being raised without the whole test blowing up. In production these exceptions show error pages. This may be moot anyway because I have a [PR in with Puma](https://github.com/puma/puma/pull/1557) to fix this at the source.

swrobel:
Just to clarify, before:
failures_r_spec_example_groups_admin_when_auth_d_as_a_non_admin_renders_forbidden_529
After:
failures_r_spec_example_groups_admin_when_auth_d_as_a_non_admin_raises_an_exception_648

twalpole:
Ok - but Capybara 3 no longer uses the Puma Runner class so I'm unclear how `leak_stack_on_error` would ever get set to false, since it defaults to true in the Server class

swrobel:
You, sir, are quite correct. I was using Capybara 2.x locally. Upgrading to 3.x fixed the issue.

PR #2012, opened 2018-04-14
Use javascript to set value for date, time and datetime inputs

hron:
According to the documentation `input[type=date]` and others of the
family represent their values differently depending on browser's
language settings. The only reliable way to set value for such kind
of inputs is doing so with help of Javascript.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date
https://developers.google.com/web/updates/2012/08/Quick-FAQs-on-input-type-date-in-Google-Chrome

### DIFF
--- a/lib/capybara/selenium/node.rb
+++ b/lib/capybara/selenium/node.rb
@@ -1,18 +1,7 @@
 # frozen_string_literal: true
 
 class Capybara::Selenium::Node < Capybara::Driver::Node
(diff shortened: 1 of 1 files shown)


twalpole:
Using JS for this is not acceptable since it doesn’t trigger the correct events.  If you want to submit a PR that adds support for different keystrokes for different languages that would be considered

Issue #2011, opened 2018-04-14
Selenium with headless Chrome not including 


darraghbuckley:
## Meta
Capybara Version:
3.0.2
Driver Information (and browser if relevant):
selenium-webdriver

## Actual Behavior
``NameError: uninitialized constant Selenium
from /Users/darragh/.rvm/gems/ruby-2.4.2/gems/capybara-3.0.2/lib/capybara.rb:497:in `block in <top (required)>``

## Steps to reproduce
```
require 'capybara'

app = proc { |env| [200, { "Content-Type" => "text/html" }, [''] ] }

Capybara.server = :webrick
session = Capybara::Session.new(:selenium_chrome_headless, app)
session.visit("/")
```

The issue is specific to `selenium_chrome_headless`. The error looks to be because `selenium-webdriver` hasn't been `require`'d when it's referenced here:

https://github.com/teamcapybara/capybara/blob/master/lib/capybara.rb#L497

That gem is usually lazy-loaded when the driver is initialized here:

https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selenium/driver.rb#L476

This looks to have been introduced in this change:

https://github.com/teamcapybara/capybara/commit/0275eab42c610cd1ccde7947b051d0b00857b9ce

The error looks to not be raised in tests because the tests independently require the gem here:

https://github.com/teamcapybara/capybara/blob/master/spec/selenium_spec_chrome.rb#L4

I can fix it on my end by requiring `selenium-webdriver` before touching Capybara. That's a satisfactory solution for me but it's kinda out of the pattern of the other drivers. If this is enough of an issue where you want a fix, I'm happy to put in a Pull Request. Otherwise, count this issue more as an FYI and feel free to close it.

twalpole:
If you have time please submit a PR, otherwise I'll fix this sometime next week.

twalpole:
Closed via 431482b751e5a65e9902eca1ece7abbd36646bee

PR #2013, opened 2018-04-15
Use `load_selenium` class method for requiring `selenium-webdriver`.

darraghbuckley:
This change relates to this issue:

https://github.com/teamcapybara/capybara/issues/2011

`Capybara::Selenium::Driver#load_selenium` is bumped to a class method which is
called either during `Selenium::Driver` instantiation or before Selenium is
used. Tests have also been updated to remove now unnecessary requires of
`selenium-webdriver`.

This should make `Selenium::Driver` behave more like `RackTest::Driver`. It also
has the benefit of making sure the misnamed Selenium errors are patched
~everywhere.

This change passes local tests for me that were run with:

`bundle exec rake chrome_spec`

`CI=true bundle exec rake travis`

All in all, I'm not sure the lazy-loading is worthwhile (though that's, of
course, a choice for the maintainers).

A small aside, that there's also the use of the `webdrivers` gem which is
conditional on the `CI` environment variable. This means that CI behaves
differently to other environments in a way that's perhaps not expected (though
the failure case looks to be that tests would succeed in CI but fail locally,
which seems better than the opposite).

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -55,7 +55,7 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
   end
 
   def initialize(app, **options)
-    Capybara::Selenium::Driver.load_selenium
+    self.class.load_selenium
     @session = nil
     @app = app
     @browser = nil
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -13,6 +13,23 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
 
   attr_reader :app, :options
 
+  def self.load_selenium
+    require 'selenium-webdriver'
+    # Fix for selenium-webdriver 3.4.0 which misnamed these
+    unless defined?(::Selenium::WebDriver::Error::ElementNotInteractableError)
+      ::Selenium::WebDriver::Error.const_set('ElementNotInteractableError', Class.new(::Selenium::WebDriver::Error::WebDriverError))
+    end
+    unless defined?(::Selenium::WebDriver::Error::ElementClickInterceptedError)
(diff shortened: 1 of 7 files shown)


lib/capybara/selenium/driver.rb
```diff
@@ -38,7 +55,7 @@ def browser
   end
 
   def initialize(app, **options)
-    load_selenium
+    Capybara::Selenium::Driver.load_selenium
```
twalpole: self.class.load_selenium

spec/selenium_spec_ie.rb
```diff
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
-require "selenium-webdriver"
 require 'shared_selenium_session'
 require 'rspec/shared_spec_matchers'
 
 Capybara.register_driver :selenium_ie do |app|
   # ::Selenium::WebDriver.logger.level = "debug"
+  Capybara::Selenium::Driver.load_selenium
```
twalpole: This isn't necessary since the Capybara::Selenium::Driver initialization calls it

spec/shared_selenium_session.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
-require "selenium-webdriver"
```
twalpole: See above

spec/selenium_spec_marionette.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
-require "selenium-webdriver"
```
twalpole: See above

twalpole:
I commented on some things, basically this is fine but I don't really want the changes to the test files -  just leave them how they were.

darraghbuckley:
Updated -- thanks again for Capybara!

twalpole:
Thanks for this

PR #2019, opened 2018-04-24
[WIP] Run Chrome remote configuration in CI

ianks:
This is a WIP PR to start running selenium with the `remote` browser type on CI. This will most likely cause some breakage; which I will address before this is merged :grinning:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: ruby
 sudo: required
 dist: trusty
+services:
+  - docker
 rvm:
   - 2.5.0
   - 2.3.6
@@ -30,6 +32,9 @@ cache:
 
 matrix:
   include:
+    - gemfile: Gemfile
+      rvm: 2.5.0
+      env: CAPYBARA_CHROME_REMOTE=true
     - gemfile: gemfiles/Gemfile.rspec-34
       rvm: 2.3.6
       env: CAPYBARA_FF=true
@@ -66,6 +71,7 @@ matrix:
     - gemfile: gemfiles/Gemfile.beta-versions
     - gemfile: gemfiles/Gemfile.edge-marionette
     - rvm: rbx-3
+    - env: CAPYBARA_CHROME_REMOTE=true
 before_install:
   - gem update --system
   # - gem update bundler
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -85,4 +91,8 @@ before_script:
       echo "require('awful'); tags = {}; tags[1] = awful.tag({'name'})" > ~/.config/awesome/rc.lua;
       awesome &
     fi
+  - if [[ -z "${CAPYBARA_CHROME_REMOTE}" ]]; then
```
twalpole: This check is the wrong direction - it's running docker when not needed, and not when it is needed

Gemfile
```diff
@@ -10,3 +10,7 @@ gem 'xpath', git: 'git://github.com/teamcapybara/xpath.git'
 group :doc do
   gem 'redcarpet', platforms: :mri
 end
+
+group :test do
```
twalpole: Please remove this change

.travis.yml
```diff
@@ -77,12 +83,23 @@ before_install:
       bundle config local.selenium-webdriver ../selenium/build/rb;
     fi
 before_script:
-  - if [[ -z $HEADLESS ]]; then
+  - if [[ -z $HEADLESS && -z $CAPYBARA_CHROME_REMOTE ]]; then
```
twalpole: Did you mean || here?

twalpole: Nevermind- misread it

Rakefile
```diff
@@ -36,12 +38,16 @@ Cucumber::Rake::Task.new(:cucumber) do |task|
 end
 
 task :travis do
+  rspec_opts << "--format documentation"
```
twalpole: Is there an actual reason for this change? or was it just for debugging?

ianks: Debugging. Although realistically we might want to switch to `--format process` since `fuu` requires a raw TTY which is incompatible with Travis. Thoughts?

twalpole: I don't think it's necessary to switch in this PR

twalpole:
I see this is passing now, if it's complete please squash the commits and then I'll go ahead and merge.

twalpole:
Thanks!

Issue #1713, opened 2016-06-17
Not working with Thin


stevebooks:
When running the following script:
```
require "capybara"

Capybara.server do |app, port|
   require 'rack/handler/thin'
   Rack::Handler::Thin.run(app, :Port => port)
end

html = DATA.read
app = proc { |env| [200, { "Content-Type" => "text/html" }, [html] ] }

sess = Capybara::Session.new(:selenium, app)
sess.visit("/")

__END__

<!doctype html>
<html>
  <body>
  </body>
</html>
```

I can see from the command line that the thin server has booted. I can also manually visit the URL and it works.

Error:
```
Thin web server (v1.7.0 codename Dunder Mifflin)
Maximum connections set to 1024
Listening on localhost:58347, CTRL+C to stop
/Users/stevebooks/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/capybara-2.7.1/lib/capybara/server.rb:112:in `rescue in boot': Rack application timed out during boot (RuntimeError)
```

Gem version: 2.7.1
Ruby version: 2.3.1
Thin Version: 1.7.0

From what I can tell, inside Capybara::Server, the `@server_thread.join(0)` always returns false. This is used to determine if the server is `responsive?`

stevebooks:
More info:
`Capybara.server_host = "localhost"` fixes it. 😞 
This took a while to debug. Anything we can do here so that others don't run into the same issue?

twalpole:
You should be using the 3 parameter version of the server block which also provides the expected host

    Capybara.register_server :thin do |app, port, host|
        require 'rack/handler/thin'
       Rack::Handler::Thin.run(app, :Port => port, :Host => host)
    end

    Capybara.server = :thin

PirunSeng:
@stevebooks I'm facing a similar problem. Can you show me your full configuration and versions please?

@twalpole I've tried your config, but get this error:
`undefined method 'register_server' for Capybara:Module`

Not sure if your solution does not work for me since I have "poltergeist" instead of "selenium".

stevebooks:
@PirunSeng switching to the 3 parameter version worked for me as @twalpole suggested. What version of capybara are you running? Mine is `2.7.1`

PirunSeng:
@stevebooks my capybara is 2.5.
When I have Capybara.register_server ... I get an error "undefined register_server for Capybara module..."

stevebooks:
Try upgrading, they might have changed the API

[...]

PR #2022, opened 2018-04-26
Introduce thin server definition

ignisf:
This commit introduces thin support out of the box.

It features a :Silent option to silence log output in the tests and makes sure
thin does not capture signals we want the testing framework to handle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,13 @@ Capybara.server = :puma # Until your setup is working
 Capybara.server = :puma, { Silent: true } # To clean up your test output
 ```
 
+Or in case you are using thin:
+
+```ruby
+Capybara.server = :thin
+Capybara.server = :thin, { Silent: true } # To silence thin's start message
+```
+
 ## <a name="using-capybara-with-cucumber"></a>Using Capybara with Cucumber
 
 The `cucumber-rails` gem comes with Capybara support built-in. If you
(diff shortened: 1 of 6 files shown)


ignisf:
Not sure how to handle the build failure on jruby... Should I just remove the dev dependency from the .gemspec and Gemfiles?

README.md
```diff
@@ -97,13 +97,22 @@ Capybara.app = MyRackApp
 
 If you need to test JavaScript, or if your app interacts with (or is located at)
 a remote URL, you'll need to [use a different driver](#drivers).  If using Rails 5.0+, but not using the Rails system tests from 5.1, you'll probably also
-want to swap the "server" used to launch your app to Puma in order to match Rails defaults.
+want to swap the "server" used to launch your app to Puma or thin in order to match Rails defaults.
```
twalpole: `thin` is not the Rails default

twalpole:
What is the motivation behind this?  My understanding is that `thin` has relatively low usage level when compared to puma/unicorn/etc. and from the unanswered issues and PRs it doesn't really seem to be maintained much anymore.  Why should we bake this into Capybara when it's easy enough for a user to register their own server if they want to use `thin`?

ignisf:
Hello,

Just thought there should be a suggested way to use Capybara with Rails 5.2 and thin as it seems it's not that simple. I'm fine with this being documented somewhere in Capybara's docs, just so people do not rely on issue discussions (https://github.com/teamcapybara/capybara/issues/1713 for example).

README.md
ignisf: Seems I did not understand the meaning of the sentence correctly.

twalpole:
I'm not going to merge this as is, if you want to submit a PR that is purely documentation I think that would be more appropriate.

PR #2024, opened 2018-04-28
Add a mechanism to fill in `input[type=date]` in Chrome and Edge

hron:
## Abstract 

This is a work in progress change to allow Capybara works properly with `input[type=date]` and others in the family with different locale and date format settings. I've made some research on this and here are my thoughts.

## How different browsers display `input[type=date]`

It seems all browsers do displaying differently. Firefox relies on its internal languages set, Chrome & Edge use OS locale, but not for date picker. Additionally Chrome on Windows uses date short format from OS settings "Change date and time formats".

## How different browsers allow a user to set value for `input[type=date]`

Apparently there is no single way to do it. Firefox magically allows a user to type in date using `%Y-%m-%d` format and everything works regardless of how the controls are displayed. 

Chrome allows typing, but the typed parts go to date's parts and must respect the current date format. Also if the year part is the first one, you must type right arrow to move cursor to the next date part. 

Edge doesn't allow typing date at all! You must use arrows to choose the date.

## The Suggested approach

I suggest to tackle every browser separately. The current pull request is the demonstration for Chrome and Edge, but this approach could be used for all other inputs.

- [x] Fix filling in `input[type=date]` in Chrome and Edge for different date formats
- [ ] Fix filling in `input[type=time]` in Chrome and Edge for different time formats
- [ ] Fix filling in `input[type=datetime-local]` in Chrome and Edge for different time formats

Please let me know if the approach I suggested is good enough and it makes sense to use it for all other inputs from the family.

### DIFF
--- a/lib/capybara/selenium/node.rb
+++ b/lib/capybara/selenium/node.rb
@@ -268,7 +268,7 @@ private
         sequence[positions[2] - 1] = date.day
         sequence.each.with_index do |dp, i|
           native.send_keys :left, :left, :left, *Array.new(i, :right)
-          # When the date format contains month name Chrome needs this sleep to show monthname after number's been typed
+          # When the date format contains month name Chrome needs this sleep to show monthname after number's typed
           sleep 0.5
           native.send_keys :backspace, dp.to_s
         end
--- a/lib/capybara/selenium/node.rb
+++ b/lib/capybara/selenium/node.rb
@@ -253,40 +253,22 @@ private
     yield
   end
 
-  def set_date(value) # rubocop:disable Naming/AccessorMethodName
-    if value.respond_to?(:to_date)
-      date = value.to_date
-
-      case driver.browser.browser
-      when :chrome
-        native.send_keys :left, :left, :left
-        native.send_keys :backspace, '1', :right, :backspace, '2', :right, :backspace, '3', :left, :left
-        positions = driver.evaluate_script("arguments[0].value", self).split('-').map(&:to_i)
-        sequence = []
-        sequence[positions[0] - 1] = date.year
-        sequence[positions[1] - 1] = date.month
-        sequence[positions[2] - 1] = date.day
-        sequence.each.with_index do |dp, i|
-          native.send_keys :left, :left, :left, *Array.new(i, :right)
-          # When the date format contains month name Chrome needs this sleep to show monthname after number's typed
-          sleep 0.5
-          native.send_keys :backspace, dp.to_s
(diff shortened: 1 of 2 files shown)


twalpole:
First reaction is this seems overly complex and that the format sniffing code will probably have an issue if the user has `min`/`max` attributes specified on the input.  Second reaction is that Edge support is not a priority until we can get Capybara testing against Edge in CI somewhere.

hron:
I see eye to eye with you on the concerns, but the alternative approach involves a complex detection of date format on each platform. It could be even more complex that this solution.

Moreover even if the date format sniffing code is removed Capybara can't just use `native.send_keys *formatted_date` for all possible formats. There is cases when year is not on the end and Capybara must use arrow keys to move the cursor because the year part could take 5 digits.

Any advices for me on resolving this issue? Maybe it's better to use JS to set value and trigger the missing 'change' events somehow?

twalpole:
If it's not possible to detect the required format cleanly, or the browsers locale then using JS and triggering the normal events  (probably at least `input` and `change`) might be a solution.  However if it is possible to detect the browsers locale then it should be possible to build up a map of the required formats (I don't believe arrow is necessary since tab should do the same thing, but even if arrow was necessary I'm pretty sure we could support that) and let the user specify the locale to use.  That would be a preferred solution since it would correctly mimc what a user could actually do.

hron:
@twalpole Here is a variant for setting value using javascript and triggering events manually. 

I am researching the date format detection further, but it seems the only way to detect the format is to use `kernel32.dll` on Windows to get access to [LOCALE_SSHORTDATE](https://msdn.microsoft.com/en-us/library/windows/desktop/dd373886%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396). I'm not a C++ expert, but as far as I can see Chrome uses [GetLocaleInfo](https://msdn.microsoft.com/en-us/library/windows/desktop/dd318101(v=vs.85).aspx) to decide how to show the input.

Firefox doesn't use the system locale, but browser's locale affects the format. I'm not sure how Edge detects the format, but it seems it ignores locale at all. ;)

I believe detecting the locale properly would be a more fragile solution than using the specified API. Although I agree it mimics user behavior better.  

Anyway, let me know what you think. If you believe detecting the locale is worth it I will try and implement a solution using that approach.

twalpole:
Mimicing user behavior is definitely preferred here, so if it's possible I would prefer that - If it's not I'm going to leave some comments on your JS implementation.

PR #2026, opened 2018-04-29
Add `without` method to session to temporarily escape from `within`

smoyth:
I have a need for this in a helper method for filling out a select2 box. The first step is to fill in the box itself, which triggers a popup element for doing the typedown search. The popup is inserted dynamically at the foot of the page, outside the within scope. So I need the within for only _part_ of my helper method.

I hope this is making some sense! Let me know what you think!

And thanks for the awesome tool!

### DIFF
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -59,7 +59,7 @@ module Capybara
     SESSION_METHODS = %i[
       body html source current_url current_host current_path
       execute_script evaluate_script visit refresh go_back go_forward
-      within within_element within_fieldset within_table within_frame switch_to_frame
+      within within_element within_fieldset within_table within_frame without switch_to_frame
       current_window windows open_new_window switch_to_window within_window window_opened_by
       save_page save_and_open_page save_screenshot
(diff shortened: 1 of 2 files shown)


twalpole:
The name `without` isn't clear as to what it would do (one level back up the scope tree or back to the document).  That combined with the fact it isn't really a necessary feature means this will not be included, thanks for the effort though.  If you need to get out of a scope you can always just call methods on the return of `document` or use the fact that with `xpath` starting with `//` as opposed to `.//` searches from the document root.

    page.document.find('div#my_div')   # will escape out of any within scope and find from document root

    page.find(:xpath, '//div[@id=my_div]')  # // causes to find from document root no matter what within scope you're in

You can also implement the xpath query using the xpath gem dsl with `XPath.anywhere` as shown in the `:select2_option` selector implemented in the UI plugins experiment PR - https://github.com/teamcapybara/capybara/pull/1980/files

smoyth:
Ah, thanks. Might be nice to add those ways to get out of the `within` somewhere in the docs. I searched for quite awhile for a way to do this...

PR #2037, opened 2018-05-12
Add links for rubygems page

igas:
It would be useful to have direct links to http://rubygems.org/gems/capybara/ page


twalpole:
I don't want to change the homepage link at the moment, but would accept this without that change.

twalpole:
That’s fine, please squash into one commit

Issue #2038, opened 2018-05-14
Default Selenium::Node#set(clear: :option)


cyrilchampier:
## Meta
Capybara Version:
(2.18.0)
Driver Information (and browser if relevant):
selenium-webdriver 3.7.0 / chrome headless

## Expected Behavior
In our app, we make a global use of the `clear: :backspace` behavior implemented here:
`Capybara::Selenium::Node#set('', clear: :backspace)`
https://github.com/teamcapybara/capybara/commit/339bebf55ec45b427970419f02e19a9b2be687f6

Is there any way currently to set this clear behavior by default ?
If not, would you accept a PR to add a `default_clear_option` in `SessionConfig::OPTIONS` ?

twalpole:
Since the options `set` accepts are driver specific I wouldn't accept a PR for `default_clear_option`, I would however accept a PR implmenting `default_set_options` which would be a Hash.

cyrilchampier:
perfect, I'll do that in the next days, thanks :)

twalpole:
Closing this - will wait for PR submission.

PR #2039, opened 2018-05-17
feature: add a default_set_options configuration

cyrilchampier:
relative to https://github.com/teamcapybara/capybara/issues/2038

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -82,6 +82,7 @@ module Capybara
     # [reuse_server = Boolean]  Reuse the server thread between multiple sessions using the same app object (Default: true)
     # [threadsafe = Boolean]  Whether sessions can be configured individually (Default: false)
     # [server = Symbol]  The name of the registered server to use when running the app under test (Default: :webrick)
+    # [default_set_options = Hash]  The default options passed to Node::set (Default: {})
     #
     # === DSL Options
     #
@@ -479,6 +480,7 @@ Capybara.configure do |config|
   config.automatic_label_click = false
   config.enable_aria_label = false
   config.reuse_server = true
+  config.default_set_options = {}
 end
 
 Capybara.register_driver :rack_test do |app|
--- a/lib/capybara/spec/session/node_spec.rb
+++ b/lib/capybara/spec/session/node_spec.rb
@@ -102,6 +102,13 @@ Capybara::SpecHelper.spec "node" do
       expect { @session.first('//textarea[@readonly]').set('changed') }.to raise_error(Capybara::ReadOnlyElementError)
     end
 
+    it 'should use global default options', requires: [:js] do
+      Capybara.default_set_options = { clear: :backspace }
+      element = @session.first('//input')
+      expect(element.base).to receive(:set).with('gorilla', clear: :backspace)
+      element.set('gorilla')
+    end
+
     context "with a contenteditable element", requires: [:js] do
       it 'should allow me to change the contents' do
         @session.visit('/with_js')
--- a/History.md
+++ b/History.md
@@ -1,3 +1,10 @@
+# Version 3.2.0
+Release date: unreleased
+
+### Added
+
+* New global configuration `default_set_options` used in `Capybara::Node::Element#set` as default `options` hash 
+
 # Version 3.1.0
 Release date: 2018-05-10
 
--- a/lib/capybara/node/element.rb
+++ b/lib/capybara/node/element.rb
@@ -84,11 +84,12 @@ module Capybara
       # Set the value of the form element to the given value.
       #
       # @param [String] value    The new value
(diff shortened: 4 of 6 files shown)


cyrilchampier:
this one fails probably because of my dev: https://travis-ci.org/teamcapybara/capybara/jobs/380140664 altough I can reproduce in local,
but I have no clue for those 2:
https://travis-ci.org/teamcapybara/capybara/jobs/380140669
https://travis-ci.org/teamcapybara/capybara/jobs/380140671

lib/capybara/selenium/node.rb
```diff
@@ -33,14 +33,15 @@ def value
...
-  # @param [Hash{}] options  Driver specific options for how to set the value
+  # @param [Hash{}] options  Driver specific options for how to set the value. Take default values from {Capybara#default_set_options}
   # @option options [Symbol,Array] :clear (nil) The method used to clear the previous value <br/>
   #   nil => clear via javascript <br/>
   #   :none =>  append the new value to the existing value <br/>
   #   :backspace => send backspace keystrokes to clear the field <br/>
   #   Array => an array of keys to send before the value being set, e.g. [[:command, 'a'], :backspace]
   def set(value, **options)
     raise ArgumentError, "Value cannot be an Array when 'multiple' attribute is not present. Not a #{value.class}" if value.is_a?(Array) && !multiple?
+    options = Capybara.default_set_options.merge(options)
```
twalpole: This should be implemented in capybara/node/element.rb so it affects all drivers, not in selenium specific code.  Also it should pull from `session_options.default_set_options` not `Capybara_default_set_options` so it works in "threadsafe" mode too - see https://github.com/teamcapybara/capybara/blob/master/spec/per_session_config_spec.rb

lib/capybara/spec/session/node_spec.rb
```diff
@@ -102,6 +102,12 @@
       expect { @session.first('//textarea[@readonly]').set('changed') }.to raise_error(Capybara::ReadOnlyElementError)
     end
 
+    it 'should use global default options', requires: [:js] do
```
twalpole: test needs to reset the setting at the end, or add it to https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/spec_helper.rb#L21

lib/capybara/spec/session/node_spec.rb
```diff
@@ -102,6 +102,12 @@
...
+      Capybara.default_set_options = { clear: :backspace }
+      expect_any_instance_of(Capybara::Selenium::Node).to receive(:set_text).with('gorilla', clear: :backspace)
```
twalpole: Find the element first, and then set the expectation on `element.base` rather than using any_instance_of.  Also you want to expect on :set, not :set_text since set_text is selenium specific.

twalpole:
I'm guessing the failures were Travis hiccups -- it can be flaky when our tests have to rely on timing (in order to test the waiting behavior) - I told them to retry and they passed.  I've added a few comments with needed fixes - also please add an entry to `History.md`

twalpole:
@cyrilchampier Yes - create a new section - Since this is an added feature it needs to be 3.2.0  - date should be 'unreleased'

lib/capybara.rb
```diff
@@ -82,6 +82,7 @@ class << self
     # [reuse_server = Boolean]  Reuse the server thread between multiple sessions using the same app object (Default: true)
     # [threadsafe = Boolean]  Whether sessions can be configured individually (Default: false)
     # [server = Symbol]  The name of the registered server to use when running the app under test (Default: :webrick)
+    # [default_node_set_options = Hash]  The default options passed to Node::set (Default: {})
```
twalpole: Please change this back to `default_set_options` - The only public `set` API is on Node, so there's not need to specify 'node' in the name

twalpole:
@cyrilchampier Looking good other than the needing to reset the setting after the test - once that's fixed please squash all the commits and it should be ready for merge -  Thanks!

cyrilchampier:
@twalpole I already reset it globally here https://github.com/teamcapybara/capybara/pull/2039/files#diff-b89bf9c5aa2c58704dbee3ea98fa438cR32 you also want it to be resetted at the end of the test method ?

twalpole:
@cyrilchampier Ah - missed that because my original comment was still there - all good, then please just squash all the commits

cyrilchampier:
@twalpole thank should be good now, there is only 1 flaky test. Thanks for your patience!

lib/capybara/spec/session/node_spec.rb
twalpole: This was fixed

twalpole:
@cyrilchampier Ok -- once the tests pass I'll merge this in - thanks a lot

PR #2053, opened 2018-06-15
Add SemVer compatibility badge

greysteil:
Adds a SemVer compatibility badge that displays the percentage of CI runs that pass when updating `capybara` between SemVer compatible versions. Data comes from Dependabot (disclosure: I built it). I've also removed the Gemnasium badge, since it now just serves an advert for GitLab 😕.

Here's what the badge looks like:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=capybara&package-manager=bundler&version-scheme=semver)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 [![Build Status](https://secure.travis-ci.org/teamcapybara/capybara.svg)](https://travis-ci.org/teamcapybara/capybara)
 [![Build Status](https://ci.appveyor.com/api/projects/status/github/teamcapybara/capybara?svg=true)](https://ci.appveyor.com/api/projects/github/teamcapybara/capybara)
-[![Dependency Status](https://gemnasium.com/teamcapybara/capybara.svg)](https://gemnasium.com/teamcapybara/capybara)
 [![Code Climate](https://codeclimate.com/github/teamcapybara/capybara.svg)](https://codeclimate.com/github/teamcapybara/capybara)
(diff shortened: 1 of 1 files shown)


twalpole:
Thanks for pointing out the need to remove the Gemnasium badge, but I'm unclear as to exactly what this badge is telling anyone?  Is the number based on other projects that use capybara, or on the CI runs of capybaras own tests, or something else?

greysteil:
It’s other projects using capybara - if you click it there’s an explanation there, but if it’s not clear let me know and I can change the text!

greysteil:
Flaky tests - yep, it's a problem, although we filter out any repos where the base branch isn't passing, so it's only tests that are flaky that are a problem (not ones that are broken).

Links to open source failures - I want to add this next, and ideally build out something that makes it easy to check whether pre-releases work on everyone's repos. In the meantime, you can see all the Capybara PRs that Dependabot has created [here](https://github.com/pulls?utf8=%E2%9C%93&q=is%3Apr+author%3Aapp%2Fdependabot+is%3Aclosed+capybara+in%3Atitle).

twalpole:
Ok -- one more question -- are failures that are not required by semver to be compatible included in the overall percentage - ie Capybara 2.18.0 → 3.0.0 shows a "compatibility" score of 75% -- does that factor in to the overall 97% ?

greysteil:
Those get filtered out of the SemVer score, as does anything pre-1.0.0. We think it’s still interesting to show them, but they’re no part of SemVer compatibility so not part of the overall number.

twalpole:
Ok -- thanks for the PR -- I need to think about it for a little bit

greysteil:
👍 - if you come up with anything you’d like me to change at all then let me know - I want to make something useful here.

twalpole:
After some though I'm ok with this - thanks

PR #2063, opened 2018-06-30
Sync for visibility on make_visible in attach_file

emilong:
I noticed that when I specified `make_visible` in `attach_file`, the file input would become visible, but not immediately. I thought I'd try wrapping the `visible?` call in a `synchronize` and it seems to work in my case.

No idea how I'd add a test for a timing thing like this, but if anyone has suggestions, I'm very grateful. Also grateful to hear if there's another approach I should take in general.

Thanks!

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -231,7 +231,7 @@ module Capybara
         # Allow user to update the CSS style of the file input since they are so often hidden on a page
         if make_visible
           ff = find(:file_field, locator, options.merge(visible: :all))
-          while_visible(ff, make_visible) { |el| el.set(path) }
+          while_visible(ff, make_visible, options) { |el| el.set(path) }
         else
           find(:file_field, locator, options).set(path)
         end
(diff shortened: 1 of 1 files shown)


twalpole:
Do you have any idea why your input doesn't become visible immediately?  I guess possibly if you had a CSS animation on the element it could gradually change the values over time, is that true in your case?  If so I think you'd be better off passing a custom hash to `make_visible`, rather than true, that overrides the animation timing or just changing your CSS so it doesn't apply animation timings to file inputs.

Also rather than adding a synchronize call and passing through options this could just be done as

    raise ExpectationNotMet, "The style changes in :make_visible did not make the file input visible" unless element.matches_selector?(:file_field, visible: true)

emilong:
Thanks for getting back to me! Really good question about why it doesn't become invisible immediately.... I was using a hash for `make_visible` because we're using the a position absolute (`position: absolute; left: -9999999px`) trick to position the element offscreen in an attempt to make it continue to be visible to screenreaders. The hash that I was using was `{ opacity: 1, display: 'block', visibility: 'visible', position: "static" }`.

However, inspired by your comment, I just tried `{ opacity: 1, display: 'block', visibility: 'visible', position: "absolute", left: "0" }` and that seems to have make it work... Maybe the browser is faster at relocating the element when changing the positional value only rather than the whole positioning system?

In any case, this seems to work, so I'll close out this PR. Thanks again for your feedback!

twalpole:
You're welcome -- No clue why that change would affect timing with no CSS animations/transforms - but glad you got it working.  Currently the default CSS used when `make_visible` is true is  `{ opacity: 1, display: 'block', visibility: 'visible' }`, I wonder whether there is something more universal we could set

PR #2064, opened 2018-07-02
Avoid testing control+shift+click in Firefox

elliterate:
Click with the control key triggers a context-click and, [in Firefox, the shift key causes it to bypass JavaScript event handling][1].

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=692139

### DIFF
--- a/lib/capybara/spec/session/node_spec.rb
+++ b/lib/capybara/spec/session/node_spec.rb
@@ -377,6 +377,7 @@ Capybara::SpecHelper.spec "node" do
     end
 
     it "should allow multiple modifiers", requires: [:js] do
+      pending "Firefox doesn't generate an event for shift+control+click" if marionette_gte?(62, @session)
       @session.visit('with_js')
       @session.find(:css, '#click-test').click(:control, :alt, :meta, :shift)
       # Selenium with Chrome on OSX ctrl-click generates a right click so just verify all keys but not click type
--- a/spec/selenium_spec_marionette.rb
+++ b/spec/selenium_spec_marionette.rb
@@ -163,5 +163,13 @@ RSpec.describe Capybara::Selenium::Node do
       tr.click
       expect(tr.base).to have_received(:warn).with(/Clicking the first cell in the row instead/)
     end
+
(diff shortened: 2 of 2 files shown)


twalpole:
This has been passing with Firefox previously, but I see that it's breaking in the next beta.  If FF is going to keep this in the broken state we shouldn't be disabling the test for the rest of the browsers.  We would allow Firefox to skip/pending the test and add a new FF specific test in selenium_spec_marionette.rb

twalpole:
Also, since the element being clicked on captures `click`, `dblclick`, and `contextmenu` and outputs the modifier keys pressed this really sounds likes FF or geckodriver breaking things for no reason (I could see it being correct if just control and shift would pressed during a click - but not when alt, control, meta and shift are pressed)

elliterate:
> This has been passing with Firefox previously, but I see that it's breaking in the next beta.

Per the linked bug report, this "feature" of "shift + context-click" has been present in Firefox for many years. What I suspect happened with the latest beta is simply that they began treating "control + click" as a `contextmenu` event (like Chrome) rather than a `click` event, thereby exposing it to this custom behavior.

If anything, I think this change makes Firefox _more_ internally consistent. I would find it stranger if "shift + ctrl + click" _didn't_ behave the same in the presence of the "meta" or "alt" keys. In the absence of some other specific command mapping, I would expect it to just ignore them.

But if you think it's important to test all modifiers at once (rather than simply 2+), I'm happy to create the alternate test that you suggested.

twalpole:
Yes - I would prefer to leave the all modifier test there since it passes fine with Chrome, and is there a reason it can't then be a 3 modifier test for FF (just not shift)?

elliterate:
> is there a reason it can't then be a 3 modifier test for FF (just not shift)?

It can be. By “2+,” I was just referring to the fact that the test name simply says “multiple modifiers.”

I’ve updated the branch, skipping the existing test for Marionette and adding a custom one for Marionette that tests three modifiers.

twalpole:
Thanks

PR #2066, opened 2018-07-10
fix: move caret even with clear backspace

cyrilchampier:


twalpole:
Hmmm -- It's a shame there's no cross browser/cross platform keystroke we can use for this instead of setSelectionRange

lib/capybara/spec/session/node_spec.rb
```diff
@@ -100,9 +100,15 @@
     end
 
     it "should fill the field even if the caret was not at the end", requires: [:js] do
+      @session.first('//input').set('monkey')
       @session.execute_script("var el = document.getElementById('test_field'); el.focus(); el.setSelectionRange(0, 0);")
       @session.first('//input').set('')
       expect(@session.first('//input').value).to eq('')
+
+      @session.first('//input').set('monkey')
```
twalpole: This test can't go here.  The `clear: :backspace` behavior is Selenium specific so the test needs to go in `spec/shared_selenium_session.rb` with the other tests for that behavior

cyrilchampier: okay, I'll move the test tomorrow. Appart from that, are you ok with the PR?

twalpole: Looking at it more -- it appears the issue is Chrome specific -- wondering whether it's worth fixing it only in chrome_node.rb - I am curious why it only affects Chrome though since it seems like setting the caret to the beginning should affect FF too.

twalpole:
Changing set_text to do `native.send_keys(*([:end] + backspaces + [value.to_s]))` fixes this (at least on MacOS) without having to resort to execute_script -- if it works on other platforms too it would be my preferred solution

twalpole:
If using `:end` doesn't work on other platforms it should be possible to check `driver.browser.capabilities[:platform]` and prepend the correct key for other platforms.

cyrilchampier:
`:end` seems a much better way to move caret indeed!
I only have a mac to test, if travis is green, can we consider it ok ?

twalpole:
@cyrilchampier I don't fully understand why sending :end actually works, but if it works on mac and travis I'm fine considering it ok.  Please remove the extraneous `:focus_` and squash the commits down to 1

twalpole:
@cyrilchampier Do you have time to make the requested cleanup here so I can get this merged?  If not I'll go ahead and make the changes tomorrow.

cyrilchampier:
sorry, i am in holidays until wednesday.
i will not be able to check that until then.

cyrilchampier:
tx, I was getting married this week end, a bit busy ;)

twalpole:
Congratulations!

PR #2069, opened 2018-07-14
Remove 'at the moment' w.r.t. Selenium RC from the README

searls:
I think at this point in 2018, it's overwhelmingly unlikely Selenium 1.x support will be backported


twalpole:
Valid point :)   We also don't support Selenium 2 anymore so I need to update that too -  Thanks

PR #2070, opened 2018-07-17
Ignore leading whitespace in JavaScript snippets

elliterate:


twalpole:
Could you make the same change for Capybara::Node::Element#evaluate_script too?

elliterate:
>Could you make the same change for Capybara::Node::Element#evaluate_script too?

Done!

twalpole:
Thanks!

PR #2071, opened 2018-07-19
Add test for clicking label of a blocked element

elliterate:
An element positioned outside of the viewable document is considered hidden, effectively making that test a duplicate of [the "should check via the label if input is hidden" test][1]. The test was originally added as part of 272a052, but if one removes the implementation change, the test still passes in Chrome and Firefox. If we cover an otherwise visible element, we can address the scenario of a visible but not-interactable element, more thoroughly validating the need for the original implementation change.

[1]: https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/session/check_spec.rb#L177

### DIFF
--- a/lib/capybara/spec/session/check_spec.rb
+++ b/lib/capybara/spec/session/check_spec.rb
@@ -195,6 +195,13 @@ Capybara::SpecHelper.spec '#check' do
           @session.click_button('awesome')
           expect(extract_results(@session)['cars']).to include('pagani')
         end
+
+        it 'should check via the label if input is visible but blocked by another element' do
+          expect(@session.find(:checkbox, 'form_cars_bugatti', unchecked: true, visible: :all)).to be_truthy
+          @session.check('form_cars_bugatti', allow_label_click: true)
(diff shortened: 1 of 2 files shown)


lib/capybara/spec/session/check_spec.rb
```diff
@@ -189,7 +189,7 @@
           expect(end_time - start_time).to be < 10
         end
 
-        it 'should check via the label if input is moved off the left edge of the page' do
+        it 'should check via the label if input is visible but blocked by another element' do
```
twalpole: This test is actually relevant for non-selenium drivers - please don't change - add a new test if needed/wanted

elliterate: Can do. Do you have a preference for the next supercar brand to add? Bugatti, perhaps?

twalpole: Bugatti is definitely acceptable :)

twalpole:
Thanks!

PR #2075, opened 2018-07-23
WIP: Add Capybara.predicates_wait setting - defaults to true

twalpole:
All the finders wait by default for matching elements to exist, as do the predicate methods.  There's a reasonable argument that can be made that in some use cases waiting is not generally useful for the predicates.  This adds a setting that allows to default the predicate methods to non-waiting -  Still needs tests,  and a final decision on whether or not this is actually a good idea.


luke-hill:
This is something I've been debating when writing (or beginning), to write up the initial work for site_prism 3. As it's a good question to have.

I think long term it should probably be configurable

PR #2077, opened 2018-07-24
Fix required Ruby version [ci skip]

y-yagi:
Since baabfa5d675a1765d572421427eb41d36472c054, Capybara requires Ruby 2.3.0 or later.


twalpole:
Whoops, could have sworn I'd updated that -- thanks

PR #2083, opened 2018-08-10
Fire HTML5 drag and drop events in selenium driver

twalpole:
This attempts to enable HTML5 native drag and drop when using the Selenium driver.  Events fired is not complete and the order may not be 100% correct.  Needs testing and feedback from users.


fukayatsu:
@twalpole This fixed our test for feature which using Sortable.js.
Thank you so much!

> https://github.com/RubaXa/Sortable/issues/563

PR #2085, opened 2018-08-17
animation disabler can accept CSS as well as boolean

michaelglass:
We use a similar disable animation feature in our test suite. I want to use the capybara version of the animation disabler but it's not quite versatile enough. This adds the ability to allowlist certain css classes.

If y'all are interested in accepting this, I'd be happy to actually fix the test here.

Thanks!

### DIFF
--- a/lib/capybara/server/animation_disabler.rb
+++ b/lib/capybara/server/animation_disabler.rb
@@ -3,8 +3,20 @@
 module Capybara
   class Server
     class AnimationDisabler
+      def self.selector_for(css_or_bool)
+        case css_or_bool
+        when String
+          css_or_bool
+        when true
+          '*'
+        else
+          raise CapybaraError, 'Capybara.disable_animation supports either a String (the css selector to disable) or a boolean'
+        end
+      end
+
       def initialize(app)
         @app = app
+        @disable_markup = DISABLE_MARKUP_TEMPLATE % AnimationDisabler.selector_for(Capybara.disable_animation)
       end
 
       def call(env)
@@ -20,18 +32,20 @@ module Capybara
 
     private
 
+      attr_reader :disable_markup
+
       def html_content?
         !!(@headers['Content-Type'] =~ /html/)
       end
 
       def insert_disable(html)
-        html.sub(%r{(</head>)}, DISABLE_MARKUP + '\\1')
+        html.sub(%r{(</head>)}, disable_markup + '\\1')
       end
 
-      DISABLE_MARKUP = <<~HTML
+      DISABLE_MARKUP_TEMPLATE = <<~HTML
         <script defer>(typeof jQuery !== 'undefined') && (jQuery.fx.off = true);</script>
         <style>
-          * {
+           %s {
              transition: none !important;
              animation-duration: 0s !important;
              animation-delay: 0s !important;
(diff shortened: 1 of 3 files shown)


twalpole:
Rather than adding the  Capybara.enable_animations_for configuration option - Please change `Capybara.disable_animation` to take either the existing bool or a string, where the string would be the CSS selector used ('*' if true) .  This would require the user to specify the `:not(...)` if that's what they want but that's probably better anyway due to the limitations `:not` has on what can be passed to it.

lib/capybara/server/animation_disabler.rb
```diff
@@ -28,10 +28,19 @@ def insert_disable(html)
...
+        else
+          ''
+        end
+      end
+
       DISABLE_MARKUP = <<~HTML
         <script defer>(typeof jQuery !== 'undefined') && (jQuery.fx.off = true);</script>
         <style>
-          * {
+          *#{enable_animations_for} {
```
twalpole: Forcing a leading `*` restricts the CSS selectors that can be passed for no real benefit -- please remove and just set the selector if `*` if true is passed.

Gemfile
```diff
@@ -6,6 +6,7 @@ gem 'bundler', '~> 1.1'
 gemspec
 
 gem 'xpath', git: 'git://github.com/teamcapybara/xpath.git'
+gem 'pry'
```
twalpole: Please don't add unnecessary gems

lib/capybara/session.rb
```diff
@@ -85,7 +85,7 @@ def initialize(mode, app = nil)
       end
       @server = if config.run_server && @app && driver.needs_server?
         server_options = { port: config.server_port, host: config.server_host, reportable_errors: config.server_errors }
-        server_options[:extra_middleware] = [Capybara::Server::AnimationDisabler] if config.disable_animation
+        server_options[:extra_middleware] = [Capybara::Server::AnimationDisabler]
```
twalpole: Why are you always loading AnimationDisabler middleware?

spec/shared_selenium_session.rb
```diff
@@ -324,25 +324,34 @@
     end
 
     context 'AnimationDisabler' do
-      before(:context) do # rubocop:disable RSpec/BeforeAfterAll
```
twalpole: You need to create a new session with AnimationDisabler enabled

lib/capybara/session.rb
michaelglass: one approach: load the middleware conditionally, and have the animation setting be per-session.

another approach: load the middleware always, but have it do something (or not) based on the configuration value.

twalpole: This then breaks per session configuration ( "threadsafe" mode )

michaelglass:
rolled back the global session stuff

twalpole:
This is looking good now -- pease squash down to one commit and I'll merge it in -- Thanks!

twalpole:
Well it was looking good -- not sure what else you changed when squashing it, but looks like you changed the original tests causing them to fail, and others are failing too now.

michaelglass:
just changed the tests as I felt like they were inadequately testing the feature. Will figure it out this afternoon or tomorrow.

twalpole:
The original tests were correctly testing that the links were removed in well under the amount of time the animation was set to happen in, so those definitely should be restored to what they were originally -- the new tests are up to you

michaelglass:
@twalpole reverted as many changes as possible. Thanks for your extensive review. The end result is clearly much better.

twalpole:
Thanks!

PR #2087, opened 2018-08-24
Support set default value for text predicates and matchers normalize whitespace behaviour

stegalin:
Parameter to support old (2.18.0) normalize whitespace behaviour

### DIFF
--- a/lib/capybara/queries/text_query.rb
+++ b/lib/capybara/queries/text_query.rb
@@ -90,7 +90,8 @@ module Capybara
       end
 
       def text(node: @node, query_type: @type)
-        node.text(query_type, normalize_ws: options[:normalize_ws])
+        normalize_ws = options.fetch(:normalize_ws, session_options.default_normalize_ws)
+        node.text(query_type, normalize_ws: normalize_ws)
       end
 
       def default_type
--- a/lib/capybara/spec/session/assert_text_spec.rb
+++ b/lib/capybara/spec/session/assert_text_spec.rb
@@ -15,6 +15,20 @@ Capybara::SpecHelper.spec '#assert_text' do
     expect(@session.assert_text('text with whitespace', normalize_ws: true)).to eq(true)
   end
 
+  context 'with enabled default collapsing whitespace' do
+    before { Capybara.default_normalize_ws = true }
+
+    it 'should be true if the given unnormalized text is on the page' do
+      @session.visit('/with_html')
+      expect(@session.assert_text('text with   whitespace', normalize_ws: false)).to eq(true)
+    end
+
+    it 'should support collapsing whitespace' do
+      @session.visit('/with_html')
+      expect(@session.assert_text('text with whitespace')).to eq(true)
+    end
+  end
+
   it 'should take scopes into account' do
     @session.visit('/with_html')
     @session.within("//a[@title='awesome title']") do
--- a/lib/capybara/spec/session/has_text_spec.rb
+++ b/lib/capybara/spec/session/has_text_spec.rb
@@ -39,6 +39,20 @@ Capybara::SpecHelper.spec '#has_text?' do
     expect(@session).to have_text('text with whitespace', normalize_ws: true)
   end
 
+  context 'with enabled default collapsing whitespace' do
+    before { Capybara.default_normalize_ws = true }
+
+    it 'should search unnormalized text' do
+      @session.visit('/with_html')
+      expect(@session).to have_text('text with   whitespace', normalize_ws: false)
+    end
+
+    it 'should search whitespace collapsed text' do
+      @session.visit('/with_html')
+      expect(@session).to have_text('text with whitespace')
+    end
+  end
+
   it 'should be false if the given text is not on the page' do
     @session.visit('/with_html')
     expect(@session).not_to have_text('xxxxyzzz')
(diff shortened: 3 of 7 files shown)


twalpole:
In principle I'm ok with adding this setting, however it should only affect the predicates and matchers not the actual `#text` method.   - ie it needs to be implemented in `Capybara::Queries::TextQuery#text` rather than `Document#text`, `Element#text`, and `Simple#text`.  Also it needs to allow for the user to specify `false` if normalize_ws is defaulted to 'true' which the current code `normalize_ws ||= session_options.default_normalize_ws` would not.

lib/capybara/spec/session/assert_text_spec.rb
```diff
@@ -15,6 +15,19 @@
     expect(@session.assert_text('text with whitespace', normalize_ws: true)).to eq(true)
   end
 
+  context 'with enabled default collapsing whitespace' do
+    before do
+      @previous_default_normalize_ws = Capybara.default_normalize_ws
```
twalpole: Since `normalize_ws` is reset in spec_helper there's no need to store/restore the previous

lib/capybara/spec/session/has_text_spec.rb
```diff
@@ -39,6 +39,19 @@
     expect(@session).to have_text('text with whitespace', normalize_ws: true)
   end
 
+  context 'with enabled default collapsing whitespace' do
+    before do
+      @previous_default_normalize_ws = Capybara.default_normalize_ws
```
twalpole: Since normalize_ws is reset in spec_helper there's no need to store/restore the previous

stegalin:
@twalpole , hope, I have fixed your comments. Please, take a look

lib/capybara/node/element.rb
```diff
@@ -51,6 +51,7 @@ def native
       # method.
       #
       # @param type [:all, :visible]  Whether to return only visible or all text
+      # @param normalize_ws [Boolean] Whether to normalize whitespaces
```
twalpole: Please don't document this option - it's not to be encouraged

lib/capybara/queries/text_query.rb
```diff
@@ -90,7 +90,8 @@ def check_case_insensitive?
       end
 
       def text(node: @node, query_type: @type)
-        node.text(query_type, normalize_ws: options[:normalize_ws])
+        normalize_ws = options.key?(:normalize_ws) ? options[:normalize_ws] : session_options.default_normalize_ws
```
twalpole: options.fetch(:normalize_ws, session_options.default_normalize_ws)

lib/capybara/spec/session/assert_text_spec.rb
```diff
@@ -15,6 +15,20 @@
     expect(@session.assert_text('text with whitespace', normalize_ws: true)).to eq(true)
   end
 
+  context 'with enabled default collapsing whitespace' do
+    before { Capybara.default_normalize_ws = true }
+
+    it 'should be true if the given normalized text is on the page' do
```
twalpole: unnormalized

lib/capybara/spec/session/has_text_spec.rb
```diff
@@ -39,6 +39,20 @@
     expect(@session).to have_text('text with whitespace', normalize_ws: true)
   end
 
+  context 'with enabled default collapsing whitespace' do
+    before { Capybara.default_normalize_ws = true }
+
+    it 'should search correctly normalized text' do
```
twalpole: unnormalized

stegalin:
@twalpole thank you. Please, take a look again

twalpole:
LGTM -- Thanks

nashbridges:
@twalpole is it OK to mention this setting in UPGRADING notes (v2->v3)?

twalpole:
@nashbridges As long as it's mentioned that it will probably go away for 4.0, and is really only meant as a way for users to more easily migrate to 3.x -- Users really should be testing for text that users are expected to see

jeffshantz:
"Users really should be testing for text that users are expected to see"

Do users see newlines in HTML?

twalpole:
@jeffshantz  Do users see the new lines created by block elements  - yes, do they see the new lines created by &lt;br> elements, - yes

jeffshantz:
@twalpole I'm not talking about <br> elements.  Let's say I had this in my view:
```
Your balance: <%= @account.balance %>
```

Suppose I reformat my view (contrived in this example, but you get the point) and add some whitespace for better readability:

```
Your balance:
<%= @account.balance %>
```

Notice that I did not add a <br>.  Nevertheless, my tests looking for "Your balance: $100" are now failing, because Capybara is seeing "Your balance:\n$100".

twalpole:
What does it show as in the browser and what driver are you using?

twalpole:
@jeffshantz  https://gist.github.com/twalpole/cd87c608a20aa1759d0ca2f69e72f356 is standalone code showing it correctly matching the example you've given to the text without the linefeed (because the linefeed in the HTML as written wouldn't be shown to the user) when running with the rack_test, selenium, and apparition drivers.  If it's not working correctly for you you're going to need to provide more info on exactly what you're doing (modifying the gist I linked to recreate your issue would be best)

PR #2091, opened 2018-08-31
Allow check/uncheck/choose/attach_file/fill_in to work on element called on when no locator specified

twalpole:
Experiment to allow calling `check`/`uncheck`/`choose` to operate on the element they're called on when no locator is specified.


ibrahima:
Thanks for considering this change! I didn't really consider the scoping issue when I requested that, but I think I just assumed that like you said, it would scope it to the current element when you don't pass in a locator.

Does this change the behavior if you were to call `page.check()` without any locator? Or does that already check the first checkbox that it finds?

twalpole:
`page.check()` should behave the same -- check a checkbox on the page as long as there's only one visible checkbox on the page.  If there were multiple checkboxes it would raise an ambiguity error.

ibrahima:
Ah I see, that makes sense.

PR #2094, opened 2018-09-06
Fix compatibility issue with puma version < 3.8.0

hieuk09:
`Rack::Handler::Puma.config` is only added in [puma version 3.8.0](https://github.com/puma/puma/releases/tag/v3.8.0). Therefore, the tests will crash if I use puma version less than 3.8.0.

This PR adds a check to see if `.config` is defined. If it isn't, `Rack::Handler::Puma` will use `.run` instead. I'm not sure how to add tests for check different Puma versions, any advices would be very appreciated.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -453,7 +453,12 @@ Capybara.register_server :puma do |app, port, host, **options|
   # Therefore construct and run the Server instance ourselves.
   # Rack::Handler::Puma.run(app, { Host: host, Port: port, Threads: "0:4", workers: 0, daemon: false }.merge(options))
 
-  conf = Rack::Handler::Puma.config(app, { Host: host, Port: port, Threads: '0:4', workers: 0, daemon: false }.merge(options))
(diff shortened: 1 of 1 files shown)


twalpole:
Thanks for the PR but the problem with calling `Rack::Handler::Puma.run` is that it installs exit signal handlers and prevents test frameworks from quitting running tests when Ctrl-C is pressed. Puma 3.8.0 is over a year old, so I would suggest you upgrade the version you're running to something modern.  If you do need to stick with an older version for some reason then just register your own server block with `Capybara.register_server` and then set `Capybara.server` to use that registered server.

hieuk09:
@twalpole: thank you for your response. 

Because this costs me a few hours looking into the issue, should `Capybara` raise an error ("Capybara supports only puma version newer than 3.8.0") or document this behavior somewhere?

twalpole:
Capybara doesn’t actually depend on Puma since the user can register any server they want. However I would accept a PR that raises an error if the built-in puma server registration is started with < 3.8

PR #2095, opened 2018-09-06
Raise error if puma version is lower than 3.8.0

hieuk09:
As discuss in https://github.com/teamcapybara/capybara/pull/2094, since `capybara` doesn't support `puma` version lower than 3.8.0, raising an error in that case could help people identify this problem.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -448,6 +448,10 @@ Capybara.register_server :puma do |app, port, host, **options|
     require 'rack/handler/puma'
   rescue LoadError
     raise LoadError, 'Capybara is unable to load `puma` for its server, please add `puma` to your project or specify a different server via something like `Capybara.server = :webrick`.'
+  else
+    unless Rack::Handler::Puma.respond_to?(:config)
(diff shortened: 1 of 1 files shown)


lib/capybara.rb
```diff
@@ -453,7 +453,12 @@ module Selenium; end
   # Therefore construct and run the Server instance ourselves.
   # Rack::Handler::Puma.run(app, { Host: host, Port: port, Threads: "0:4", workers: 0, daemon: false }.merge(options))
 
-  conf = Rack::Handler::Puma.config(app, { Host: host, Port: port, Threads: '0:4', workers: 0, daemon: false }.merge(options))
+  conf = if Rack::Handler::Puma.respond_to?(:config)
+    Rack::Handler::Puma.config(app, { Host: host, Port: port, Threads: '0:4', workers: 0, daemon: false }.merge(options))
+  else
+    raise LoadError, 'Capybara requires `puma` version 3.8.0 or higher, please upgrade `puma` or register and specify your own server block'
```
twalpole: Could you please do this in an `else` added to the begin/rescue above so all puma viability checking is kept to the one section.

twalpole:
Thanks -- LGTM -- will merge once tests are finished.

Issue #2098, opened 2018-09-11
Drag and drop tests stop working in 3.6.0+


pbougie:
Capybara Version: 3.6.0
Driver Version: ChromeDriver 2.41.578706
Chrome Version: 69.0.3497.81

I have various drag and drop tests that work fine in version 3.5.1. They simply stop working when Capybara is upgraded to versions 3.6.0, 3.7.0, or 3.7.1.

What has changed? I can't seem to find much information about it.

twalpole:
Prior to 3.6.0 all that was generated for drag/drop was mousedown, mousemove, mouseup, in 3.6.0+ Capybara generates HTML5 native drag and drop events for Chrome - https://github.com/teamcapybara/capybara/commit/0538cb6688c2182995ba6beeac1384a78499263f if the element has the HTML5 `draggable` attribute set to "true".  It passes the tests in the Capybara suite for HTML5 drag and drop but it's possible it's missing something - what does your drag and drop use/require and what does the element you're dragging look like?

pbougie:
I'm dragging `<a>` elements into `<td>` elements using the Dragula JS library. I am not using the HTML5 `draggable` attribute.

Also just tried Firefox Nightly with the same results.

twalpole:
Below is a small ruby snippet that shows drag/drop working on the dragula example site, so if we can figure out what the difference between your usage and the dragula example is it would help narrow this down

```
require "capybara/dsl"
require 'selenium/webdriver'

sess = Capybara::Session.new(:selenium_chrome)
sess.visit('https://bevacqua.github.io/dragula/')
src = sess.first('#left-defaults > div')
dest = sess.first('#right-defaults > div')
src.drag_to dest
sleep 15 # keep session open so we can see it has dragged

```

pbougie:
Unfortunately, it's not publicly accessible. I'll try to put something together to reproduce the problem. Thanks.

twalpole:
After thinking about this for a while - I know why it's happening.  Images and links (the <a> elements you're dragging) default to having the HTML5 `draggable="true" `.  A workaround would be to add a `draggable="false"` attribute to all the elements you're dragging using dragula until we can fix this.  I guess I need to figure out how to tell whether another library is handling dragging -- hmmmm.

twalpole:
The `draggable="false"` workaround won't work due to a bug in Selenium treating `draggable` it as a boolean attribute rather than the enumerated attribute that it is

pbougie:
[...]

PR #2099, opened 2018-09-11
Only enable HTML5 dragging if default is not prevented on the mousedown event

twalpole:
This attempts to fix Issue #2098.  The issue is caused by the fact some elements (images and links) default to HML5 draggle = true.  Because of this we need to only generate HTML5 drag and drop events when the original mousedown event is not defaultPrevented.


twalpole:
Already merged

PR #2101, opened 2018-09-21
Fix date and time field string entry

botandrose:
Fixes a regression introduced in 744e990 where `input[type=date]`, `input[type=datetime]`, and `input[type=datetime-local]` can no longer be set by `String` value.

Below is a heavily-edited version of `capybara/selenium/node.rb` to illustrate the issue.
If `#set` is called with `String` value, `#set_text` is ultimately called with a `SettableValue` value, where `#to_s` is immediately called upon it, resulting in a String value of e.g. `"#<SettableValue:0x00000000008a0138>"`, which is then subsequently entered into the date field via `#send_keys`, resulting in a garbage value in the date field.

```ruby
  def set(value, **options)
    # ...
    case self[:type]
    when 'date'
      set_date(value)
    end
  end

  def set_date(value)
    value = SettableValue.new(value)
    return set_text(value) unless value.dateable?
    # ...
  end

  def set_text(value, clear: nil, **_unused)
    value = value.to_s
    # ...
    send_keys(value)
  end
```

Potential fixes:
1. Implement `#to_s` on SettableValue as an alias of `#value`, or something similar.
2. Replace the `value` variable mutation in `#set_date` and friends, and use the original value if we fall back to `#set_text`. This is the path I have gone with in this PR.

What do you think?

### DIFF
--- a/lib/capybara/selenium/node.rb
+++ b/lib/capybara/selenium/node.rb
@@ -355,6 +355,10 @@ private
       @value = value
     end
 
+    def to_s
+      value.to_s
+    end
+
     def dateable?
       !value.is_a?(String) && value.respond_to?(:to_date)
     end
--- a/spec/selenium_spec_chrome.rb
+++ b/spec/selenium_spec_chrome.rb
@@ -84,4 +84,32 @@ RSpec.describe 'Capybara::Session with chrome' do
       end
     end
   end
+
+  describe 'filling in Chrome-specific date and time fields with keystrokes' do
+    let(:datetime) { Time.new(1983, 6, 19, 6, 30) }
+
+    before do
+      @session = TestSessions::Chrome
+      @session.visit('/form')
+    end
+
+    it 'should fill in a date input with a String' do
+      @session.fill_in('form_date', with: "06/19/1983")
+      @session.click_button('awesome')
+      expect(Date.parse(extract_results(@session)['date'])).to eq datetime.to_date
+    end
+
+    it 'should fill in a time input with a String' do
+      @session.fill_in('form_time', with: "06:30A")
+      @session.click_button('awesome')
+      results = extract_results(@session)['time']
+      expect(Time.parse(results).strftime('%r')).to eq datetime.strftime('%r')
+    end
+
+    it 'should fill in a datetime input with a String' do
+      @session.fill_in('form_datetime', with: "06/19/1983\t06:30A")
+      @session.click_button('awesome')
+      expect(Time.parse(extract_results(@session)['datetime'])).to eq datetime
+    end
+  end
 end


twalpole:
Thanks for catching this -- I thought we had tests for setting by String -- guess not.  I'd prefer this to be implemented via adding #to_s to SettableValue that just returns `value.to_s` please.  If you make that change and squash your commits I'll merge it in.

twalpole:
and the tests should probably be in lib/capybara/spec/session/fill_in_spec.rb rather than selenium specific

botandrose:
@twalpole Ah, just saw your strikeout after I forced-pushed a new commit.

botandrose:
@twalpole Yeah, was wondering if the browser-specific fields would have an impact. Was hoping that you guys had somehow already solved that in the test plumbing.

twalpole:
@botandrose Nope, the browser agnostic version is to pass a Date/Time object - If the keystrokes that need to be sent for FF/Chrome are different then the tests will need to go into the browser specific test files  `selenium_spec_chrome.rb` and `selenium_spec_marionette.rb`.  We had tried to allow Capybara to have a lookup for the correct format but we couldn't come up with a way to accurately detect what locale the browser was actually in.

twalpole:
FYI - the tests run in US locale so the strings necessary to be sent for Chrome would be `06/19/1983`,  `06:30A`, and `06/19/1983\t06:30A`  -- working out FF next

twalpole:
FF seems broken as far as sending keystrokes to date fields - so not sure if that's new with the latest geckodriver or just been broken for a while

botandrose:
@twalpole Gotcha. Well, I'll see if I can't get this green for Chrome, at least, and then we can go from there. Thank you for the help, and for putting in the legwork for figuring out the correct strings to assert!

twalpole:
FF also doesn't support `datetime-local` inputs so the \t in those keys is screwing that up

twalpole:
@botandrose You need to visit the right page for those tests, also the month and day order are still wrong on the datetime test.

botandrose:
@twalpole Yeah, sorry, I'm having a hard time getting these tests to run on my local machine, so I'm kinda stabbing in the dark, and addressing the failures as I see them. :/

twalpole:
`rake spec_chrome` should run the chrome tests locally

botandrose:
@twalpole Okay, we're green, and these new tests -- while only on Chrome -- will prevent this regression from occurring in the future. What do you think? Good enough to merge?

twalpole:
Thank you

botandrose:
@twalpole < 1 hour from opening to feedback, and < 12 hours from opening to merge! Thanks for providing a great first contributor experience to Capybara!

twalpole:
@botandrose You're welcome.  Just released in 3.8.1

PR #2102, opened 2018-09-21
Optimize attribute checking for :element selector

twalpole:
This parses attribute value regexes to do substring matching in the XPath in order to limit the number of returned elements that need to be checked - ~~stolen~~ borrowed from Watir


twalpole:
This was merged as 907b9e269f3057778a2587af455dbccd1114979e

Issue #2106, opened 2018-10-02
wait_for_pending_requests doesn't work if multiple drivers are used


abicky:
`Capybara::Server#wait_for_pending_requests` uses `Capybara::Server::Middleware#pending_requests?`, but capybara uses only one middleware if `Capybara.reuse_server` is true (its default value). As a result, `Capybara::Server#wait_for_pending_requests` doesn't waits pending requests when we use the driver whose middleware is not used.

This behavior makes tests unstable if they are executed randomly.

## Meta
Capybara Version: 3.8.2

Driver Information (and browser if relevant): selenium-webdriver 3.14.0 with Google Chrome 69.0.3497.100 (chromedriver 2.42)

## Expected Behavior

`wait_for_pending_requests` waits for pending requests even if the server which belongs to another driver is reused.

## Actual Behavior

`wait_for_pending_requests` doesn't wait for pending requests.

## Steps to reproduce

An reproducible code is [here](https://gist.github.com/abicky/99b2e0783feefce3be8a8a5097cf6d20#file-test-rb) and you can execute the code like below:

```
git clone https://gist.github.com/99b2e0783feefce3be8a8a5097cf6d20.git capybara-demo
docker build . -t capybara-demo
docker run --rm capybara-demo
```

An output is here:

```
Loaded suite test
Started
Google Chrome 69.0.3497.100
========= driver1 =========
Capybara starting Puma...
* Version 3.12.0 , codename: Llamas in Pajamas
* Min threads: 0, max threads: 4
* Listening on tcp://127.0.0.1:33445
Seeping...
end of the request
Seeping...
wait_for_pending_requests: (pending_requests?: true)
end of the request
end of wait_for_pending_requests
========= driver2 =========
Seeping...
end of the request
Seeping...
wait_for_pending_requests: (pending_requests?: false)
end of wait_for_pending_requests
end of the request
.
Finished in 3.2033909 seconds.
-------------------------------------------------------------------------------
1 tests, 0 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------
0.31 tests/s, 0.00 assertions/s
```

twalpole:
This is as designed - If you're instantiating multiple sessions and only using one in a test you need to be using the first session created or setting `Capybara.reuse_server = false`

abicky:
I've created a PR https://github.com/teamcapybara/capybara/pull/2107.
[...]

PR #2107, opened 2018-10-03
Reuse also middlewares

abicky:
This PR resolves https://github.com/teamcapybara/capybara/issues/2106.

### DIFF
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -13,6 +13,10 @@ module Capybara
       def ports
         @ports ||= {}
       end
+
+      def middlewares
+        @middlewares ||= {}
+      end
     end
 
     attr_reader :app, :port, :host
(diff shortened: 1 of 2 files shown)


twalpole:
Unfortunately this has been tried previously and leads to deadlocks during session reset when one session is waiting but a later session hasn't yet been reset.  Implementing something like this requires reworking reset so that all open sessions get reset before waiting for all sessions pending_requests to end

twalpole:
The real solution to this would be tying all requests in the middleware to a specific session (cookie, query parameter, etc) but I haven't spent the time to work out whether or not that's actually possible with multiple subdomains, etc

abicky:
Umm, how about a change like below? I think `Capybara::Server` shouldn't depend on `Capybara.current_session`, but this is easy solution.

```diff
diff --git a/lib/capybara/server.rb b/lib/capybara/server.rb
index ee57f5c5..8804b155 100644
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -55,6 +55,10 @@ module Capybara
     end

     def wait_for_pending_requests
+      if Capybara.current_session.server != self
+        raise "#{__method__} is called but the current session server is another. Please set Capybara.reuse_server to false."
+      end
+
       timer = Capybara::Helpers.timer(expire_in: 60)
       while pending_requests?
         raise 'Requests did not finish in 60 seconds' if timer.expired?
```

If it looks good, I'll fix tests.

twalpole:
That will cause `reset_sessions!` to raise whenever there is more than one session being used

abicky:
I close this PR cf. https://github.com/teamcapybara/capybara/issues/2106

PR #2108, opened 2018-10-04
Fix ruby warning

y-yagi:
This fixes following warning:

```
capybara/lib/capybara.rb:396: warning: instance variable @specified_session not initialized
```


twalpole:
Thanks

PR #2123, opened 2018-10-30
added methods to get screen resolution of any devices

gowthamraj198:
- get_screen_resolution_height to get screen height
- get_screen_resolution_width to get screen width


twalpole:
What is the use case for this?  The Window object already has a `size` method (`current_window.size`), and you really shouldn't need to be checking window size in a test anyway - since you set it to the size you want to run the tests in.

PR #2124, opened 2018-11-01
Fix doc on result

colorbox:
Hi ✋ 

I found wrong point on capybara/result doc.
This PR fix it

## before
screen shot 2018-10-31 at 14 41 08

## after

screen shot 2018-10-31 at 14 41 29


twalpole:
thanks

PR #2130, opened 2018-11-14
Support a retrying before block for assert_selector/assert_no_selector

twalpole:
This is a potential addition to provide an easier solution to an issue that arose with chromedriver/chrome and hover (Firefox works fine).  The issue was that chromedriver doesn't appear to process mouse location as elements on the page are removed meaning CSS :hover doesn't appear to get processed if the call to `hover` occurs during animation.  This PR would add a `:before` option to assert_selector/assert_no_selector and all methods based on them.  The option takes a block which is executed before every time the assertion/expectation is checked
 
    el = find('#an_element')
    expect(page).to have_selector(:css, '.popup', before: { el.hover })

That would repeatedly call `el.hover` each time `have_selector` retries finding the popup element.  

I'm not fully sold on this approach yet since it really is only a way to work around browser deficiencies.  If I anyone can come up with non browser bug instances where this would be useful that would be helpful.

### DIFF
--- a/lib/capybara/node/matchers.rb
+++ b/lib/capybara/node/matchers.rb
@@ -723,6 +723,10 @@ module Capybara
         args.first.is_a?(Symbol) ? args.shift : session_options.default_selector
       end
 
+      def extract_before_block(args)
+        args.last.is_a?(Hash) ? args.last.delete(:before) : nil
+      end
+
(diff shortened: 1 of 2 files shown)


twalpole:
Closing because I can't justify adding this

PR #2133, opened 2018-11-19
Fix typo

sue445:
https://github.com/teamcapybara/capybara/blob/master/History.md#version-3111

# Before
image

# After
image


twalpole:
Thanks

PR #2135, opened 2018-11-20
Add Selenium Firefox Headless driver

andrewhavens:
Since Firefox is the default browser when using Selenium, and it's easy to get started using, it would be nice to include a headless configuration for Firefox. So that's what I've done in this PR. Now we can simply set our driver to `:selenium_firefox_headless` to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,7 @@ and add it to your Gemfile if you're using bundler.
 Capybara pre-registers a number of named drives that use Selenium - they are:
 
   * :selenium                 => Selenium driving Firefox
+  * :selenium_headless        => Selenium driving Firefox in a headless configuration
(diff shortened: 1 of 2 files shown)


twalpole:
Since the already existing Firefox driver is just  `:selenium` please just name the new one `:selenium_headless` rather than `:selenium_firefox_headless` - Also you can't use the `headless!` method since it wasn't added until seleniuv-webdriver 3.7.0 and we technically support > 3.5.0, so that would need to manually add the `--headless` arg.

andrewhavens:
@twalpole Thanks for the review! I've amended my PR based on your feedback. Note that the Firefox headless argument is "single dash" (i.e. `-headless`).

twalpole:
@andrewhavens Thanks for the changes - I'll get this merged in later today

PR #2138, opened 2018-11-29
Improve compatibility of html5 drag and drop simulation with client libraries.

erkki:
This change in specific makes it possible to test ember apps that are using https://github.com/mharris717/ember-drag-drop for dnd handling. Plausibly other client libraries using similar approaches would benefit as well.

### DIFF
--- a/lib/capybara/selenium/extensions/html5_drag.rb
+++ b/lib/capybara/selenium/extensions/html5_drag.rb
@@ -29,8 +29,11 @@ class Capybara::Selenium::Node
     HTML5_DRAG_DROP_SCRIPT = <<~JS
       var source = arguments[0];
       var target = arguments[1];
+      var sourceRect = source.getBoundingClientRect();
+      var targetRect = target.getBoundingClientRect();
 
       var dt = new DataTransfer();
+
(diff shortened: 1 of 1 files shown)


twalpole:
Could you add a test for this behavior?

erkki:
> Could you add a test for this behavior?

I'll take a look, haven't familiarised myself with the test suite just yet, wanted to get this out for comments first.

twalpole:
I'm fine with this change as long as it comes with some sort of test (probably in `spec/shared_selenium_session.rb` with the other HTML5 drag/drop tests)

twalpole:
Actually thinking about this more -- the target would not (generally) get a dragover event with the source location as clientX/Y -  so if the intention is to provide direction via the dragover events the first dragover event should probably calculate clientX/Y as the first position inside the target on the line from the source center to target center.

twalpole:
@erkki Please try the `dragover_more` branch and see if that works for you.

twalpole:
Closing due to lack of response.  We can reopen if there is future interest in this

erkki:
Sorry for the delayed response, just tested this on our test suite and it looks good!

PR #2141, opened 2018-12-07
HTML5 emulation continuation

twalpole:
This builds on PR #2138 but changes the first dragover event clientX/Y to be the entry point to the target rect on the line joining the source center and target center.  It also adds a test to ensure clientX/Y locations are set.


twalpole:
Closing due to lack of response.  Will reopen if there is interest

PR #2144, opened 2018-12-19
unfreeze nokogiri

ahorek:


rigoleto:
What's the status on this? Capybara is falling behind the latest Nokogiri updates, which include important security and performance updates.

twalpole:
@rigoleto - Capybara specifies ~> 1.8 as its version requirement which means >= 1.8.0 and < 2.0.0. Since the latest version of nokogiri is 1.10.3 that’s fully compatible with the current Capybara restrictions. If your project isn’t updating nokogiri it’s not because of Capybara

ahorek:
Yes, that why I closed the issue.

PR #2150, opened 2019-01-17
Deal with race condition when reseting between tests

perryn:
Hello,

A while ago we were experiencing annoying intermittent test failures with "Timed out waiting for Selenium session reset"

After painstaking investigation I determined that our (Ember) application was reacting to the session being cleared and navigating the browser to our login page. 

This created a race condition where sometimes this happened *after* capybara had issued @browser.navigate.to('about:blank') but before the browser had actually gone there.

Thus the loop waiting for the page to be empty timed out because the browser was not at about:blank.

I monkey patched Capybara in our code base to add this check that we are actually at about:blank inside the loop, and navigate there again if we are not.

Since I did this, we have not encountered the problem again for 4 months.

It seems to me that there is a general problem here where JS applications under test could trigger a race condition by triggering a navigation at the wrong time, so I submit this change for consideration.

I could not see any specs that covered this area/behaviour so this is a change to production code only, but please let me know if there are specs somewhere I should have updated

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -386,6 +386,11 @@ private
       raise Capybara::ExpectationNotMet, 'Timed out waiting for Selenium session reset' if timer.expired?
 
       sleep 0.01
+
+      # It has been observed that it is possible that asynchronous JS code in
+      # the application under test can navigate the browser away from about:blank
+      # if the timing is just right. Ensure we are still at about:blank... 
+      @browser.navigate.to('about:blank') if current_url != 'about:blank'
     end
   end
 
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -390,7 +390,7 @@ private
       # It has been observed that it is possible that asynchronous JS code in
       # the application under test can navigate the browser away from about:blank
       # if the timing is just right. Ensure we are still at about:blank... 
-      @browser.navigate.to('about:blank') if current_url != 'about:blank'
+      @browser.navigate.to('about:blank') unless current_url == 'about:blank'
     end
   end
 


twalpole:
Hmmm -- I assume you're using Chrome?? This would probably have gone away as soon as Chrome releases its W3C WebDriver compatible version of chromedriver.  Anyway I don't have anything against this really although I would prefer it if you were able to produce a test that replicates the behavior.  The test should go somewhere around https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/session/reset_session_spec.rb#L39

perryn:
Hi

Yes we are using chrome.

Its late and I have the dumb: I can't seem to run that spec locally

```
bundle exec rspec -r ./spec/spec_helper.rb spec/session/reset_session_spec.rb 
No examples found.


Finished in 0.00033 seconds (files took 0.44356 seconds to load)
0 examples, 0 failures
```
what am I missing?

twalpole:
Sorry, Capybaras specs are a little different than most due to the need to have some of the specs inside the gem so other drivers can use them for their own tests.  The easiest way to run a specific set of tests from those is to tag them with `:focus_` metadata - so for the reset session specs change  https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/session/reset_session_spec.rb#L3 to `Capybara::SpecHelper.spec '#reset_session!', :focus_ do` and then run `bundle exec rake spec_chrome`

twalpole:
@perryn Were you able to create a test case for this?

connorshea:
I think I'm experiencing the same thing in my project https://travis-ci.com/connorshea/ContinueFromCheckpoint/builds/98707862

I'm not sure how to make it reproducible, though.

perryn:
Not yet sorry, I plan to get back to it soon

twalpole:
Any updates on this?  I would like to do a Capybara release in the next day or so, and would include this if we have a test for it

perryn:
Hi @twalpole,

Unfortunately I have been unable to devise a test for this that would reliably fail in the absence of my patch. 

The basic problem is that I am trying to guard against a race condition and so it is of course highly dependent on exact timings. I quickly end up trying to tune values passed into sleep and setTimeout to try and make the race condition more likely, and those values will be highly context/machine dependant.

I also attempted to re-create the situation that the race condition creates by intercepting the first navigation to about:blank, but I was thwarted here by the other guards that capybara already has to prevent deal with such shenanigans.

I considered writing a new type of test to test the loop in isolation (with mocked out values for window.location) but that, it seems to me, would be one of those cases where the test simply restates the implementation code and would be of little value.

At this point, I can only re-state that we had a test suite that would fail annoying often without this patch.

Thoughts?

twalpole:
Ok, if you could swap it from an `if !=` to an `unless ==` then I'll go ahead and merge it

twalpole:
Nevermind - I just did it in githubs editor -- will merge

connorshea:
Apologies for not testing this sooner.

So I wrote a _really_ basic little script to test this with my feature tests (there are only 40 of them, so it's not a massive test suite or anything):

```ruby
20.times do |i|
  puts "RUN ##{i}"
  system('bundle exec rspec spec/features')
  puts
end
```

I ran the script with the current stable release of capybara and had 3/20 of the runs fail with "Timed out waiting for Selenium session reset".

Then I switched to the current git master for capybara and re-ran the test suite, none of them failed. Then I re-ran it again to make sure, so it now failed 0/40 times vs. 3/20 for the version before this PR was merged.

PR #2161, opened 2019-02-24
Raise all server errors by default, including Exception

### LINKED ISSUE #2160: Capybara hides server errors by default
Is there any reason why `Capybara.config.server_errors` default is still `[StandardError]` instead of `[Exception]`?

I have been bitten by this multiple times, some errors (e.g. `WebMock::NetConnectNotAllowedError < Exception`) are just invisible using the defaults and makes testing/debugging harder.

I found [this comment](https://github.com/teamcapybara/capybara/issues/1245#issuecomment-76280570):

> Would PR #1488 solve your issue(s) in the midterm - and then we can consider raising all exceptions by default in Capybara v3

Can we change the defaults now? I can open up a PR if you are interested.

ebeigarts:
Fixes #2160


twalpole:
Thanks

PR #2169, opened 2019-03-19
Update README.md

natetech333:
Remove slang, 3 letter acronyms and IRC/Chat abbreviations


twalpole:
Not really convinced of the need for this but it doesn't hurt, so why not.

natetech333:
Because I thought AFAIK was some kind of new Java Technology similar to AKKA. In my opinion, slang does not belong in an engineering document. Thank you for accepting.

PR #2179, opened 2019-04-23
Support multiple expression types in on selector

twalpole:
This is the part of PR #2163 that enables the multiple expression types on a selector


twalpole:
This was merged via dcd7fab897cc9fd0b794715dcd0cf690d8134980

PR #2183, opened 2019-05-02
Update README.md

artur-beljajev:
There is a `teardown` callback available in Rails and using it eliminates any chance that `super` will not be called in overridden `setup` method.


twalpole:
Thanks!

PR #2184, opened 2019-05-10
Description was wrong

sawadatsuyoshi-mf:
I see no such method as `excatly`. Rather, I found `exactly`.


twalpole:
Thanks for catching this typo, at least I only misspelled it in the change log :)

PR #2188, opened 2019-05-15
Simulate file/external data drop in chrome and firefox

twalpole:
I tried to simulate this for Safari as well, but there are two issues. It doesn't provide a `DataTransfer` constructor, however even if I worked around that by triggering a cut event to get access to a system built `DataTransfer` instance that object won't allow me to add files to it (just silently ignores).


twalpole:
Merged in 94f7f8881a4a50f7fbb24ee0a9eab2e5df218d3c

PR #2190, opened 2019-05-17
Improve API documentation

ybiquitous:
This change does followings:

* Change just reference texts to links (such as class or method) as possible.
* Add missing API documents such as `Capybara.test_id`.
* Fix invalid links.
* Fix broken syntax such as a missing argument name, invalid indentations, and so on.

See also the [v3.20.0 latest documentation](https://www.rubydoc.info/gems/capybara/3.20.0).


lib/capybara.rb
```diff
@@ -45,13 +45,55 @@ class << self
...
+    #   See {Capybara.configure}
+    # @!method disable_animation
+    #   See {Capybara.configure}
+    # @!method test_id
+    #   See {Capybara.configure}
+    # @!method predicates_wait
+    #   See {Capybara.configure}
+    # @!method default_normalize_ws
     #   See {Capybara.configure}
     SessionConfig::OPTIONS.each do |method|
```
ybiquitous: Same order as defined in `SessionConfig`:

https://github.com/teamcapybara/capybara/blob/31e3c56b94cc2eef8778bad18f58ed2e6bf51648/lib/capybara/session/config.rb#L15-L63

twalpole:
@ybiquitous Thanks for this, however the reason most of the `Capybara.xxxx` settings aren't documented above the `SessionConfig::OPTIONS.each` is because they're going to be deprecated in favor of specifying them through `Capybara.configure`.  If you want to modify this PR to change the all the added settings links to link to `Capybara.configure` I'll be happy to accept it.

ybiquitous:
@twalpole Thanks for the quick feedback! Sorry, I understand why some docs are missing. 👍 

I'm willing to rewrite docs on this PR, but I have some questions:

- Will currently described methods such as `Capybara.default_max_wait_time` be deprecated?
- If so, how should I rewrite links such as `{Capybara.default_max_wait_time}`? For example, `{Capybara::SessionConfig#default_max_wait_time}`?
- (Sadly, it seems YARD can't have links to options table in method docs...)

https://github.com/teamcapybara/capybara/blob/94f7f8881a4a50f7fbb24ee0a9eab2e5df218d3c/lib/capybara.rb#L28-L42

- As well, how should I rewrite links for currently undescribed methods such as `Capybara.test_id`?

https://github.com/teamcapybara/capybara/blob/94f7f8881a4a50f7fbb24ee0a9eab2e5df218d3c/lib/capybara/node/finders.rb#L110

twalpole:
@ybiquitous default_max_wait_time _probably_ won't get deprecated because it's used by so many projects/other gems/etc.  For others doing something like {Capybara.configure test_id} would probably work ok to have `test_id` show up in the docs linked to Capybara.configure.

ybiquitous:
@twalpole Thank you for the advice! `{Capybara.configure test_id}` works!

I'm going to fix docs, please wait. 🙏

ybiquitous:
The PR diff has increased than I expected. 😓 

If it would be desirable, I'm ready to split this big ball to small PRs.

lib/capybara/node/actions.rb
```diff
@@ -6,36 +6,32 @@ module Actions
       # @!macro waiting_behavior
       #   If the driver is capable of executing JavaScript, this method will wait for a set amount of time
       #   and continuously retry finding the element until either the element is found or the time
-      #   expires. The length of time +find+ will wait is controlled through {Capybara.default_max_wait_time}
+      #   expires. The length of time {Capybara::Node::Finders#find} will wait is controlled through {Capybara.configure default_max_wait_time}.
```
twalpole: This is a macro that gets applied to other methods - please just change `+find+` to `this method` - I think $0 is supposed to reference to the method the macro is applied to but after a quick (but limited test) it doesn't appear to work in this context - not really sure why

lib/capybara/node/actions.rb
```diff
@@ -6,36 +6,32 @@ module Actions
...
       #
-      #   @option options [false, true, Numeric] wait (Capybara.default_max_wait_time) Maximum time to wait for matching element to appear.
+      #   For each option details, see {Capybara::Node::Finders#find} options.
```
twalpole: I don't want this change - the `wait` parameter declaration should stay for each method this macro is included in

lib/capybara/node/actions.rb
```diff
@@ -6,36 +6,32 @@ module Actions
...
-      # {Capybara::Node::Actions#click_link} for what locator will match against for each type of element
+      # Finds a button or link and clicks it.  See {#click_button} and
+      # {#click_link} for what locator will match against for each type of element
       #
-      # @overload click_link_or_button([locator], **options)
-      #   @macro waiting_behavior
-      #   @param [String] locator      See {Capybara::Node::Actions#click_button} and {Capybara::Node::Actions#click_link}
-      #
-      # @return [Capybara::Node::Element]  The element clicked
+      # @see #click_button
```
twalpole: Please leave this as it was - including the waiting behavior macro

lib/capybara/node/actions.rb
```diff
@@ -6,36 +6,32 @@ module Actions
...
+      # Finds a link by id, {Capybara.configure test_id} attribute, text or title and clicks it. Also looks at image
       # alt text inside the link.
       #
-      # @overload click_link([locator], **options)
-      #   @macro waiting_behavior
-      #   @param [String] locator         text, id, Capybara.test_id attribute, title or nested image's alt attribute
-      #   @param options                  See {Capybara::Node::Finders#find_link}
-      #
+      # @param [String] locator         text, id, {Capybara.configure test_id} attribute, title or nested image's alt attribute
+      # @option (see Capybara::Node::Finders#find)
```
twalpole: Please leave the parameters as they were

lib/capybara/node/actions.rb
```diff
@@ -45,13 +41,11 @@ def click_link(locator = nil, **options)
...
-      # \<button> element. All buttons can be found by their id, name, Capybara.test_id attribute, value, or title. \<button> elements can also be found
+      # \<button> element. All buttons can be found by their id, name, {Capybara.configure test_id} attribute, value, or title. \<button> elements can also be found
       # by their text content, and image \<input> elements by their alt attribute
       #
-      # @overload click_button([locator], **options)
-      #   @macro waiting_behavior
-      #   @param [String] locator      Which button to find
-      #   @param options     See {Capybara::Node::Finders#find_button}
+      # @param [String] locator      Which button to find
+      # @option (see Capybara::Node::Finders#find)
```
twalpole: Please leave as this was - including the waiting behavior, and linking to find_button rather than find

lib/capybara/node/actions.rb
```diff
@@ -70,28 +64,25 @@ def click_button(locator = nil, **options)
...
-      #   @option options [String] placeholder    Match fields that match the placeholder attribute
-      #   @option options [String, Array<String>, Regexp] class    Match fields that match the class(es) provided
-      #   @option options [Hash] fill_options     Driver specific options regarding how to fill fields (Defaults come from Capybara.default_set_options)
-      #
-      # @return [Capybara::Node::Element]  The element filled_in
-      def fill_in(locator = nil, with:, currently_with: nil, fill_options: {}, **find_options)
-        find_options[:with] = currently_with if currently_with
-        find_options[:allow_self] = true if locator.nil?
-        find(:fillable_field, locator, find_options).set(with, fill_options)
+      # @param [String] locator                    Which field to fill in
```
twalpole: Why the change?

lib/capybara/node/actions.rb
```diff
@@ -136,10 +127,10 @@ def choose(locator = nil, **options)
       # @overload check([locator], **options)
       #   @param [String] locator           Which check box to check
       #
-      #   @option options [String] option  Value of the checkbox to select
-      #   @option options [String, Regexp] id       Match fields that match the id attribute
-      #   @option options [String] name     Match fields that match the name attribute
-      #   @option options [String, Array<String>, Regexp] class    Match fields that match the class(es) provided
+      #   @option options [String] :option  Value of the checkbox to select
```
twalpole: Not really sure the change to symbols is necessary -- probably should be a separate PR so we can compare the changes it makes to the docs more easily

lib/capybara/node/finders.rb
```diff
@@ -5,43 +5,45 @@ module Node
...
       #
+      #     page.find('#foo').find('.bar')
+      #     page.find(:xpath, './/div[contains(., "bar")]')
+      #     page.find('li', text: 'Quox').click_link('Delete')
+      #
       # @!macro waiting_behavior
       #   If the driver is capable of executing JavaScript, this method will wait for a set amount of time
       #   and continuously retry finding the element until either the element is found or the time
-      #   expires. The length of time +find+ will wait is controlled through {Capybara.default_max_wait_time}
+      #   expires. The length of time {#find} will wait is controlled through {Capybara.configure default_max_wait_time}
```
twalpole: Same as previous comment on macro

lib/capybara/node/finders.rb
```diff
@@ -5,43 +5,45 @@ module Node
...
       #   and defaults to 2 seconds.
-      #   @option options [false, true, Numeric] wait (Capybara.default_max_wait_time) Maximum time to wait for matching element to appear.
```
twalpole: Don't remove the `wait` option from the macro

lib/capybara/node/finders.rb
```diff
@@ -53,23 +55,21 @@ def find(*args, **options, &optional_filter_block)
...
-      # @param (see Capybara::Node::Finders#find)
-      #
       # @macro waiting_behavior
       #
-      # @option options [Symbol] match        The matching strategy to use.
-      #
-      # @return [Capybara::Node::Element]      The found element
-      # @raise  [Capybara::ElementNotFound]    If the element can't be found before time expires
+      # @overload ancestor([kind], locator, **options)
+      # @param (see #find)
```
twalpole: Does this actually link?? Also I don't really see the need for removing the return and raise from this methods declaration.

lib/capybara/node/finders.rb
```diff
@@ -78,57 +78,59 @@ def ancestor(*args, **options, &optional_filter_block)
...
-      #
       # @macro waiting_behavior
       #
-      # @option options [Symbol] match        The matching strategy to use.
-      #
-      # @return [Capybara::Node::Element]      The found element
-      # @raise  [Capybara::ElementNotFound]    If the element can't be found before time expires
+      # @overload sibling([kind], locator, **options)
+      # @param (see #find)
+      # @return (see #find)
```
twalpole: why?

lib/capybara/node/actions.rb
ybiquitous: `The length of time {$0} will wait` also works well, but I also think just `this method` is good here. I'll fix it. 👍 

image

lib/capybara/node/actions.rb
ybiquitous: Because this macro is almost same as the macro defined in `Finders` class, I want to be DRY. For it, I think that just adding the link to `Finders#find` options is good.
Generally, to maintain duplicate documents is hard. But this change is out of this PR's original purpose… 😓 

https://github.com/teamcapybara/capybara/blob/bcbefe7acceef14ca7b90c4aacfa0b9ce8520e20/lib/capybara/node/finders.rb#L11-L16

lib/capybara/node/actions.rb
ybiquitous: Because I think that to use the original method signature is more simple and DRY than using `@overload`.
And I think that to use `@param` for keyword arguments is better than `@options` according to the [YARD advice](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#option) below.

> Note: For keyword parameters, use `@param`, not `@option`.

But this is also too much for this PR. 😓

lib/capybara/node/finders.rb
ybiquitous: I want to be DRY also here, but this is too much. 💦

lib/capybara/node/actions.rb
ybiquitous: Yes, I think so 👍

twalpole:
It may be possible to share the `waiting_behavior` macro between the two files if you can ensure the file that defines it is processed before the other one -- this may be possible with a .yardopts file - not sure if RubyDoc.info will use that though.

ybiquitous:
> It may be possible to share the `waiting_behavior` macro between the two files if you can ensure the file that defines it is processed before the other one -- this may be possible with a .yardopts file - not sure if RubyDoc.info will use that though.

Umm..., Isn't it easier to use ["Reference Tags"](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#Reference_Tags) than macros?
For example:

```rb
# @option (see Capybara::Node::Finders#find)
```

twalpole:
But that makes the docs harder to read/use. The commonly used options (like `wait`)  should be shown with each method not requiring going to another page to see

ybiquitous:
Surely you're right 👍 

It seems that a `.yardopts` file is available when looking the code of YARD itself. 🤔 

https://github.com/lsegal/yard/blob/12f56cf7d58e7025085f00b9f9f2f62c24b13d93/.yardopts#L1
https://github.com/lsegal/yard/blob/12f56cf7d58e7025085f00b9f9f2f62c24b13d93/yard.gemspec#L20
https://www.rubydoc.info/gems/yard/file/docs/Tags.md

If so, we will need to add `.yardopts` to the Capybara's gemspec.
Anyway, if I have time, I'll try sharing the macro.

At first, I have to split this PR to smaller PRs. 💪

PR #2196, opened 2019-05-21
Use `@example` YARD tag for `Capybara.string` API doc

ybiquitous:
The YARD `@example` tag (which can take an optional title) makes a generated HTML a bit more readable.
See the [YARD doc](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#example).

## Before

image

## After

image


ybiquitous:
Thank you!

PR #2197, opened 2019-05-22
Make links in `Capybara` module API doc

ybiquitous:
This changes plain-text references to links in the `Capybara` module as possible.
Also, this fixes incorrect references and typos.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -73,20 +73,20 @@ module Capybara
     # [asset_host = String]               Where dynamic assets are hosted - will be prepended to relative asset locations if present (Default: nil)
     # [run_server = Boolean]              Whether to start a Rack server for the given Rack app (Default: true)
     # [raise_server_errors = Boolean]     Should errors raised in the server be raised in the tests? (Default: true)
-    # [server_errors = Array\<Class\>]    Error classes that should be raised in the tests if they are raised in the server and Capybara.raise_server_errors is true (Default: [Exception])
+    # [server_errors = Array\<Class\>]    Error classes that should be raised in the tests if they are raised in the server and {configure raise_server_errors} is true (Default: [Exception])
     # [default_selector = :css/:xpath]    Methods which take a selector use the given type by default (Default: :css)
     # [default_max_wait_time = Numeric]   The maximum number of seconds to wait for asynchronous processes to finish (Default: 2)
     # [ignore_hidden_elements = Boolean]  Whether to ignore hidden elements on the page (Default: true)
     # [automatic_reload = Boolean]        Whether to automatically reload elements as Capybara is waiting (Default: true)
     # [save_path = String]  Where to put pages saved through save_(page|screenshot), save_and_open_(page|screenshot) (Default: Dir.pwd)
-    # [automatic_label_click = Boolean]   Whether Node#choose, Node#check, Node#uncheck will attempt to click the associated label element if the checkbox/radio button are non-visible (Default: false)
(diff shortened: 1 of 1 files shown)


lib/capybara.rb
```diff
@@ -73,20 +73,20 @@ class << self
...
     # [default_selector = :css/:xpath]    Methods which take a selector use the given type by default (Default: :css)
     # [default_max_wait_time = Numeric]   The maximum number of seconds to wait for asynchronous processes to finish (Default: 2)
     # [ignore_hidden_elements = Boolean]  Whether to ignore hidden elements on the page (Default: true)
     # [automatic_reload = Boolean]        Whether to automatically reload elements as Capybara is waiting (Default: true)
     # [save_path = String]  Where to put pages saved through save_(page|screenshot), save_and_open_(page|screenshot) (Default: Dir.pwd)
-    # [automatic_label_click = Boolean]   Whether Node#choose, Node#check, Node#uncheck will attempt to click the associated label element if the checkbox/radio button are non-visible (Default: false)
+    # [automatic_label_click = Boolean]   Whether {Capybara::Node::Actions#choose}, {Capybara::Node::Actions#check}, {Capybara::Node::Actions#uncheck} will attempt to click the associated label element if the checkbox/radio button are non-visible (Default: false)
```
twalpole: I don't like these pointing to Capybara::Node::Actions since Actions is a module purely for code separation and these methods are actually called on instances of Capybara::Node::Element  -  Do you know if there is a way to get yard to parse the methods into Element, and link to that instead?

ybiquitous: Umm… I don't know how… 😕 

But when writing `Capybara::Node::Element#choose` instead of `{Capybara::Node::Actions#choose}`, YARD generates a link to `Capybara::Node::Actions#choose` (maybe I think YARD searchs actually defined documents).

See below:

image

```diff
diff --git a/lib/capybara.rb b/lib/capybara.rb
index cdc23b88..3ec3f941 100644
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -79,7 +79,7 @@ module Capybara
     # [ignore_hidden_elements = Boolean]  Whether to ignore hidden elements on the page (Default: true)
     # [automatic_reload = Boolean]        Whether to automatically reload elements as Capybara is waiting (Default: true)
     # [save_path = String]  Where to put pages saved through save_(page|screenshot), save_and_open_(page|screenshot) (Default: Dir.pwd)
-    # [automatic_label_click = Boolean]   Whether {Capybara::Node::Actions#choose}, {Capybara::Node::Actions#check}, {Capybara::Node::Actions#uncheck} will attempt to click the associated label element if the checkbox/radio button are non-visible (Default: false)
+    # [automatic_label_click = Boolean]   Whether {Capybara::Node::Element#choose}, {Capybara::Node::Element#check}, {Capybara::Node::Element#uncheck} will attempt to click the associated label element if the checkbox/radio button are non-visible (Default: false)
     # [enable_aria_label = Boolean]  Whether fields, links, and buttons will match against aria-label attribute (Default: false)
     # [reuse_server = Boolean]  Reuse the server thread between multiple sessions using the same app object (Default: true)
     # [threadsafe = Boolean]  Whether sessions can be configured individually (Default: false)
```

ybiquitous: If you hope, I'm ready to rewrite `{Capybara::Node::Actions#…}` to `{Capybara::Node::Element#…}` like above. 💪

twalpole: Please do rewrite it to ::Element - that way if I ever figure out how to get Yard to just treat those methods as part of Element the links should just continue working.

twalpole: Actually if you could write them as {Capybara::Node::Element#choose Element#choose} I think that would be best (so it shows as Element#choose in the docs)

ybiquitous: It makes sense! I pushed the change. 
The following is a screenshot on my local machine ⬇️ 

image

twalpole:
Thankyou

PR #2198, opened 2019-05-22
Small improvements for `Capybara::Helpers` API doc

ybiquitous:
- Change plain-text references to links.
- ~~Use the YARD syntax for inline code, instead of the Markdown syntax.~~ See https://github.com/teamcapybara/capybara/pull/2198#discussion_r286291677.

### DIFF
--- a/lib/capybara/helpers.rb
+++ b/lib/capybara/helpers.rb
@@ -41,7 +41,7 @@ module Capybara
     ##
     #
     # Injects a `<base>` tag into the given HTML code, pointing to
-    # `Capybara.asset_host`.
+    # {Capybara.configure asset_host}.
     #
     # @param [String] html     HTML code to inject into
     # @param [URL] host (Capybara.asset_host) The host from which assets should be loaded


lib/capybara/helpers.rb
```diff
@@ -40,8 +40,8 @@ def to_regexp(text, exact: false, all_whitespace: false, options: nil)
 
     ##
     #
-    # Injects a `<base>` tag into the given HTML code, pointing to
-    # `Capybara.asset_host`.
+    # Injects a +<base>+ tag into the given HTML code, pointing to
```
twalpole: This changes breaks the docs when built locally (bundle exec rake yard) - I think rather than swapping out all of the markdown we probably just need a .yardopts file specifying that markdown is being used so rubydoc.info realizes that.

ybiquitous: Oh, honestly I couldn't notice the specified yard option. 😓 

https://github.com/teamcapybara/capybara/blob/87e79c2daf75035e6d5d22b43ba25cc9a7cd9920/Rakefile#L44

However sadly, `bundle exec rake yard` command breaks also `Capybara.configure` doc… ⬇️ 

image

twalpole: hmmm -- I wonder which breaks more things

ybiquitous: OK, anyway now I revert inline-code styles changed in this PR, for the future to switch to markdown styles.

In this PR, I just fix links. 💪

ybiquitous: About the current problem that 2 styles of YARD and Markdown are mixed, let's consider on another occasion.

ybiquitous: Now I rebased the commit and modified this PR's description. Please review again. 🙏

twalpole:
thanks!

PR #2199, opened 2019-05-22
Improve `waiting_behavior` YARD macro

ybiquitous:
## Summary

- Unify the almost same macro definitions defined both in `Capybara::Node::Actions` and `Capybara::Node::Finders`.
- ~~The new macro definition is in `Capybara::Node::Base`.~~
  - ~~Please give me help if anyone knows a better place! 🙏~~
  - **→Resolved. See <https://github.com/teamcapybara/capybara/pull/2199#issuecomment-495012213>.**
- Change the `+find+` word in the macro to `this method`.
  - See <https://github.com/teamcapybara/capybara/pull/2190/files#r285646233>.

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -6,9 +6,10 @@ module Capybara
       # @!macro waiting_behavior
       #   If the driver is capable of executing JavaScript, this method will wait for a set amount of time
       #   and continuously retry finding the element until either the element is found or the time
-      #   expires. The length of time +find+ will wait is controlled through {Capybara.default_max_wait_time}
+      #   expires. The length of time this method will wait is controlled through {Capybara.configure default_max_wait_time}.
       #
-      #   @option options [false, true, Numeric] wait (Capybara.default_max_wait_time) Maximum time to wait for matching element to appear.
+      #   @option options [false, true, Numeric] wait
+      #     Maximum time to wait for matching element to appear. Defaults to {Capybara.configure default_max_wait_time}.
 
       ##
       #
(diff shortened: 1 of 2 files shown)


lib/capybara/node/base.rb
```diff
@@ -2,6 +2,14 @@
 
 module Capybara
   module Node
+    # @!macro waiting_behavior
+    #   If the driver is capable of executing JavaScript, this method will wait for a set amount of time
+    #   and continuously retry finding the element until either the element is found or the time
+    #   expires. The length of time this method will wait is controlled through {Capybara.configure default_max_wait_time}.
+    #
+    #   @option options [false, true, Numeric] wait
```
ybiquitous: I have a question. The `default_max_wait_time` option defined in `Capybara.configure` can take a `Numeric` value, but on the other hand, the option here can take `false`, `true`, or `Numeric`. Is this correct? 🤔 

https://github.com/teamcapybara/capybara/blob/87e79c2daf75035e6d5d22b43ba25cc9a7cd9920/lib/capybara.rb#L78

twalpole: true/false are for overriding behaviors around waiting, they are not valid values for default_max_wait_time - false ends up being equivalent to `wait: 0` while true can locally override things like the setting of `predicates_wait`

twalpole:
Can't the macro definition be left in one of the files it's actually used in?

lib/capybara/node/base.rb
ybiquitous: Thanks. I see. 👍

ybiquitous:
> Can't the macro definition be left in one of the files it's actually used in?

When I try, it works! Thanks! 😄 

Because the macro defined in `Finders#find` is hard to find, I decide to leave one defined in `Actions`.

ybiquitous:
I rebased the change and updated this PR description!

twalpole:
Thank you

PR #2201, opened 2019-05-24
Improve `Capybara::Node::Element` API doc

ybiquitous:
Summary:
- Change plain-text references to links.
- Add missing descriptions.
- Fix the broken list in `#send_keys` doc.

The current doc at v3.20.2: <https://www.rubydoc.info/gems/capybara/3.20.2/Capybara/Node/Element>

### DIFF
--- a/lib/capybara/node/element.rb
+++ b/lib/capybara/node/element.rb
@@ -43,7 +43,7 @@ module Capybara
 
       ##
       #
-      # Retrieve the text of the element. If `Capybara.ignore_hidden_elements`
+      # Retrieve the text of the element. If {Capybara.configure ignore_hidden_elements}
       # is `true`, which it is by default, then this will return only text
       # which is visible. The exact semantics of this may differ between
       # drivers, but generally any text within elements with `display:none` is
@@ -61,7 +61,7 @@ module Capybara
 
       ##
       #
-      # Retrieve the given attribute
+      # Retrieve the given attribute.
       #
       #     element[:title] # => HTML title attribute
       #
@@ -74,7 +74,7 @@ module Capybara
 
       ##
       #
-      # Retrieve the given CSS styles
+      # Retrieve the given CSS styles.
       #
       #     element.style('color', 'font-size') # => Computed values of CSS 'color' and 'font-size' styles
       #
@@ -109,7 +109,7 @@ module Capybara
       # Set the value of the form element to the given value.
       #
       # @param [String] value    The new value
-      # @param [Hash{}] options  Driver specific options for how to set the value. Take default values from `Capybara#default_set_options` - See {Capybara::configure}
+      # @param [Hash] options    Driver specific options for how to set the value. Take default values from {Capybara.configure default_set_options}.
       #
       # @return [Capybara::Node::Element]  The element
       def set(value, **options)
@@ -122,7 +122,7 @@ module Capybara
 
       ##
       #
-      # Select this node if it is an option element inside a select tag
+      # Select this node if it is an option element inside a select tag.
       #
       # @!macro action_waiting_behavior
       #   If the driver dynamic pages (JS) and the element is currently non-interactable, this method will
@@ -139,7 +139,7 @@ module Capybara
 
       ##
       #
-      # Unselect this node if it is an option element inside a multiple select tag
+      # Unselect this node if it is an option element inside a multiple select tag.
       #
       # @macro action_waiting_behavior
       # @return [Capybara::Node::Element]  The element
@@ -150,7 +150,7 @@ module Capybara
 
       ##
       #
-      # Click the Element
+      # Click the Element.
       #
       # @macro action_waiting_behavior
       # @!macro click_modifiers
@@ -169,7 +169,7 @@ module Capybara
 
       ##
       #
-      # Right Click the Element
+      # Right Click the Element.
       #
       # @macro action_waiting_behavior
       # @macro click_modifiers
@@ -183,7 +183,7 @@ module Capybara
 
       ##
       #
-      # Double Click the Element
+      # Double Click the Element.
       #
       # @macro action_waiting_behavior
       # @macro click_modifiers
@@ -197,7 +197,7 @@ module Capybara
 
       ##
       #
-      # Send Keystrokes to the Element
+      # Send Keystrokes to the Element.
       #
       # @overload send_keys(keys, ...)
       #   @param keys [String, Symbol, Array<String,Symbol>]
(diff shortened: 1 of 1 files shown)


lib/capybara/node/element.rb
```diff
@@ -43,14 +43,15 @@ def native
...
-      # Retrieve the text of the element. If `Capybara.ignore_hidden_elements`
+      # Retrieve the text of the element. If {Capybara.configure ignore_hidden_elements}
       # is `true`, which it is by default, then this will return only text
       # which is visible. The exact semantics of this may differ between
       # drivers, but generally any text within elements with `display:none` is
       # ignored. This behaviour can be overridden by passing `:all` to this
       # method.
       #
       # @param type [:all, :visible]  Whether to return only visible or all text
+      # @param normalize_ws [Boolean] When `true`, remove all whitespace from the returning text
```
ybiquitous: [note] I think this `normalize_ws` is required and add it, but I'm worried whether this modification is correct… 🤔

lib/capybara/node/element.rb
```diff
@@ -197,73 +198,73 @@ def double_click(*keys, wait: nil, **offset)
...
+      # * :f5
+      # * :f6
+      # * :f7
+      # * :f8
+      # * :f9
+      # * :f10
+      # * :f11
+      # * :f12
+      # * :meta
+      # * :command      - alias of :meta
```
ybiquitous: 🙀 

image


https://www.rubydoc.info/gems/capybara/3.20.2/Capybara/Node/Element#send_keys-instance_method

lib/capybara/node/element.rb
```diff
@@ -429,17 +430,17 @@ def drop(*args)
...
       #   @param [:top, :bottom, :center, :current] position
-      #   @param :offset
+      #   @param [[Integer, Integer]] offset
       #
-      # Scroll the page or current element until the given element is aligned at the top, bottom, or center of it
       # @overload scroll_to(element, align: :top)
+      #   Scroll the page or current element until the given element is aligned at the top, bottom, or center of it.
       #   @param [Capybara::Node::Element] element   The element to be scrolled into view
-      #   @param [:top, :bottom, :center] :align Where to align the element being scrolled into view with relation to the current page/element if possible
+      #   @param [:top, :bottom, :center] align Where to align the element being scrolled into view with relation to the current page/element if possible
```
ybiquitous: [note] I think the positions of overload methods is incorrect…

image

https://www.rubydoc.info/gems/capybara/3.20.2/Capybara/Node/Element#scroll_to-instance_method

lib/capybara/node/element.rb
```diff
@@ -429,17 +430,17 @@ def drop(*args)
 
       ##
       #
-      # Scroll the page or element
+      # Scroll the page or element.
       #
-      # Scroll the page or element to its top, bottom or middle
       # @overload scroll_to(position, offset: [0,0])
+      #   Scroll the page or element to its top, bottom or middle.
```
ybiquitous: [question] Is `middle` a mistake of `center`? The parameter description is below: ⬇️ 

```ruby
# @param [:top, :bottom, :center, :current] position
```

lib/capybara/node/element.rb
```diff
@@ -429,17 +430,17 @@ def drop(*args)
...
       #   @param [:top, :bottom, :center, :current] position
-      #   @param :offset
+      #   @param [[Integer, Integer]] offset
```
ybiquitous: [note] I add a type to the `offset` parameter, but I'm worried a bit whether it's correct… 🤔

lib/capybara/node/element.rb
twalpole: `normalize_ws` is something that was added as a stopgap to add migration from Capybara 2 to 3.  It wasn't added to the documentation because its use really shouldn't be encouraged and will probably go away in Capybara 4.  Also it doesn't remove all whitespace, it normalizes the whitespace (collapses multiple whitespace to one, strips beginning and end, etc)

ybiquitous: Oh, I see. 👍 
But `normailze_ws` is yet confused because the method signature contains it.

image
https://www.rubydoc.info/gems/capybara/3.20.2/Capybara/Node/Element#text-instance_method

How about removing it from doc explicitly by using `@overload`? Like this: ⬇️ 

```ruby
# @overload text(type = nil)
#   @param type [:all, :visible]  Whether to return only visible or all text
#   @return [String]              The text of the element
def text(type = nil, normalize_ws: false)
```

I try it on my local: ⬇️ 
image

lib/capybara/node/element.rb
twalpole: The type is correct - please note that it defaults to [0,0]

lib/capybara/node/element.rb
twalpole: LGTM

lib/capybara/node/element.rb
twalpole: In the text description center and middle are generally equivalent. If you want to change it to center to match the parameter value I don't have an issue with that.

lib/capybara/node/element.rb
```diff
@@ -462,11 +463,11 @@ def scroll_to(pos_or_el_or_x, y = nil, align: :top, offset: nil)
       ##
       #
       # Execute the given JS in the context of the element not returning a result. This is useful for scripts that return
-      # complex objects, such as jQuery statements. +execute_script+ should be used over
-      # +evaluate_script+ whenever possible. `this` in the script will refer to the element this is called on.
+      # complex objects, such as jQuery statements. {#execute_script} should be used over
+      # {#evaluate_script} whenever possible. `this` in the script will refer to the element this is called on.
```
twalpole: Please change "whenever possible" to "whenever a result is not expected or needed"

lib/capybara/node/element.rb
ybiquitous: OK, I keep as it is. 👍

lib/capybara/node/element.rb
```diff
@@ -510,6 +511,11 @@ def evaluate_async_script(script, *args)
         JS
       end
 
+      ##
```
twalpole: Please don't document this, and instead mark it as `@api private` - It's not really intended for outside use

lib/capybara/node/element.rb
ybiquitous: Thanks 👍

lib/capybara/node/element.rb
twalpole: I think for now I just want to leave this the way it was, shows in the method but isn't otherwise documented.

ybiquitous:
I pushed 79a9c0ce791a78c871cec78981120668f90869aa to fix all reviews.
After this PR is approved, I will rebase it.

twalpole:
LGTM - please rebase and then I'll merge - Note: if you end your PR description with "[ci skip]" when only changing docs it will prevent firing off the travis test runs.

ybiquitous:
Thanks a lot! I rebased now.

PR #2209, opened 2019-05-27
Improve `Capybara::Selector` API doc

ybiquitous:
- Fix invalid indentations.
- Change plain-text references to links as possible.
- Mark HTML tags as code (e.g. `<a>`) in order to avoid shown as HTML.
- Replace a non-ascii character(`—`) with a ascii character(`-`), and remove RuboCop disable comments.
  - The characters are mixed in the current doc.

See the latest doc: <https://www.rubydoc.info/gems/capybara/3.21.0/Capybara/Selector>

[skip ci]


ybiquitous:
I recommend seeing this diff on "Hide whitespace changes" mode.

PR #2210, opened 2019-05-27
Markup API doc in Markdown style

ybiquitous:
- Specify `--markup markdown` option to YARD.
- Add and publish `.yardopts` file, instead of giving the option in `Rakefile`.
  This `.yardopts` file allows to use the same options on <https://www.rubydoc.info/>.
- Rewrite the `Capybara.configure` method doc because it's broken in Markdown style. See below.
  - The latest doc is [here](https://www.rubydoc.info/gems/capybara/3.21.0/Capybara#configure-class_method).

[skip ci]

## `Capybara.configure`  change

In addition, I re-sort the options in alphabetical order because I think too hard to find an option in the long list.
But this big change may cause a discussion. Please give me any feedback! 

### Before

Run `bundle exec rake yard`:
image

### After

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ bundler_stubs
 bin
 gem-private_key.pem
 save_path_tmp
+vendor/bundle
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -68,143 +68,42 @@ module Capybara
     #
     # #### Configurable options
     #
-    # * **allow_gumbo**
-    #   * When `nokogumbo` is available, whether it will be used to parse HTML strings.
-    #   * Type: `Boolean`
-    #   * Default: `false`
-    #
-    # * **always_include_port**
-    #   * Whether the Rack server's port should automatically be inserted into every visited URL
-    #     unless another port is explicitly specified.
-    #   * Type: `Boolean`
-    #   * Default: `false`
-    #
-    # * **app_host**
-    #   * The default host to use when giving a relative URL to visit, must be a valid URL e.g. `http://www.example.com`.
-    #   * Type: `String`, `nil`
-    #
-    # * **asset_host**
-    #   * Where dynamic assets are hosted - will be prepended to relative asset locations if present.
-    #   * Type: `String`
-    #   * Default: `nil`
-    #
-    # * **automatic_label_click**
-    #   * Whether {Capybara::Node::Element#choose Element#choose}, {Capybara::Node::Element#check Element#check}, {Capybara::Node::Element#uncheck Element#uncheck}
-    #     will attempt to click the associated `<label>` element if the checkbox/radio button are non-visible.
-    #   * Type: `Boolean`
-    #   * Default: `false`
-    #
-    # * **automatic_reload**
-    #   * Whether to automatically reload elements as Capybara is waiting.
-    #   * Type: `Boolean`
-    #   * Default: `true`
-    #
-    # * **default_max_wait_time**
-    #   * The maximum number of seconds to wait for asynchronous processes to finish.
-    #   * Type: `Numeric`
-    #   * Default: `2`
-    #
-    # * **default_normalize_ws**
-    #   * Whether text predicates and matchers use normalize whitespace behavior.
-    #   * Type: `Boolean`
-    #   * Default: `false`
-    #
-    # * **default_selector**
-    #   * Methods which take a selector use the given type by default. See also {Capybara::Selector}.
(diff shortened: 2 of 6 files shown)


ybiquitous:
@twalpole My documentation modification started from #2190 is last with this PR. I believe there are no other places which we should fix about documentation. Thanks a lot for your kindness. 😄

.gitignore
```diff
@@ -21,3 +21,4 @@ bundler_stubs
 bin
 gem-private_key.pem
 save_path_tmp
+vendor/bundle
```
twalpole: Why was this necessary?

twalpole:
I would prefer the configure doc changes to have the type and default on the same line as the option name if possible - I find it tough to scan for anything the way it is now.

.gitignore
ybiquitous: Sorry, I miss the reason. 🙏 

I prefer a style to use the [`bundle install --deployment`](https://bundler.io/v2.0/man/bundle-install.1.html#DEPLOYMENT-MODE) command, which installs the gems to `vendor/bundle` directory.
If you don't like it, I'll remove it. 👍

ybiquitous:
Make sense. I'll try another better solution. 💪 

BTW, when I rebased #2209, `Capybara::Selector` doc breaks, so I'll fix it also.

lib/capybara.rb
```diff
@@ -66,39 +66,145 @@ class << self
     #       config.app_host   = 'http://www.google.com'
     #     end
     #
-    # === Configurable options
-    #
-    # [app_host = String/nil]             The default host to use when giving a relative URL to visit, must be a valid URL e.g. http://www.example.com
```
ybiquitous: [question] A description about `app_host` default value is missing, is it no problem?

ybiquitous:
I changed from multi-line style to one-line style via 00376db3e088f6da9e41f2ee3d7dd3ed8298dd3a, how about?

```markdown
- **{name}** ({Type} = `{default value}`) - {description}

- **allow_gumbo** (Boolean = `false`) - When `nokogumbo` is available, whether it will be used to parse HTML strings.
```

Screenshot:

image

lib/capybara/selector.rb
```diff
@@ -19,156 +19,156 @@
...
 #   * Locator: Match id or name
 #   * Filters:
-#     * :id (String, Regexp, XPath::Expression) - Match id attribute
-#     * :name (String) - Match name attribute
-#     * :class (String, Array<String>, Regexp, XPath::Expression) - Matches the class(es) provided
-#     * :style (String, Regexp, Hash)
+#       * :id (String, Regexp, XPath::Expression) - Match id attribute
+#       * :name (String) - Match name attribute
+#       * :class (String, Array<String>, Regexp, XPath::Expression) - Matches the class(es) provided
+#       * :style (String, Regexp, Hash)
```
ybiquitous: [note] It's necessary to deepen 3rd level indentation, please see an example below (which was one before the update).

image

lib/capybara/selector.rb
```diff
@@ -19,156 +19,156 @@
...
 # * **:checkbox** - Find checkboxes
 #   * Locator: Match id, {Capybara.configure test_id} attribute, name, or associated label text
 #   * Filters:
-#     * *:id (String, Regexp, XPath::Expression) - Matches the id attribute
-#     * *:name (String) - Matches the name attribute
-#     * *:class (String, Array<String>, Regexp, XPath::Expression) - Matches the class(es) provided
-#     * *:checked (Boolean) - Match checked fields?
-#     * *:unchecked (Boolean) - Match unchecked fields?
-#     * *:disabled (Boolean) - Match disabled field?
-#     * *:option (String) - Match the value
```
ybiquitous: [note] I think they are with an extra asterisk(`*`).

ybiquitous:
I fixed the 2 problems, please review again and give me feedbacks. 🙏

.gitignore
twalpole: Ok -- it's fine

lib/capybara.rb
twalpole: It defaults to nil - so it's fine

lib/capybara/selector.rb
twalpole: yes

twalpole:
LGTM - Thanks a lot for all this work 👏

ybiquitous:
You are welcome.
Thank you so much, too! ☺️

Issue #2221, opened 2019-06-19
AnimationDisabler does not disable animations on :before or :after


DanielHeath:
In https://github.com/teamcapybara/capybara/blob/master/lib/capybara/server/animation_disabler.rb#L11 the selector is `*`, which misses some animations. 

`*, *:after, *:before` would catch more, as `:after` is not matched by `*`, but can be animated.

twalpole:
PRs accepted

DanielHeath:
awesome, will take a look in a bit.

DanielHeath:
so, this is very hard to add tests for, because webdriver doesn't expose pseudoelement content as document content.

PR #2222, opened 2019-06-19
Disable animations on pseudoelement content as well as regular elements

DanielHeath:
Resolves https://github.com/teamcapybara/capybara/issues/2221

Very hard to add test cases for due to https://github.com/SeleniumHQ/selenium/issues/2517 .

### DIFF
--- a/lib/capybara/server/animation_disabler.rb
+++ b/lib/capybara/server/animation_disabler.rb
@@ -46,7 +46,7 @@ module Capybara
       DISABLE_MARKUP_TEMPLATE = <<~HTML
         <script defer>(typeof jQuery !== 'undefined') && (jQuery.fx.off = true);</script>
         <style>
-           %<selector>s {
+           %<selector>s, %<selector>s::before, %<selector>s::after {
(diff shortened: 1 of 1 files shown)


twalpole:
Because these are referring to pseudo elements (vs pseudo classes) I believe it should be ::before and ::after (double colon)

DanielHeath:
Ahh - https://developer.mozilla.org/en-US/docs/Web/CSS/::before indicates that `::before` is preferred, though both are accepted by browsers.

twalpole:
As for testing - it can probably be done by registering a `transitionend` and `animationend` handler that add something to the page and making sure they don't fire when animations are disabled.  Haven't actually tried that though so it's only "probably"

DanielHeath:
I put a bit of time in to figuring it out; this kind of issue tends to only trigger on more complex layouts. I might be able to make the `:before` content cover the link after it's been clicked, then assert that clicking the link obscures it.

twalpole:
I've added a test for this - so merging now -- thanks!

PR #2226, opened 2019-07-05
Optionally allow `all` results to be reloaded when stale

twalpole:
When running through a list of links on a page with code like

    all('#my_container a').each do |link|
      link.click
      expect(page).to have_text 'something'
      go_back
    end

it would be nice to allow the `all` results to be reloaded, otherwise the elements will be stale.  This should not be the default behavior because on dynamic pages or pages with changing sets of links between refreshes the results can be highly confusing.


luke-hill:
This would be a cool and welcome addition! If you manage to revisit and spec this up and release it it would be amazing.

PR #2230, opened 2019-07-18
Fix code example indentations in YARD

ybiquitous:
Because the YARD option `--markup markdown` has been given since PR #2210, **4** spaces is needed for code snippet (code example) instead of 2 spaces.

For example, see <https://www.rubydoc.info/gems/capybara/3.26.0/Capybara/Node/Matchers#assert_all_of_selectors-instance_method>.

image

See also the [Markdown spec](https://daringfireball.net/projects/markdown/syntax#precode) below:

> To produce a code block in Markdown, simply indent every line of the block by at least 4 spaces or 1 tab.

[skip ci]

### DIFF
--- a/lib/capybara/node/matchers.rb
+++ b/lib/capybara/node/matchers.rb
@@ -141,8 +141,8 @@ module Capybara
       # or descendants of the current node.  If options are provided, the assertion
       # will check that each locator is present with those options as well (other than `:wait`).
       #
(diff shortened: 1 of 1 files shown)


twalpole:
Thanks

ybiquitous:
You're welcome! This bug is a part of my mistake. 🙇 

It seems there are no such bugs in other places.

PR #2231, opened 2019-07-18
Use `@see` tag of YARD instead of plain text "see"

ybiquitous:
Current doc: <http://www.rubydoc.info/gems/capybara/3.26.0/Capybara/Minitest/Assertions>

About `@see` tag: <https://www.rubydoc.info/gems/yard/file/docs/Tags.md#see>

[skip ci]


twalpole:
Thanks again

PR #2241, opened 2019-08-29
Add `no-sandbox` option for selenium_chrome_headless

toshimaru:
## Issue

When I tried to run RSpec on my docker environment, I ran into an error:

```
     Selenium::WebDriver::Error::UnknownError:
       unknown error: Chrome failed to start: exited abnormally
         (unknown error: DevToolsActivePort file doesn't exist)
         (The process started from chrome location /usr/bin/chromium is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
```

## Cause

Enabling Chrome sandbox causes this error.

> Passing '--no-sandbox' flag when creating your WebDriver session. Special test environments sometimes cause Chrome to crash when the sandbox is enabled.

via. [Chrome doesn't start or crashes immediately - ChromeDriver - WebDriver for Chrome](https://sites.google.com/a/chromium.org/chromedriver/help/chrome-doesn-t-start)

## Solution

Adding `--no-sandbox` option fixed the issue.

## Environment

My `Dockerfile` is:

```dockerfile
FROM ruby:2.6
RUN apt-get update -qq && apt-get install -y nodejs chromium-driver
```

Base image is `ryby:2.6` and `chromium-driver` is installed.

### DIFF
--- a/lib/capybara/registrations/drivers.rb
+++ b/lib/capybara/registrations/drivers.rb
@@ -28,6 +28,7 @@ Capybara.register_driver :selenium_chrome_headless do |app|
   Capybara::Selenium::Driver.load_selenium
   browser_options = ::Selenium::WebDriver::Chrome::Options.new.tap do |opts|
     opts.args << '--headless'
+    opts.args << "--no-sandbox"
     opts.args << '--disable-gpu' if Gem.win_platform?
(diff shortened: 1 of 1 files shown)


twalpole:
This option isn’t required for people not running in a container environment and has negative security implications so we won’t be adding it to the default config. If you need it in your environment you should register your own driver configuration

PR #2244, opened 2019-09-07
Fix eaddrinuse

lukaso:
This fixes a problem where Capybara is trying to boot a server and throws an `Errno::EADDRINUSE` exception.

I know the test is calling a private method, however, I don't know how you would like that to be dealt with, or if it is OK.

### DIFF
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -105,10 +105,23 @@ module Capybara
     end
 
     def find_available_port(host)
-      server = TCPServer.new(host, 0)
-      server.addr[1]
-    ensure
-      server&.close
+      port = 0
+      while port.zero?
+        begin
+          server = TCPServer.new(host, port)
+          port = server.addr[1]
+        ensure
+          server&.close
+        end
+        begin
+          server = TCPServer.new(host, port)
+        rescue Errno::EADDRINUSE
+          port = 0
+        ensure
+          server&.close
+        end
+      end
+      port
     end
   end
 end
--- a/spec/server_spec.rb
+++ b/spec/server_spec.rb
@@ -74,6 +74,21 @@ RSpec.describe Capybara::Server do
     end
   end
 
+  def use_port(host, port)
+    server = TCPServer.new(host, port)
+  ensure
+    server&.close
+  end
+
+  it 'should handle that getting available ports fails randomly' do
+    expect {
+      100000.times do |count|
+        port = Capybara::Server.new(Object.new).send(:find_available_port, "0.0.0.0")
+        use_port("0.0.0.0", port)
+      end
+    }.not_to raise_error
+  end
+
   it 'should return its #base_url' do
     app = proc { |_env| [200, {}, ['Hello Server!']] }
     server = described_class.new(app).boot
--- a/spec/server_spec.rb
+++ b/spec/server_spec.rb
@@ -81,12 +81,12 @@ RSpec.describe Capybara::Server do
   end
 
   it 'should handle that getting available ports fails randomly' do
-    expect {
-      100000.times do |count|
-        port = Capybara::Server.new(Object.new).send(:find_available_port, "0.0.0.0")
-        use_port("0.0.0.0", port)
+    expect do
+      100000.times do
+        port = described_class.new(Object.new).send(:find_available_port, '0.0.0.0')
+        use_port('0.0.0.0', port)
       end
-    }.not_to raise_error
+    end.not_to raise_error
   end
 
   it 'should return its #base_url' do


twalpole:
Unfortunately I don't think this actually fixes anything.  The race is between when find_available_port returns a currently unused port and when the rack app is actually started inside `Capybara::Server#boot`.  This change doesn't actually change the length of time between those two, so theoretically shouldn't make any difference.  In reality it may make a slight difference depending on how free ports are determined/assigned by the ruby/system implementation being run on, but doesn't actually fix anything.

lukaso:
I disagree. My two hypotheses were what you say 1) that the same port is being offered twice and 2) that the time between the two TCPSocket calls was the problem. Neither turned out to be true.

If you run the failing test, you will see that if fails regardless of the amount of time between the calls, and without any race condition. In my tests, I was able to find ports that did take longer to close, so putting a 1 second, or 5 second wait worked, but then there were other ports where no amount of waiting helped.

This fix stops the failing test from failing.

If you take the first commit, it's just the failing test and you will see that it fails. You can also put in waits after closing the socket, and see that this fixes some of the problems, but not all. My fix fixes all the failures.

I absolutely agree there is a race condition as you describe, but that was not the error we have been seeing on our CI tests as we were starting 3 sessions at separate times.

Let me know if this makes sense, and I can add log outputs if that will help.

On my mac I ran this. You see some ports where I have put sleeps to see if that fixes anything.

```ruby
def find_available_port(host)
  server = TCPServer.new(host, 0)
  server.addr[1]
ensure
  port = server.addr[1]
  server&.close
  # these were fixed by waiting 1 second
  if [50895, 50935, 50936].include? port
    sleep(1)
  end
  # No amount of waiting fixed this port
  if [50936].include? port
    sleep(20)
  end
end
```

twalpole:
"we were starting 3 sessions at separate times" - the race condition isn't with starting multiple sessions, it's with other processes on the system acquiring ports.

The results of your test would tend to indicate a pretty serious bug in ruby, in that it's returning ports that are already in use by another process which shouldn't be happening.

twalpole:
Looking further this has to do with specifying '0.0.0.0' for the host when one of the interfaces on the machine already has that port assigned, but another doesn't. Are you actually using '0.0.0.0' while testing rather than specifying the actual interface you want the tests to run on?

twalpole:
To clarify - the behavior of TCPServer.new appears to be (although I can't find it documented anywhere) - if passed host of '0.0.0.0' and port of '0' will find a free port on any interface and return a server bound all available interfaces not already using  that port.  However, when passed a host of '0.0.0.0' and an actual port number instead of '0' it requires all available interfaces to not be using that port to succeed.

lukaso:
We haven't specified an interface line 'en0'.

I've now tested the exact same code but with '127.0.01' (which we use on the CI system) and 'localhost' and though the behaviour differs slightly, the end result is the same.

```
...
Host: localhost: Port: 61499: count: 16322
Traceback (most recent call last):
	3: from test_race.rb:29:in `<main>'
	2: from test_race.rb:18:in `use_port'
	1: from test_race.rb:18:in `new'
test_race.rb:18:in `initialize': Address already in use - bind(2) for "localhost" port 61500 (Errno::EADDRINUSE)
```

```
...
Port: 61499: count: 47
Traceback (most recent call last):
	3: from test_race.rb:26:in `<main>'
	2: from test_race.rb:18:in `use_port'
	1: from test_race.rb:18:in `new'
test_race.rb:18:in `initialize': Address already in use - bind(2) for "127.0.0.1" port 61500 (Errno::EADDRINUSE)
```

I agree this is a serious bug. I have only tested on a Mac, but given I see the same behaviour on an ubuntu server running our CI, I have a feeling it's much deeper. Also, using port 0 to get a random port I think is part of the TCP/IP spec, so might be deep in the code.

"Alternatively, it can provide port 0 to bind() as its connection parameter. That triggers the operating system to automatically search for and return a suitable available port in the TCP/IP dynamic port number range." from https://www.lifewire.com/port-0-in-tcp-and-udp-818145

lukaso:
I've run my script with all three interfaces on a docker image and everything worked perfectly, so this may be mac specific (but still matches the behaviour I was seeing on an ubuntu server.

`docker run -it --rm --name my-running-script -v "$PWD":/usr/src/myapp -w /usr/src/myapp ruby:2.5 ruby test_race.rb`

test_race.rb:
```ruby
require 'socket'

def find_available_port(host)
  server = TCPServer.new(host, 0)
  server.addr[1]
ensure
  port = server.addr[1]
  server&.close
  if [50895, 50935, 50936].include? port
    sleep(1)
  end
  if [50936].include? port
    sleep(20)
  end
end

def use_port(host, port)
  server = TCPServer.new(host, port)
ensure
  server&.close
end

host = "127.0.0.1"
# host = "0.0.0.0"
# host = "localhost"
count = 0
while true do
  port = find_available_port(host)
  use_port(host, port)
  count += 1
  puts "Host: #{host}: Port: #{port}: count: #{count}"
end
```

lukaso:
I've run in docker with the exact same version of ruby, and that also doesn't fail `2.6.3p62`, so this particular issue appears to be mac specific.

twalpole:
By “interface” I meant IP address. I would expect similar behavior if “local host” was used on a machine which supports ip4 and ip6 since it will resolve to both 127.0.0.0.1 and ::1 which would allow for the same behavior. As for it not showing up in docket, that could just be the docket instance just not having any processed bound to a port on ly one up address. This is definitely behavior that doesn’t match the docs though since  the docs state “ Internally, ::new calls getaddrinfo() function to obtain addresses. If getaddrinfo() returns multiple addresses, ::new tries to create a server socket for each address and returns first one that is successful.” (or I could be misreading that)

lukaso:
Something definitely feels wrong about what I've found on the Mac. I've been able to reproduce it on a second Mac, so it definitely isn't just one machine (was also on ruby 2.4.4 so a different version).

However, the script doesn't fail on the build box (Kernel: Linux 4.4.0-142-generic Ubuntu 14.04.5 LTS). And the build boxes only run builds so the OS shouldn't be blocking ports too often. The bug manifests itself quite rarely (weekly on hundreds of builds).

There's no evidence that a previous version of the server has started (because it's the first occurrence of that port and all subsequent attempts also fail).

twalpole:
Yeah -- definitely weird behavior. Now that I think I understand the behavior I'm attempting to update your test to be simpler and more deterministic.  Once I have that I'll get this merged.

twalpole:
Hmm.. there doesn't actually appear to be a way to make this deterministic - or I can't figure out exactly what socket settings for the existing socket will actually cause this.

lukaso:
I've built a PR #2248 based on your #2246 which might actually be better. It both reduces the race condition time, and restarts the server.

twalpole:
Unfortunately I don't think that will work as expected - see comments on that PR.  I'm going to go with my simplified version of this PR (pr/2244 branch) once I fix it to run on 2.4 and clean up the rubocop issues.

lukaso:
Thanks!

rmacklin:
I believe this (from https://github.com/teamcapybara/capybara/pull/2244#issuecomment-529106006):
> The race is between when find_available_port returns a currently unused port and when the rack app is actually started inside `Capybara::Server#boot`

is true (there is still a race condition there).

Also, I think this (from https://github.com/teamcapybara/capybara/pull/2244#issuecomment-529178051):

> the race condition isn't with starting multiple sessions, it's with other processes on the system acquiring ports.

is confusing/potentially misleading - "starting multiple sessions" can happen in other processes on the system if you are using [Rails parallel testing with processes](https://guides.rubyonrails.org/testing.html#parallel-testing-with-processes) to run a suite of Capybara tests in parallel.

Even with the fix in this PR + https://github.com/teamcapybara/capybara/commit/4cd9d8790ba8df5fa5f9e292065d560faa14dd23 ([released in `v3.30.0`](https://github.com/teamcapybara/capybara/blob/f05a2a033a958c4fa52889a309cec0829a57a62a/History.md#version-3300)), I have witnessed problems running a parallelized Capybara test suite because the port returned by `find_available_port` is the same in two of the parallel processes. Here is part of the log from a CI job that demonstrates this:
```
Capybara starting Puma...
* Version 3.12.2 , codename: Llamas in Pajamas
* Listening on tcp://127.0.0.1:44895
Capybara starting Puma...
* Version 3.12.2 , codename: Llamas in Pajamas
* Listening on tcp://127.0.0.1:45777
Capybara starting Puma...
* Version 3.12.2 , codename: Llamas in Pajamas
#<Thread:0x0000564bbcb48198@/home/circleci/project/vendor/bundle/ruby/2.6.0/gems/capybara-3.30.0/lib/capybara/server.rb:76 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	10: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/capybara-3.30.0/lib/capybara/server.rb:77:in `block in boot'
	 9: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/capybara-3.30.0/lib/capybara/config.rb:64:in `block in server='
	 8: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/capybara-3.30.0/lib/capybara/registrations/servers.rb:40:in `block in <main>'
	 7: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/capybara-3.30.0/lib/capybara/registrations/servers.rb:40:in `tap'
	 6: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/capybara-3.30.0/lib/capybara/registrations/servers.rb:41:in `block (2 levels) in <main>'
	 5: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/puma-3.12.2/lib/puma/binder.rb:90:in `parse'
	 4: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/puma-3.12.2/lib/puma/binder.rb:90:in `each'
	 3: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/puma-3.12.2/lib/puma/binder.rb:106:in `block in parse'
	 2: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/puma-3.12.2/lib/puma/binder.rb:273:in `add_tcp_listener'
	 1: from /home/circleci/project/vendor/bundle/ruby/2.6.0/gems/puma-3.12.2/lib/puma/binder.rb:273:in `new'
/home/circleci/project/vendor/bundle/ruby/2.6.0/gems/puma-3.12.2/lib/puma/binder.rb:273:in `initialize': Address already in use - bind(2) for "127.0.0.1" port 44895 (Errno::EADDRINUSE)
2020-02-12 02:08:27 WARN Selenium [DEPRECATION] :driver_opts is deprecated. Use :service with an instance of Selenium::WebDriver::Service instead.
Capybara starting Puma...
* Version 3.12.2 , codename: Llamas in Pajamas
* Listening on tcp://127.0.0.1:34159
Capybara starting Puma...
* Version 3.12.2 , codename: Llamas in Pajamas
* Listening on tcp://127.0.0.1:36105
```
(You can see that we're running Capybara 3.30.0 from the stacktrace.)

The first worker's Capybara launched a puma server at port `44895`, and then the third worker also tried to launch its own puma server at that same port and got
```
Address already in use - bind(2) for "127.0.0.1" port 44895 (Errno::EADDRINUSE)
```

What makes this problem perhaps even worse is that the `Errno::EADDRINUSE` exception is thrown in a new `Thread`:
https://github.com/teamcapybara/capybara/blob/6cd27d937ab393848054962352f89e96e78f975a/lib/capybara/server.rb#L76-L78
which doesn't prevent the third worker process from continuing on. So that worker proceeds to run its tests against port `44895` even though that application is controlled by the first worker process, which causes basically all of tests in the third worker to fail (because those tests assume the web app will show data from the third worker's database when it's actually showing data from the first worker's database).

I'm curious if setting `@server_thread.abort_on_exception = true` would be considered improved behavior in Capybara. (Of course, this would not help with the race condition, but it would prevent the third worker process from attempting to run tests that are essentially doomed to fail.) Thoughts?

twalpole:
@rmacklin If running parallel tests (using processes) in Rails you would be better off specifying specific server ports for each process based off the worker number in the parallelize setup, rather than trying to let Capybara find available ports  - untested code but something like

    parallelize_setup do |worker|
        Capybara.server_port = 8080 + worker
    end

   Will start the servers on 8080, 8081, 8082, 8083, etc

rmacklin:
@twalpole Yep, that's what I ended up doing to work around this issue. I'm still curious about setting `@server_thread.abort_on_exception = true` in Capybara, though - it seems undesirable for the process to continue along even when the app server thread raised an error before it could boot.

twalpole:
@rmacklin You're free to set it - but in single process use (the most common) it's undesirable to have everything die due to an exception in the server during a test.  I'm not sure off the top of my head whether or not `abort_on_exception` can be toggled on/off multiple times, but if it can then we could possibly set it to `true` until the server has booted, then change to `false` - might require some research.

PR #2246, opened 2019-09-07
Decrease time for race condition when finding available port

twalpole:
This is an attempt at "fixing"  issue 2245.  The only true fix would be to require servers to somehow return the port they register when a random port is required, rather than attempting to pass them which port to use.  That would be a breaking change so we can't do it currently, and I'm not sure whether it's worth doing if we can minimize the actual time for the race to occur.

### DIFF
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -16,7 +16,7 @@ module Capybara
       end
     end
 
-    attr_reader :app, :port, :host
+    attr_reader :app, :host
 
     def initialize(app,
                    *deprecated_options,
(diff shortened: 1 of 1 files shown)


lukaso:
You could improve this by catching the `Errno::EADDRINUSE` exception in the boot function and then  trying to reserve a port again? What I don't know is how the port is passed back to the rest of capybara for use. If the port is gathered before then, then this would require a larger amount of surgery.

I mean, could you rerun these lines:
```ruby
  @port ||= reserve_available_port(host)
  @checker = Checker.new(@host, self.port)
  Capybara::Server.ports[port_key] = resolved_port
```

and try to boot again?

twalpole:
The port is determined before (it is not passed back from the server). We can't catch the EADDRINUSE exception in the boot function because that starts up the servers (application instances) in another thread - and the edge cases around that with clean shutdown of apps, etc will get hairy.

PR #2247, opened 2019-09-07
Fix keyword argument separation warnings on Ruby master

jeremyevans:
Ruby 2.7 will warn if converting a final hash positional argument
to keyword arguments, and Ruby 3 will no longer do such conversion.

This attempts to fix the warnings, mostly by adding ** when calling
methods that accept keyword arguments.  Unfortunately, that is not
backwards compatible on Ruby <2.7 if the method being called does
not accept keyword arguments.  So to enable backwards compatibility,
allow such methods to accept keyword arguments, even though they
don't actually use the keyword arguments.

I'm guessing this commit doesn't handle all cases it needs to handle,
and it may break on !ruby-head because some additional methods need
to accept keyword arguments.  I'll monitor CI for this PR and attempt to
fix the remaining issues.

Ruby master forwardable also needs to be modified to fix some of these
warnings, I have a patch for that which I will submit upstream later today.

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -23,7 +23,7 @@ module Capybara
       # @return [Capybara::Node::Element]  The element clicked
       #
       def click_link_or_button(locator = nil, **options)
-        find(:link_or_button, locator, options).click
+        find(:link_or_button, locator, **options).click
       end
       alias_method :click_on, :click_link_or_button
 
@@ -39,7 +39,7 @@ module Capybara
       #
       # @return [Capybara::Node::Element]  The element clicked
       def click_link(locator = nil, **options)
-        find(:link, locator, options).click
+        find(:link, locator, **options).click
       end
 
       ##
@@ -55,7 +55,7 @@ module Capybara
       #   @param [Hash] options        See {Capybara::Node::Finders#find_button}
       # @return [Capybara::Node::Element]  The element clicked
       def click_button(locator = nil, **options)
-        find(:button, locator, options).click
+        find(:button, locator, **options).click
       end
 
       ##
@@ -88,7 +88,7 @@ module Capybara
       def fill_in(locator = nil, with:, currently_with: nil, fill_options: {}, **find_options)
         find_options[:with] = currently_with if currently_with
         find_options[:allow_self] = true if locator.nil?
-        find(:fillable_field, locator, find_options).set(with, fill_options)
+        find(:fillable_field, locator, **find_options).set(with, **fill_options)
       end
 
       # @!macro label_click
@@ -119,7 +119,7 @@ module Capybara
       #
       # @return [Capybara::Node::Element]  The element chosen or the label clicked
       def choose(locator = nil, **options)
-        _check_with_label(:radio_button, true, locator, options)
+        _check_with_label(:radio_button, true, locator, **options)
       end
 
       ##
@@ -147,7 +147,7 @@ module Capybara
       #
       # @return [Capybara::Node::Element]  The element checked or the label clicked
       def check(locator = nil, **options)
-        _check_with_label(:checkbox, true, locator, options)
+        _check_with_label(:checkbox, true, locator, **options)
       end
 
       ##
@@ -175,7 +175,7 @@ module Capybara
       #
       # @return [Capybara::Node::Element]  The element unchecked or the label clicked
       def uncheck(locator = nil, **options)
-        _check_with_label(:checkbox, false, locator, options)
+        _check_with_label(:checkbox, false, locator, **options)
       end
 
       ##
@@ -205,7 +205,7 @@ module Capybara
         if el.respond_to?(:tag_name) && (el.tag_name == 'input')
           select_datalist_option(el, value)
         else
-          el.find(:option, value, options).select_option
+          el.find(:option, value, **options).select_option
         end
       end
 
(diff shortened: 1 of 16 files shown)


twalpole:
There has to be a better solution than adding keywords parameter acceptance to methods that don't actually take any parameters, especially since it breaks arity detection of drivers using older API versions.  If that really is required I don't see this making it into the Ruby 2.7 final

jeremyevans:
It's not actually required to do make the methods accept keyword arguments, but it does require more changes (all methods that delegate arguments must not delegate empty option hashes in <2.7).  I'll work on a version that does not require adding keyword arguments to any method not already accepting them.

While things could certainly change before 2.7 final, I would not expect the keyword argument behavior to differ substantially from the current master (which now has warnings for all behavior that will require changes). See https://bugs.ruby-lang.org/issues/14183 for extensive discussion of the reasoning behind the changes .

jeremyevans:
There seems to be a failure on 2.4 and Firefox, and I'm not sure if it is related to these changes: https://travis-ci.org/teamcapybara/capybara/jobs/582190446#L2257 It may be related, I can't tell.

Unfortunately, I can't check for handling all warnings on ruby-head using CI, because all ruby-head tests on Travis are broken and do not run.

twalpole:
The FF tests around modals are broken because geckodriver with FF 69/70 hangs with some modal handling cases.  the head are currently broken because building Selenium master is in flux currently.

twalpole:
Thanks for pointing this out -- I'll take a look and fix any others later this weekend.

twalpole:
ugh -- I don't know that this is worth it -- lots of meta-programming  getting borked... It might be easier to just not care for 2.7 and do a breaking major release to support Ruby 3

jeremyevans:
While it is certainly up to you as maintainer, ruby-core really wants to avoid a Python 2->3 situation where gems are incompatible and do not support both Ruby 2.6, 2.7, and 3.0.  Other than some extra complexity in methods that delegate arguments, it shouldn't be that difficult to have code that works the same between Ruby 2.0-3.0 regarding argument handling.  The code just needs to differentiate the handling of keyword arguments and positional arguments.  In Ruby 3, for methods that accept keyword arguments, keyword argument handling will be separate from positional argument handling, with no automatic conversion between the two, as the automatic conversion has been the source of numerous bugs (see https://bugs.ruby-lang.org/issues/14183).

For your specific examples, in Ruby 2.7:

* `def a(*) super end` indicates that `a` accepts positional arguments but not keyword arguments.  When you call `super`, it passes all arguments as positional arguments.  If the super method accepts keyword arguments, and the last positional argument passed is a hash, and the positional hash is converted to keyword arguments for backwards compatibility, a warning is produced because the behavior will change in 3.0, passing the positional argument as a hash instead of converting to keyword arguments.  You can fix the delegation using `def a(*, **) super end`, assuming you know the super method accepts keyword arguments.
* For `def a(hsh, **options) ... end; a(display: 'abc')`, you are passing keyword options to `a`.  Before Ruby 3, Ruby will convert the keyword arguments to a required positional argument.  In Ruby 3, it will no longer do the conversion, and you will have to pass a hash instead of keywords to `a`:  `a({display: 'abc'})`.

Handling methods that delegate arguments in a way that is backwards compatible is tricky in some cases.  If you can use the `forwardable` or `delegate` libraries from stdlib, that will probably be the easiest way to have code that works across versions.  You will be able to write your own delegation code, but it will add complexity.  The changes in this pull request I think will work for Capybara's use cases, but there are some other use cases where you do need to use separate delegating code in <2.7 and >=2.7 in order to get the same behavior.

I'm willing to do this argument conversion work for Capybara, once the ruby-head CI is fixed.  It will make the delegation code a little ugly for a few Ruby versions, but eventually when you drop <2.7 support, you will be able to get rid of the ugly code.

twalpole:
So frozen strings was removed from 3 because "too breaking" - but this change goes in, and is going to break everything.  I'm putting this on-hold (at least for now) because with the current state of 2.7 it's impossible to support this without breaking current user code

    def f(*a, **b)
      puts "a: #{a}"
      puts "b: #{b}"
    end  
    f('abc' => 3)

in Ruby 2.6
     a: [{"abc" => 3}]

in Ruby 2.7
     b: {"abc" => 3}

when did keywords support non symbol keys??? 

Maybe when this stabilizes I'll revisit, but for now it's not worth the hassle

jeremyevans:
I'm not arguing to reopen this, but as points of clarification:

Support for non-symbol keys in keyword splats was added along with keyword argument separation, as discussed in https://bugs.ruby-lang.org/issues/14183.

Frozen string literals by default would not fix any bugs, only improve performance slightly, and that wasn't considered worth the breakage.  Keyword argument separation fixes serious bugs in argument handling by removing implicit conversion between keyword arguments and positional arguments.  It is very well understood by the core time that code that uses keyword arguments will need to be updated for this, but it is considered necessary to improve the language.  The core team has been discussing this change publicly since December 2017, with ample time for public comment on that ticket.

"Maybe when this stabilizes" implies that the current behavior is in flux, which is not the case.  You should expect Ruby 2.7 to have the same behavior as the master branch in regards to argument handling, and all cases where it warns are cases that need to be updated in order for the code to work the same way in Ruby 3.

twalpole:
Then 2.7 is going to break a lot of things and support/takeup will be slow.

PR #2248, opened 2019-09-08
Retry boot server with new port if Errno::EADDRINUSE

lukaso:
This combines reducing the time between choosing a port and booting the server #2246 with retrying with a new port if booting fails due to a `Errno::EADDRINUSE.

The retry mechanism works when simulating a retry.

### DIFF
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -28,6 +28,7 @@ module Capybara
       @app = app
       @extra_middleware = extra_middleware
       @server_thread = nil # suppress warnings
+      @random_port = false
       @host = deprecated_options[1] || host
       @reportable_errors = deprecated_options[2] || reportable_errors
       @port = deprecated_options[0] || port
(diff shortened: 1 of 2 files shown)


twalpole:
I don't think this will work as expected since it relies on catching exceptions issued in one thread in a different thread.  It also risks leaving orphan threads.

lukaso:
Ruby gets all excepts from a thread on join, so it does work. Here's a bit of code that I used to test that assumption.

```ruby
x = Thread.new do
  raise "Boom"
end

puts 'still going'

x.join

puts "I'm dead"
```

```sh
$ ruby thread.rb
still going
thread.rb:2:in `block in <main>': Boom (RuntimeError)
```

When I forced an exception on server boot on every call (the first time), all the tests passed so I feel confident it works.

lukaso:
@twalpole this now takes the initial port retry and adds retry of the server if that fails for some reason (stops the very small edge case). I'll leave this here for now, while I try the fix you've merged to master to see if the fix worked (might take a few weeks as the error is rare).

As stated above, I've tested where every creation of a server fails the first time, and all tests pass locally, so I believe this change is safe.

lukaso:
Since merging https://github.com/teamcapybara/capybara/pull/2244 the problem has not reoccurred.

lukaso:
The issue has not recurred since https://github.com/teamcapybara/capybara/pull/2244 was merged.

Issue #2249, opened 2019-09-19
Expose pending requests URI


juanca:
**Note:** This is for issues with Capybara.  If you have a howto type question, please ask on the mailing list as requested in the README: http://groups.google.com/group/ruby-capybara

## Meta
Capybara Version: capybara-3.15.1
Driver Information (and browser if relevant): selenium-webdriver-3.142.4

## Expected Behavior

Raising an error on remaining pending requests should display more useful information -- like, the pending request URIs (at the very least). This would help developers understand the state of their tests. 

## Actual Behavior

My codebase is pretty convoluted but here is the gist of my issue. I run a test that makes an XHR into a controller-action -- keep in the mind that this particular XHR is not the testing subject. The controller-action attempts to contact an external URI but fails on a timeout because the external URI is fake. Capybara interprets the behavior as a failing assertion via pending requests because the fake URI timeout is longer than the Capybara pending requests time limit. 

Here is the output of the test:

```
Failures:

  1) I am a test
     Failure/Error: raise 'Requests did not finish in 60 seconds' if timer.expired?
     
     RuntimeError:
       Requests did not finish in 60 seconds
     # /Users/juanca/.rvm/gems/ruby-2.3.3@foo/gems/capybara-3.15.1/lib/capybara/server.rb:64:in `wait_for_pending_requests'
     # /Users/juanca/.rvm/gems/ruby-2.3.3@foo/gems/capybara-3.15.1/lib/capybara/session.rb:131:in `reset!'
     # /Users/juanca/.rvm/gems/ruby-2.3.3@foo/gems/capybara-3.15.1/lib/capybara.rb:315:in `block in reset_sessions!'
     # /Users/juanca/.rvm/gems/ruby-2.3.3@foo/gems/capybara-3.15.1/lib/capybara.rb:315:in `reverse_each'
     # /Users/juanca/.rvm/gems/ruby-2.3.3@foo/gems/capybara-3.15.1/lib/capybara.rb:315:in `reset_sessions!'
     # /Users/juanca/.rvm/gems/ruby-2.3.3@foo/gems/capybara-3.15.1/lib/capybara/rspec.rb:18:in `block (2 levels) in <main>'
```

I only found out which request was failing by logging the increment and decrement flows in the `Capybara::Server` class. I could perhaps look through my test logs but there are so many resources requested by my application that it seems like the wrong user interface for developers. 

## Steps to reproduce

[...]

PR #2250, opened 2019-09-20
Display pending requests

juanca:
Closes: https://github.com/teamcapybara/capybara/issues/2249

# Motivation

Displays more information about pending requests to developers. It can be useful to see the URI of a pending request in the test stack trace instead of following the test logs.

# Change log and testing strategy

- Backfills tests for pending requests behavior
- Add additional information on error message

### DIFF
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -61,9 +61,7 @@ module Capybara
     def wait_for_pending_requests
       timer = Capybara::Helpers.timer(expire_in: 60)
       while pending_requests?
-        if timer.expired?
-          raise "Requests did not finish in 60 seconds: #{middleware.pending_requests}"
-        end
+        raise "Requests did not finish in 60 seconds: #{middleware.pending_requests}" if timer.expired?
 
         sleep 0.01
       end
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -61,7 +61,9 @@ module Capybara
     def wait_for_pending_requests(wait_time = 60)
       timer = Capybara::Helpers.timer(expire_in: wait_time)
       while pending_requests?
-        raise 'Requests did not finish in 60 seconds' if timer.expired?
+        if timer.expired?
+          raise "Requests did not finish in 60 seconds: #{middleware.pending_requests}"
+        end
 
         sleep 0.01
       end
--- a/lib/capybara/server.rb
+++ b/lib/capybara/server.rb
@@ -58,8 +58,8 @@ module Capybara
       false
     end
 
-    def wait_for_pending_requests(wait_time = 60)
-      timer = Capybara::Helpers.timer(expire_in: wait_time)
+    def wait_for_pending_requests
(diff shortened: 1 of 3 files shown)


lib/capybara/server/middleware.rb
```diff
@@ -7,16 +7,16 @@ class Counter
...
-        def increment
-          @mutex.synchronize { @value += 1 }
+        def increment(env)
+          @mutex.synchronize { @value.push(env) }
         end
 
-        def decrement
-          @mutex.synchronize { @value -= 1 }
+        def decrement(env)
+          @mutex.synchronize { @value.delete_at(@value.index(env)) }
```
juanca: Usually, it's good practice to provide a fallback when `@value.index(env)` does not find an element. However, since we are using `@counter` privately and in a sequential increment => decrement manner, I do not think this necessitates a fallback. 

Let me know what you think.

lib/capybara/server.rb
```diff
@@ -58,10 +58,12 @@ def responsive?
       false
     end
 
-    def wait_for_pending_requests
-      timer = Capybara::Helpers.timer(expire_in: 60)
+    def wait_for_pending_requests(wait_time = 60)
```
twalpole: Please don't parameterize this in the same PR

lib/capybara/server/middleware.rb
```diff
@@ -31,8 +31,12 @@ def initialize(app, server_errors, extra_middleware = [])
         @server_errors = server_errors
       end
 
+      def pending_requests
+        @counter.value.map { |env| env["REQUEST_URI"] }
```
twalpole: Probably needs to be in a @mutex.synchronize block

lib/capybara/server/middleware.rb
```diff
@@ -31,8 +31,12 @@ def initialize(app, server_errors, extra_middleware = [])
...
+      end
+
       def pending_requests?
-        @counter.value.positive?
+        @counter.value.length.positive?
```
twalpole: Needs to be in a @mutex.synchronize block

lib/capybara/server/middleware.rb
twalpole: Yes please -- `@value.delete_at(@value.index(env) || @value.length)` just for safety

lib/capybara/server/middleware.rb
```diff
@@ -43,14 +47,14 @@ def call(env)
         if env['PATH_INFO'] == '/__identify__'
           [200, {}, [@app.object_id.to_s]]
         else
-          @counter.increment
+          @counter.increment(env)
```
twalpole: Can we just push the URI -- pushing the whole env seems wasteful and the server can modify the env which means it wouldn't match when decrementing

juanca:
Revisions submitted. Thanks for the quick review!

twalpole:
Thanks for this -- It'll probably take me a couple of days to get it merged with my current work schedule.

PR #2252, opened 2019-10-03
Remove spec files from final gem

stefansundin:
This reduces the compiled gem by 40 kb.

The main reason I want to do this is because I have started scanning my docker image with a vulnerability scanner, and it is complaining that there's a private key in there, which is coming from capybara (`spec/fixtures/key.pem`).

Disclaimer: I have not done any testing other than building the gem after the change.

Thanks!


twalpole:
These files are specifically included for use by Capybara supporting drivers in their test suites.

PR #2253, opened 2019-10-03
Fix raise() syntax

alepore:
The `ArgumentError()` syntax is wrong and generates a pretty weird error:

NoMethodError: undefined method `ArgumentError' for #<Capybara::Session>


twalpole:
Thanks!

PR #2267, opened 2019-10-31
Specify GitHub repository instead of the path

y-yagi:
I fixed this because I want to work a test against ruby-head. 
https://travis-ci.org/teamcapybara/capybara/jobs/605383442 

But if the purpose of this Gemfile is to check the test locally, feel free to close this. Thanks.

### DIFF
--- a/gemfiles/Gemfile.edge-firefox
+++ b/gemfiles/Gemfile.edge-firefox
@@ -5,7 +5,7 @@ gemspec path: '..'
 
 gem 'xpath', github: 'teamcapybara/xpath'
 
-gem 'selenium-webdriver', :path => '../../selenium/build/rb'
+gem 'selenium-webdriver', github: 'seleniumhq/selenium', glob: 'rb/*.gemspec'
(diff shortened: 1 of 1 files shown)


twalpole:
The purpose of this was to use the edge version of selenium-webdriver during the travis tests. Selenium-webdriver requires (or used to require) to be built in order for the JS atoms to be available, and couldn't be used directly from the repo. I did some work on changing that in the selenium repo a while back but haven't verified whether they've been kept up to date in a couple of months (I will have some time tomorrow/this weekend).  Since it required to be built, the package was downloaded on travis, then built, and then used locally.

y-yagi:
Thanks for your explanation! I misunderstood the purpose of that build.
I will close this. By the way, how do you feel add build that ruby-head with default Gemfile?

y-yagi:
>  ruby-head with default Gemfile?

This fixed by 1410793107e2be4d55db86c627eb68c20e81eb6b. Thansk!

PR #2270, opened 2019-11-14
Preserve clientX/clientY in HTML5 drag and drop events

nickgnd:
Hey Capybara team :wave:

First of all, thanks for the amazing work that you have done so far, we use `Capybara` on our daily job and rely on it for building and testing our web applications :bow:

---

### Context:

HTML5 Drag and Drop.

### Issue:

When using HTML5 drag and drop to drag an object onto a target, the object is always dropped in the top left corner of the window (x: 0, y: 0) instead of the previously computed position within the target.

Background: HTML5 drag and drop is currently implemented via simulating a series of HTML5 events with pre-computed values for the mouse pointer position `clientX` and `clientY`. For the last events in this series (`dragleave`, `drop`, `dragend`), the previously computed position is lost on the way. Instead of carrying it along, the events use the default options without a `clientX` and `clientY` property, which makes them default to 0.

### Expected Behavior:

The object should be dropped in the center of the target.

For the `dragleave`, `drop` and `dragend` events the mousepointer position (`clientX`, `clientY`) should no longer change. They should be the same as the last `dragover` event.

### Further details

This PR fixes this by changing how HTML5 drag and drop is mimicked. We propagate down the previously computed `clientX` and `clientY` coordinates from the last `dragover` event (initially meant as dropping position for the object) to the following events: `dragleave`, `drop` and `dragend`.

Here is a screenshot from the changed spec to give you an idea of the series of events and their mouse position. 

screen

---

We are looking forward to hearing from you, have a great day!

### DIFF
--- a/lib/capybara/selenium/extensions/html5_drag.rb
+++ b/lib/capybara/selenium/extensions/html5_drag.rb
@@ -166,17 +166,18 @@ class Capybara::Selenium::Node
         var dragOverOpts = Object.assign({clientX: targetCenter.x, clientY: targetCenter.y}, opts);
         var dragOverEvent = new DragEvent('dragover', dragOverOpts);
         target.dispatchEvent(dragOverEvent);
-        window.setTimeout(dragLeave, step_delay, dragOverEvent.defaultPrevented);
+        window.setTimeout(dragLeave, step_delay, dragOverEvent.defaultPrevented, dragOverOpts);
       }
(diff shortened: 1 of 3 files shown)


twalpole:
Seems reasonable to me -- Thanks for the PR!

Its been a while since a release, so I'll probably include this in the next release this weekend

Issue #1733, opened 2016-08-03
Last matcher


thehenster:
This idea has been [discussed before](https://github.com/jnicklas/capybara/pull/1101) but I thought I'd float a less in your face version. Something like..

```
click_link "Save", match: :last
```
or
```
fill_in "Title", match: :last
```
or
```
find(".item", match: :last)
```

If that still increases the size of the API more than the feature is worth it then no problem.

twalpole:
I think the `click_link` and `fill_in` examples seem like they would be really rare cases and probably better served by just scoping to a region of the page.  Most cases of the `find` example are probably better served by doing something along the lines of `find("div.item:last-of-type")` which then doesn't have the downsides of #all/#first

PR #2275, opened 2019-11-21
Make it so passing match: first to ancestor() finds closest ancestor

TylerRick:
Usually, when looking for an ancestor, you're interested in the _closest_ ancestor, not the most distantly related ancestor.

And when one speaks of the "_first_ ancestor", one always means the _closest_ ancestor, as you traverse parent elements upwards towards the root of the document (see [`Element.closest`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)).

This simple change makes it easier to find the closest (first) ancestor.

## Expected Behavior

```
  it 'with match: first should find closest ancestor' do
    expect(el.ancestor(:css, 'div', match: :first)[:id]).to eq 'ancestor1'
    expect(el.ancestor(:css, 'div', match: :first, text: 'Ancestor')[:id]).to eq 'ancestor1'
  end
```

## Actual Behavior

```
  1) Capybara::Session RackTest #ancestor with match: first should find closest ancestor
     Failure/Error: expect(el.ancestor(:css, 'div', match: :first)[:id]).to eq 'ancestor1'
     
       expected: "ancestor1"
            got: "ancestor3"
     
       (compared using ==)
     # ./lib/capybara/spec/session/ancestor_spec.rb:31:in `block (2 levels) in <top (required)>'
```

## `mach: last`?

Adding a `match: :last` option (#1733) could also be an option, but I think in this case, it just makes more sense for `:first` to mean closest.

### DIFF
--- a/lib/capybara/queries/ancestor_query.rb
+++ b/lib/capybara/queries/ancestor_query.rb
@@ -12,6 +12,7 @@ module Capybara
           ancestors = node.find_xpath(XPath.ancestor.to_s)
                           .map(&method(:to_element))
                           .select { |el| match_results.include?(el) }
+                          .reverse
           Capybara::Result.new(ancestors, self)
         end
       end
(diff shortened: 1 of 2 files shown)


twalpole:
Concern here is that this is a breaking change

TylerRick:
True, it is. I guess we'd have to wait until the next major version then?

TylerRick:
Unless we added a configuration option to control the order :-1: :grimacing:

twalpole:
There is now an :order option which can be specified as :reverse - that combined with match: :first will give the closest matching ancestor.

PR #2277, opened 2019-11-21
Add `closest` and `ancestors` finders

### LINKED ISSUE #2276: Provide a `closest` finder (or change `ancestor`) to return closest ancestor
I'd like to propose adding a `closest` finder, which would be to `ancestor` what `first` is to `all`.

## Motivation

[This question](https://stackoverflow.com/questions/7017530/closest-in-capybara) asks how to find the closest matching ancestor.

That question is...
- ... the top search result when searching for _capybara find closest element_  (using DuckDuckGo)
- ... the 3rd search result when searching for _capybara find closest ancestor_

The way to do that with both the [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) and the [jQuery API](https://api.jquery.com/closest/) is with `closest`. It's a useful search to be able to do. So it would be natural to expect Capybara to at least _have_ an equivalent finder. And one could even _hope_ that it would be named the same too. :) (Though I realize — and am glad — that the equivalent to `querySelector` and `querySelectorAll` in Capybara is `first` and `all` rather than those long names, so name parity isn't _everything_. But it is _something_.)

Alas, my search for such a method in Capybara has been in vain so far.

I keep running into this same question/need myself (and I'm obviously not the only one). In fact, I was surprised when I found this question and then found my own answer explaining that `.ancestor` is currently the best way to do this with the Capybara API currently.

But existing `ancestor` is not semantically the same as [`closest`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest): `closest` automatically returns the single, _closest_ match, whereas our a `ancestor` method expects to only ever find *one* match or it raises an error. I guess that works okay if you can be very specific with your query (locating by `#id`, for example), but that seems less useful than `closest` (which you could probably replace most `ancestor` queries with and your tests would still pass).

## Expected Behavior

```
  it 'should find closest ancestor (even if there are multiple matches)' do
    expect(el.closest(:css, 'div')[:id]).to eq 'ancestor1'
    expect(el.closest(:css, 'div', text: 'Ancestor')[:id]).to eq 'ancestor1'
    expect(el.all_ancestors(:css, 'div').map {|x| x[:id] }).to eq ['ancestor1', 'ancestor2', 'ancestor3']
  end
```

# Implementation

Assuming it's okay to change the order of elements returned by `AncestorQuery` so that `match: :first` works as expected (#2275), this weould be easy to build as a wrapper for `ancestor(match: :first`).

But if #2275 can't be released yet because it's a breaking change, we could implement `closest` without introducing that breaking change yet...

# Meta (not relevant)
Capybara Version: capybara-3.29.0
Driver Information (and browser if relevant): selenium-webdriver-3.141.0

TylerRick:
Fixes #2276

Depends on #2275 (breaking change) but could be rewritten to not depend on that change

### DIFF
--- a/lib/capybara/node/finders.rb
+++ b/lib/capybara/node/finders.rb
@@ -77,6 +77,32 @@ module Capybara
 
       ##
       #
+      # Find the first {Capybara::Node::Element} based on the given arguments that is also an
+      # ancestor of the element called on.
+      #
+      # {#closest} takes the same options as {#find}.
+      #
+      #     element.closest('#foo').find('.bar')
+      #     element.closest(:xpath, './/div[contains(., "bar")]')
+      #     element.closest('ul', text: 'Quox').click_link('Delete')
+      #
+      # @param (see #find)
+      #
+      # @macro waiting_behavior
+      #
+      # @return [Capybara::Node::Element]      The found element
+      # @raise  [Capybara::ElementNotFound]    If the element can't be found before time expires
+      #
+      # TODO: implement it more like first? don't use synced_resolve since it could raise Ambiguous?
+      def closest(*args, **options, &optional_filter_block)
+        options = { minimum: 1 }.merge(options) unless options_include_minimum?(options)
+        options = { match: :first }.merge(options)
+        options[:session_options] = session_options
+        synced_resolve Capybara::Queries::AncestorQuery.new(*args, options, &optional_filter_block)
+      end
+
+      ##
+      #
       # Find an {Capybara::Node::Element} based on the given arguments that is also a sibling of the element called on.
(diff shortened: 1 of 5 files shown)


twalpole:
I don't like the name `closest` because it doesn't really tell you what it's looking for - `closest_ancestor` maybe. I'm also wondering what the use case for `ancestors` would ever be.  I'm trying to think of a case when testing an app/site that I would ever want to find all the ancestors of an element and can't really think of one.

twalpole:
I'm closing this in favor of the added :order option for now. That should allow anyone who wants to easily define their own closest matching ancestor helper.  I'm not yet convinced there is a real use case for ancestors.

PR #2294, opened 2019-12-31
Support held modifier keys in drag & drop

elliotcm:
This allows simulation of holding the `alt` key for copying items during drag & drop.

Example usage:
```
meeting.drag_to(calendar, modifier_keys: [:alt])
```

### DIFF
--- a/lib/capybara/selenium/extensions/html5_drag.rb
+++ b/lib/capybara/selenium/extensions/html5_drag.rb
@@ -4,12 +4,12 @@ class Capybara::Selenium::Node
   module Html5Drag
     # Implement methods to emulate HTML5 drag and drop
 
-    def drag_to(element, html5: nil, delay: 0.05)
+    def drag_to(element, html5: nil, delay: 0.05, modifier_keys: [])
       driver.execute_script MOUSEDOWN_TRACKER
       scroll_if_needed { browser_action.click_and_hold(native).perform }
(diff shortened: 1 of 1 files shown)


twalpole:
Please write a test for this - also the title states it's implemented in Chrome but I don't see any reason it shouldn't work in the other browsers that include the Html5Drag behavior (firefox and chrome based edge).  Please also document the valid values for the modifier keys.

twalpole:
To make this more correct you should probably also be modifying `browser_action.click_and_hold(native).perform` 
to call `key_down` for each modifier key. You also need to handle the translation between :ctrl and :control in different contexts.

elliotcm:
@twalpole I'll take a look when time permits; this was a result of unblocking work on a project.

Could you point me towards the existing tests for this part of the code?

twalpole:
Current `drag_to` specs are at https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/session/node_spec.rb#L422

elliotcm:
Cheers

elliotcm:
@twalpole I can't figure out how to run a single spec file, all the guidance and tasks seem to be guided towards running the whole suite.

twalpole:
Tag the tests you want to run with `:focus_` then run with `bundle exec rake spec_chrome`

elliotcm:
Closed in favour of https://github.com/teamcapybara/capybara/pull/2295

PR #2298, opened 2020-01-21
Set range inputs via JS in selenium driver

pixeltrix:
The slider can't be controlled so set it via JS like color inputs.


pixeltrix:
I hope I've put the test in the right place 🤞

pixeltrix:
Test failures seem unrelated?

twalpole:
LGTM - thanks!

PR #2304, opened 2020-02-05
Avoid booting errors when using selenium_headless_chrome in docker

paul-mesnilgrente:
This line solves this problem:

https://stackoverflow.com/questions/50642308/webdriverexception-unknown-error-devtoolsactiveport-file-doesnt-exist-while-t

### DIFF
--- a/lib/capybara/registrations/drivers.rb
+++ b/lib/capybara/registrations/drivers.rb
@@ -29,6 +29,9 @@ Capybara.register_driver :selenium_chrome_headless do |app|
   browser_options = ::Selenium::WebDriver::Chrome::Options.new.tap do |opts|
     opts.args << '--headless'
     opts.args << '--disable-gpu' if Gem.win_platform?
+    if RbConfig::CONFIG['host_os'] =~ /linux/ && system('grep -q docker /proc/self/cgroup')
+      opts.args << '--disable-dev-shm-usage'
+    end
     # Workaround https://bugs.chromium.org/p/chromedriver/issues/detail?id=2650&q=load&sort=-id&colspec=ID%20Status%20Pri%20Owner%20Summary
     opts.args << '--disable-site-isolation-trials'
   end


twalpole:
This setting is undesirable when not running in a containerized environment for performance reasons, and therefore will be not accepted as default behavior.  I would accept a PR that only adds that if it detects it's running in a docker instance, or adds the info to the Gotchas section of the README

paul-mesnilgrente:
👍 I'll work on this and push

paul-mesnilgrente:
What do you think about this version?

twalpole:
How reliable is `grep -q docker /proc/self/cgroup` for detecting being in a docker container?

paul-mesnilgrente:
It is the solution provided in the gem [os](https://github.com/rdp/os). A part of the solution is also referenced on stackoverflow [here](https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker) (most active thread on stackoverflow when looking for "detect docker runtime").

There is also a reference to the cgroup on Docker's documentation [here](https://docs.docker.com/config/containers/runmetrics/#find-the-cgroup-for-a-given-container).

paul-mesnilgrente:
What do you guys think? I am happy to try improving it if it doesn't look good.

twalpole:
Paul, I'm having second thoughts about adding more environment specific default configurations. If we add for docker should we also be adding for other things, and how much more work maintenance of different environments would add.   Still thinking about it.

GarrisonD:
@paul-mesnilgrente I am using the latest for today `capybara (3.33.0)`, `selenium-webdriver (3.142.7)` and official `selenium/standalone-chrome-debug:3.141` container . Everything works perfectly without any tweaks. Did you set up `shm_size` to `2GB` (1st option) or mount `/dev/shm` (2nd option) as being described in README for official selenium containers? See https://github.com/SeleniumHQ/docker-selenium#running-the-images.

paul-mesnilgrente:
@GarrisonD . This problem is to address the use of the pre-registered driver configuration `:selenium_chrome_headless`. You are using a remote browser which is a different configuration.

However, if you know how to increase the `shm_size` or mount `/dev/shm` in gitlab-ci, I'll take that information :)

GarrisonD:
@paul-mesnilgrente Well, that's exactly what I was doing yesterday for my self-hoster GitLab  :)

You need to edit the config of GitLab runner(s). You can find info about both `volumes` / `shm_size` here: https://docs.gitlab.com/runner/configuration/advanced-configuration.html

paul-mesnilgrente:
Are you using the helm chart? I've seen that docs but I can't figure out how to override the `config.toml` in the chart's config

GarrisonD:
No, I don't use it but I googled and look like you can configure GitLab runner with Heml Chart the same way as with `config.toml` but with just a bit different format (`.yml` instead of `.toml`). https://docs.gitlab.com/runner/install/kubernetes.html

GarrisonD:
Well, I googled a bit more and looks like you can't configure it with Heml Chart... Happy hacking then 😆

GarrisonD:
Well, I googled a bit more... You don't have to specify it in Helm Chart config... You use Helm Chart config only to set up the Kubernetes cluster. For configuring `shared volumes` / `shm_size` / etc. you have to use Kubernetes executor (https://docs.gitlab.com/runner/executors/kubernetes.html) with `config.toml`.

Issue #2311, opened 2020-02-17
GET requests on forms do not overwrite URL parameters


faissalMT:
**Note:** This is for issues with Capybara.  If you have a howto type question, please ask on the mailing list as requested in the README: http://groups.google.com/group/ruby-capybara

## Meta
Capybara Version:
3.29.0
Driver Information (and browser if relevant):
`rack_test`

## Expected Behavior
The query params in the form target URL should be replaced. This is what both Chrome and Firefox do.  

## Actual Behavior
The query params in the form target URL are appended to. The only browser I can find that behaves this way is Links2.

## Steps to reproduce
In a view:
```
<%= form_with(url: "https://www.google.com/search?source=test", method: "get") do %>
  <%= f.text_field(:q, id: "query") %>
<% end %>
```
In a test for said page:
```
page.find("#query").fill_in("my search query")
```

twalpole:
I believe the current behavior was correct for HTML4 - but isn't for HTML5

twalpole:
Fixed via PR #2312

PR #2312, opened 2020-02-17
Rack form submit - Fix Issue #2311

twalpole:


lib/capybara/rack_test/browser.rb
```diff
@@ -30,7 +30,9 @@ def refresh
 
   def submit(method, path, attributes)
     path = request_path if path.nil? || path.empty?
-    process_and_follow_redirects(method, path, attributes, 'HTTP_REFERER' => current_url)
+    uri = build_uri(path)
+    uri.query = '' if method&.to_s&.downcase == 'get'
```
luke-hill: not sure if you're doing rubocop / care, but `casecmp('get').zero?` is more optimal here

twalpole: Yeah, but it's also much less readable.  If I created a `get?` method I might go with that as the implementation but not for inline.

PR #2327, opened 2020-03-29
Add support for ARIA button role attribute

hanachin:
This adds support for matching on the role=button when finding button element.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -81,6 +81,7 @@ module Capybara
     # - **default_selector** (`:css`, `:xpath` = `:css`) - Methods which take a selector use the given type by default. See also {Capybara::Selector}.
     # - **default_set_options** (Hash = `{}`) - The default options passed to {Capybara::Node::Element#set Element#set}.
     # - **enable_aria_label** (Boolean = `false`) - Whether fields, links, and buttons will match against `aria-label` attribute.
+    # - **enable_aria_button_role** (Boolean = `false`) - Buttons will match against element that has `role="button"` attribute.
     # - **exact** (Boolean = `false`) - Whether locators are matched exactly or with substrings. Only affects selector conditions
     #   written using the `XPath#is` method.
     # - **exact_text** (Boolean = `false`) - Whether the text matchers and `:text` filter match exactly or on substrings.
@@ -501,6 +502,7 @@ Capybara.configure do |config|
   config.visible_text_only = false
   config.automatic_label_click = false
   config.enable_aria_label = false
+  config.enable_aria_button_role = false
   config.reuse_server = true
   config.default_set_options = {}
   config.test_id = nil
--- a/lib/capybara/selector/definition/button.rb
+++ b/lib/capybara/selector/definition/button.rb
@@ -4,6 +4,7 @@ Capybara.add_selector(:button, locator_type: [String, Symbol]) do
   xpath(:value, :title, :type, :name) do |locator, **options|
     input_btn_xpath = XPath.descendant(:input)[XPath.attr(:type).one_of('submit', 'reset', 'image', 'button')]
     btn_xpath = XPath.descendant(:button)
+    btn_xpath = btn_xpath.union(XPath.descendant[XPath.attr(:role).equals('button')]) if enable_aria_button_role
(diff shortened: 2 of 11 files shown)


twalpole:
Thanks for this. Would you have time to check if there are any other ARIA roles that make sense to support in Capybaras built-in selectors rather than just a single one.  Also please make this optional, like we have with aria label support, so users don't have to pay the performance penalty if they don't need to.

lib/capybara/spec/session/click_link_or_button_spec.rb
```diff
@@ -19,6 +19,13 @@
     expect(extract_results(@session)['first_name']).to eq('John')
   end
 
+  it 'should click on a button role', requires: [:js] do
+    Capybara.default_max_wait_time = 2
+    @session.visit('/with_js')
+    @session.click_link_or_button('ARIA button')
+    @session.click_link_or_button('ARIA button has been clicked')
```
twalpole: There should be some type of expectation here

lib/capybara/spec/session/click_link_or_button_spec.rb
```diff
@@ -19,6 +19,13 @@
     expect(extract_results(@session)['first_name']).to eq('John')
   end
 
+  it 'should click on a button role', requires: [:js] do
+    Capybara.default_max_wait_time = 2
+    @session.visit('/with_js')
+    @session.click_link_or_button('ARIA button')
```
twalpole: Since you're only testing the button selector this should be `click_button`

lib/capybara/selector/definition/button.rb
```diff
@@ -5,6 +5,7 @@
     input_btn_xpath = XPath.descendant(:input)[XPath.attr(:type).one_of('submit', 'reset', 'image', 'button')]
     btn_xpath = XPath.descendant(:button)
     image_btn_xpath = XPath.descendant(:input)[XPath.attr(:type) == 'image']
+    aria_btn_xpath = XPath.descendant[XPath.attr(:role).equals('button')]
```
twalpole: Any reason this can't just be part of `btn_xpath`? -  `btn_xpath = XPath.descendant(:button).union XPath.descendant[XPath.attr(:role).equals('button')]`

lib/capybara.rb
```diff
@@ -81,6 +81,7 @@ class << self
     # - **default_selector** (`:css`, `:xpath` = `:css`) - Methods which take a selector use the given type by default. See also {Capybara::Selector}.
     # - **default_set_options** (Hash = `{}`) - The default options passed to {Capybara::Node::Element#set Element#set}.
     # - **enable_aria_label** (Boolean = `false`) - Whether fields, links, and buttons will match against `aria-label` attribute.
+    # - **enable_aria_button_role** (Boolean = `false`) - Buttons will match against element that has `role="button"` attribute.
```
hanachin: I added a new option `enable_aria_button_role`

twalpole:
@hanachin Did you go through all the defined aria roles to see if there are any more that make sense to add to Capybaras builtin selectors?

hanachin:
@twalpole There are some roles that make sense to add to Capybara builtin selectors.

- Add `role=checkbox` to `:checkbox` selector
- Add `role=switch` to `:checkbox` or `:radio_button` selector
- Add `role=radio` sing_the_radio_role) to `:radio_button` selector
- Add `role=link` sing_the_link_role) to `:link` selector
- Add `role=listbox` to `:select` selector
- Add `role=option` to `:option` selector
- Add `role=menuitemcheckbox` to `:checkbox` selector
- Add `role=menuitemradio` to `:radio_button` selector
- Add `role=gridcell` to `:button` selector, I'm not sure.
- Add `role=menuitem` to `:button` or `:link` selector, I'm not sure.

There is working ARIA examples.

- `role=switch`: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role
- other roles: https://www.w3.org/TR/wai-aria-practices/examples/

twalpole:
I've added support for the `link` role too - PRs for other roles would be appreciated although some of them may become complicated to implement (checkbox, etc)

PR #2345, opened 2020-05-04
Fix have_text description

juampi:
All other matchers have descriptions that begin with the word `have`, like `have title` or `have current path`, so doing stuff like `it { is_expected.to have_title("Hello World") }`
gives `is expected to have title "Hello World"`
but currently, `it { is_expected.to have_text("Hello World") }`
gives `is expected to text "Hello World"` which seems wrong.

This PR fixes that.


twalpole:
Thank you

PR #2346, opened 2020-05-08
Fixes ruby 2.7 deprecation warning when registering webrick

97jaz:
I'm not sure what (if any) test would be appropriate for this change (I looked at the earlier batch of 2.7 deprecation changes, and nothing stood out), but I'm happy to oblige.


twalpole:
Thank you for this

PR #2352, opened 2020-05-19
Fix grammar in Capybara::Node::Finders documentation

seanpcoyle:


lib/capybara/node/finders.rb
```diff
@@ -40,7 +40,7 @@ module Finders
       #
       #                                                class:['a', '!b', '!!!c'] # limit to elements with class 'a' and '!c' but not class 'b'
       #
-      #   @option options [String, Regexp, Hash] style  Only find elements with matching style. String and Regexp will be checked against text of the elements `style` attribute, while a Hash will be compared against the elements full style
+      #   @option options [String, Regexp, Hash] style  Only find elements with matching style. String and Regexp will be checked against text of the element's `style` attribute, while a Hash will be compared against the elements full style
```
twalpole, line 43: If we're going to be pedantic about grammar the final `elements ` in here should have an apostrophe too

PR #2359, opened 2020-05-29
Fix #set_text behavior with clear: :none option

sukechannnn:
## Overview

After 696f9d3 are committed, Capybara has been overwrite form input even if `clear: :none` option is provided.
This bug occur with rapid fill when input over 30 words.

This PR fixes that.


twalpole:
Thank you for this - I've pulled it into to PR #2361 and made a couple tweaks -- will merge it from there

PR #2365, opened 2020-06-15
Fix Ruby 2.7 warning in minitest assert_text

eileencodes:
In our application at GitHub we were seeing a warning coming from
`assert_text` on the latest version and building the source from master.

Warning we were seeing:

```
/vendor/gems/2.7.1/ruby/2.7.0/gems/capybara-3.32.2/lib/capybara/minitest.rb:56: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/vendor/gems/2.7.1/ruby/2.7.0/gems/capybara-3.32.2/lib/capybara/minitest.rb:56:in `assert_text'
```

The warning is only visible when `assert_text` is delegated from
minitest which is why I've updated the minitest_spec.rb file to include
an argument. I was unable to see or reproduce this warning in the assert
text test.

I don't really get why `ruby2_keywords` wasn't adequate for this change
but from my testing my changes work in Ruby 2.6 and 2.7 and fix the
warning in 2.7.

### DIFF
--- a/lib/capybara/minitest.rb
+++ b/lib/capybara/minitest.rb
@@ -50,15 +50,14 @@ module Capybara
 
       %w[text no_text title no_title current_path no_current_path].each do |assertion_name|
         class_eval <<-ASSERTION, __FILE__, __LINE__ + 1
-          def assert_#{assertion_name} *args
+          def assert_#{assertion_name}(*args, **kwargs)
             self.assertions +=1
             subject, args = determine_subject(args)
-            subject.assert_#{assertion_name}(*args)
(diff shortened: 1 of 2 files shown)


twalpole:
Thanks for this.  I don't understand why `ruby2_keywords` wasn't adequate either, but then the whole deprecation of keyword arguments behavior has been handled terribly by Ruby (imho) so i don't find it surprising.

eileencodes:
Is this something that I can help backport for the 3.32 series? Let me know if I can help at all with that, or if this will be in a future version 😄

PR #2367, opened 2020-06-29
Ignore failure when IE fails to perform `reset_sessions!`

luke-hill:
From SHA:

IE Does not seem to expose the endpoint for clearing local storage. So when trying to reset a session (Default behaviour in capybara/cucumber), we should just ignore this missing endpoint - As the session has still been 'reset'

This is a default hook added, and making this small tweak seems to have the desired effect of not making IE fail with an untraceable hook.

PS: I added the file to gitignore as it kept adding it to my changed files.

### DIFF
--- a/lib/capybara/selenium/driver_specializations/internet_explorer_driver.rb
+++ b/lib/capybara/selenium/driver_specializations/internet_explorer_driver.rb
@@ -15,6 +15,13 @@ module Capybara::Selenium::Driver::InternetExplorerDriver
 
 private
 
+  def clear_storage
+    clear_session_storage unless !!options[:clear_session_storage]
+    clear_local_storage unless !!options[:clear_local_storage]
+  rescue Selenium::WebDriver::Error::JavascriptError
+    # session/local storage may not be available if on non-http pages (e.g. about:blank)
+  end
+
   def build_node(native_node, initial_cache = {})
     ::Capybara::Selenium::IENode.new(self, native_node, initial_cache)
   end


lib/capybara/selenium/driver.rb
```diff
@@ -302,7 +302,11 @@ def clear_browser_state
   end
 
   def clear_browser_state_errors
-    @clear_browser_state_errors ||= [Selenium::WebDriver::Error::UnknownError]
+    @clear_browser_state_errors ||=
```
twalpole, line 305: If this needs changing for IE it should be done in the IE specialization, not in the general Selenium driver

.gitignore
```diff
@@ -11,6 +11,7 @@ spec.opts
 .ruby-version
 .java-version
 .ruby-gemset
+.rakeTasks
```
twalpole, line 14: Any idea what is creating a `.rakeTasks` file and/or what's in it - I have no issues with that when running tests.

twalpole:
What method exactly is IE not responding to? The base Selenium driver already has separate methods for clear_session_storage and clear_local_storage that will only call the Selenium methods if they exist, and otherwise implement them as JS.  This really seems like more of a Selenium configuration issue, or need to specialization of clear_local_storage/clear_session_storage in the driver specialization rather than just ignoring an error.

.gitignore
luke-hill: I can remove it if you want.

lib/capybara/selenium/driver.rb
luke-hill: This is the code that was being called (So I'm guessing there's a super call somewhere, I'll track it down).

luke-hill:
As mentioned in PR - clear_storage

```
clear_storage
=> Command not found: DELETE /session/f919e46621f35293c8cb008b743dde538d34f13a/session_storage
```

luke-hill:
Digging into this further, if we slightly refactor the conditional, that would work. I'll push up the changes to see if that's ok with you

twalpole:
No the changes as written are not acceptable  - as I stated in my previous review, if this is going to be handled in Capybara this should be handled in the IE driver specialization not in the general selenium driver. Also changing to check for nil or file changes the default for everything which is definitely not acceptable.  

The root cause of this issue is that Selenium should not be responding to `session_storage` and/or "local_storage' (which means not including ::Selenium::WebDriver::DriverExtensions::HasWebStorage) if IE doesn't support it.

luke-hill:
Just a default IE driver hitting browserstack with version as "11.0"

twalpole:
Again, how exactly are you configuring your driver? Is it configured as remote? (I’m guessing yes because you mention browserstack) Is this a case where locally IE would work, but it’s failing remotely because browserstack hasn’t implemented an endpoint?

twalpole:
This still isn't acceptable - you're changing the default behavior of the IE driver to not clear storage but I'm 99% sure clearing storage works fine for a local configuration, and your issue is browserstack specific (or possibly remote specific)

luke-hill:
I don't know what more to do sorry, it's pretty messy in there and I've been out of the game for too long to remember where to look. If you can maybe offer some advice that would be greatly appreciated.

twalpole:
IF we're going to patch this in Capybara - it would be capturing and ignoring the error in the IE specialization for the individual local and session clearing methods. I say IF, because I'm not fully sure where the real issue here is coming from.  Capybara is calling methods that have been included in the driver that Selenium returns.  Those methods shouldn't be there if the browser being used doesn't support those endpoints. So this could actually be an issue with Selenium, it could be an issue with the versions of client vs grid you're using, or it could be an issue that BrowserStack isn't implementing an endpoint they're supposed to have.  It's hard to know because you seem reticent to show your code that actually configures your driver so I can see exactly what you're doing, or answering the specific questions I ask about that.

luke-hill:
It's not that I'm reticent to do it, it's that I'm changing the caps constantly, to try and see if there's a way I can avoid it (A bit like the chrome issue I raised). I'm aware I'm late to the party, and as such maybe using 3.x is considered old.

the latest caps I'm using is

```
Selenium::WebDriver::Remote::Capabilities.new.tap do |capabilities|
        capabilities["browserName"] = browser.capitalize
        capabilities["browserName"] = "IE" if browser == :internet_explorer
        capabilities["browserVersion"] = browser_version
        capabilities["bstack:options"] = {
          "os" => "Windows",
          "osVersion" => "10",
          "local" => "false",
          "seleniumVersion" => selenium_jar_version,
          "debug" => "false",
          "consoleLogs" => "verbose",
          "networkLogs" => "false",
        }

        if browser == :chrome
          capabilities["goog:loggingPrefs"] = { browser: "ALL", driver: "ALL" }
        end
```

twalpole:
@luke-hill I've now asked you multiple times how you're configuring your driver and you've ignored questions or given bits and pieces, but never a complete answer - Could you just paste your entire `Capybara.register_driver` statement?????

luke-hill:
```
      Capybara.register_driver :selenium do |app|
        Capybara::Selenium::Driver.new(
          app,
          browser: :remote,
          desired_capabilities: browser_capabilities, (As mentioned above).
          url: hub_url
        )
      end
```

twalpole:
So from comments on some of the selenium threads, it sounds like browser stack was using the wrong version of the IE driver, and this is nothing to do with Capybara, correct?

luke-hill:
I haven't got around to re-testing this. Possibly? If you give me a small amount of time I might be able to triage further. Trying to make heads and tails of it all.

luke-hill:
Update: This has a workaround in place now which makes it work. It seems to be a JWP/W3C mismatch with Browserstack

Closing issue.

Issue #2368, opened 2020-07-13
capybara rack/test browser session erasing fragment params


HoneyryderChuck:
## Meta
Capybara Version:
3.32.2?

## Expected Behavior

Coming from [here](https://github.com/rack/rack/issues/1684). After some discussion with the rack maintainers, everyone's in agreement this is a capybara issue, mostly a rack test browser session one.

After submitting an HTML form, if a server responds with a 302, capybara takes care of following the redirects, as browsers do. 

However, if the server responds with a URI containing a fragment, capybara "cleans" the fragment when following the redirect. This is because [Capybara uses the rack-test methods for performing a get request](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/rack_test/browser.rb#L54), which in turn end up calling `Rack::MockRequest.env_for`. This will generate a rack request env, which will not take fragments into account.

After a talk with the rack maintainers, we agree that, if capybara plays the role of browser, it should be capybara's concern to propagate the fragment, in this case.

What do you think?

twalpole:
Please try the `rack_fragment` branch and see if that does what you expect.

twalpole:
Please provide a self contained example showing your issue

HoneyryderChuck:
```ruby
require "roda"
require "capybara"

class App < Roda
  route do |r|
    r.root { "This is #{request.url}" }
    r.on "fragment" do
      r.redirect "/#bang=bang"
    end
  end
end
App.plugin :common_logger 

session = Capybara::Session.new(:rack_test, App)
session.visit "/fragment"
puts session.current_url
# right now, it's http://www.example.com/
```

twalpole:
Closed via #2369

HoneyryderChuck:
it worked. thx @twalpole !

PR #2369, opened 2020-07-14
Maintain URI fragment when redirecting in the rack test driver

twalpole:
Fix for issue #2368

### DIFF
--- a/lib/capybara/rack_test/browser.rb
+++ b/lib/capybara/rack_test/browser.rb
@@ -8,6 +8,7 @@ class Capybara::RackTest::Browser
 
   def initialize(driver)
     @driver = driver
+    @current_fragment = nil
   end
 
   def app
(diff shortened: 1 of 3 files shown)


HoneyryderChuck:
@twalpole thx for taking the time to fix this.

It's still not working however, at least for rack-test adapter. I see that you're trying to get the current_fragment at the top, before you process redirects, however I don't think that's the right approach. The right one should be smth like: if last response is a redirect, get the fragment from the "Location" URL, process the redirect, and append the fragment to the last response url.

Let me know if this works.

twalpole:
@HoneyryderChuck Yes - this currently fixes for the use case where you visit a url with a fragment, and that redirects to a url without a fragment (browser maintains the fragment). Sounds like your use case is different - as requested on your issue please provide a self contained example showing your use case and the issue.

PR #2373, opened 2020-07-28
Add support for custom message to `assert_selector`

joelhawksley:
### Problem

As a developer writing tests, the built-in error messages
for assert_selector can sometimes be confusing/not clear.

### Solution
Allow for custom error messages for `assert_selector`.

This PR is just an idea. I could see us adding this option
to all the assertion helpers in Capybara, but wanted to get
feedback on the idea first :heart:. I'd be happy to implement
this feature throughout the library if desired.

### DIFF
--- a/lib/capybara/node/matchers.rb
+++ b/lib/capybara/node/matchers.rb
@@ -104,6 +104,7 @@ module Capybara
       #
       # @param (see Capybara::Node::Finders#all)
       # @option options [Integer] :count (nil)    Number of times the expression should occur
+      # @option options [String] :message (nil)   Custom message to display if assertion fails
       # @raise [Capybara::ExpectationNotMet]      If the selector does not exist
       #
       def assert_selector(*args, &optional_filter_block)
--- a/lib/capybara/queries/selector_query.rb
+++ b/lib/capybara/queries/selector_query.rb
@@ -9,7 +9,7 @@ module Capybara
 
       SPATIAL_KEYS = %i[above below left_of right_of near].freeze
       VALID_KEYS = SPATIAL_KEYS + COUNT_KEYS +
-                   %i[text id class style visible obscured exact exact_text normalize_ws match wait filter_set]
+                   %i[text id class style visible obscured exact exact_text normalize_ws match wait filter_set message]
       VALID_MATCH = %i[first smart prefer_exact one].freeze
 
       def initialize(*args,
--- a/lib/capybara/result.rb
+++ b/lib/capybara/result.rb
@@ -115,6 +115,8 @@ module Capybara
     end
 
     def failure_message
+      return @query.options[:message] if @query.options[:message]
+
       message = @query.failure_message
       if count.zero?
         message << ' but there were no matches'


twalpole:
Please give an example of the original message not being clear enough.  My issue with a static message override is that it won't take into account any of the system filters or new filters potentially added to the builtin selectors in the future.

If someone wants/needs custom error messages (which one hopes should never be seen) then they should probably be writing custom selectors for specific object types on the page which would allow them generate whatever message they want.

joelhawksley:
@twalpole thank you for your response! ❤️ I'll provide an example:

We have a test that iterates through several cases and runs `assert_selector` against each:

```ruby
[FooComponent, BarComponent, BazComponent].each do |component|
  render_inline(component.new)
  assert_selector(".my-4")
end
```

Right now, the error message for the failure is something like: `attempted to find selector .my-4, but there were no matches`. This error message isn't super helpful to a developer, because they don't know which `component` resulted in the error. (In practice, our array of cases is ~30, making the issue more frustrating).

Ideally, we could do:

```ruby
[FooComponent, BarComponent, BazComponent].each do |component|
  render_inline(component.new)
  assert_selector(".my-4", message: "#{component} does did not render my-4 selector")
end
```

Thus providing developers with clear feedback about which case is broken.

FWIW, RSpec and Minitest provide similar functionality. For example, `assert_equal` in Minitest has a message option: https://docs.ruby-lang.org/en/2.1.0/MiniTest/Assertions.html#method-i-assert_equal.

> which one hopes should never be seen

The test in question above serves as much to educate developers as it does to prevent regressions. At our scale, we do plenty of "help (error) message driven development", where linters and tests guide developers to make good decisions. I'm hoping to be able to use Capybara assertions as part of this strategy ❤️

twalpole:
RSpec and minitest provide the behavior because their assertions are very generic and without it the error messages can be pretty meaningless.  For your example this behavior could make sense, but I'm not sure it does in every case because it loses the details that the builtin failure messages provide about why exactly it's failing.  Maybe a prefix message or a template message that allows to include the normal error message would be more useful.

twalpole:
Also, since you know in your test that the selector is CSS you should probably be using `assert_css` rather than `assert_selector`

joelhawksley:
> Maybe a prefix message or a template message that allows to include the normal error message would be more useful.

I love that idea. I think the approach of providing a prefix would be great. Would you like me to update this PR to take that approach?

twalpole:
Thinking about this more I realized the fact that this behavior is already builtin to RSpec probably means implementing it as a parameter passed to the selectors doesn't make a lot of sense.  This should probably just be implemented in the minitest assertion adapters

    expect(page).to have_css('h4'), 'custom failure message'  # already works for RSpec
    expect(page).to have_css('h4', message: 'custom failure message') # Seems redundant/confusing

joelhawksley:
@twalpole correct, we are using it with Minitest. So a prefix, only for Minitest, then?

twalpole:
I'm thinking first step just a custom message, like your initial implementation, for each minitest assertion so they behave like Minitests builtin assertions.  Unfortunately that's going to get complicated pretty quickly for some of the selectors, but if you're up for it...   For instance

    assert_all_of_selectors(:css, 'div', 'p', 'span') 

is valid syntax.  That could be required to be 

   assert_all_of_selectors(:css, ['div', 'p', 'span'], 'custom message')

if the user wants to use a custom failure message but what to do about the possible options? Does it become

    assert_all_of_selectors(:css, ['div', 'p', 'span'], 'custom message', visible: :all) 

or 

    assert_all_of_selectors(:css, ['div', 'p', 'span'], visible: :all, 'custom_message') 

I think implementing this is going to mean getting rid of most of the metaprogramming in lib/capybara/minitest.rb

joelhawksley:
@twalpole would using a `message` key simplify the implementation? Regardless, I'll take a look at the custom message solution. Thanks for the feedback 👍

twalpole:
It would simplify implementation but then make it inconsistent with all the default minitest assertions which doesn't seem wise

PR #2374, opened 2020-07-28
Update to a newer nokogiri

rpbarlow:
Signed-off-by: Randy Barlow <rbarlow@credible.com>


twalpole:
The specification of ~> 1.8 allows for 1.x where x >= 8 so there is no need for this change.

PR #2377, opened 2020-08-04
Prefer symbol keys.

ioquatix:
Using symbol keys reduces memory usage and improves performance slightly.

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -258,7 +258,7 @@ module Capybara
     #
     def current_driver
       if threadsafe
-        Thread.current['capybara_current_driver']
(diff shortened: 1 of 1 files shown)


twalpole:
Thanks this is definitely in when we get to 4.0 - prior to that I need to think about whether I consider this a breaking change

ioquatix:
Great.

anikettstiwari:
@ioquatix  Can you share some stats when use symbol over string reduces memory usage

pankaj3150:
can you add more two the regester

ioquatix:
@twalpole also note that `Thread.current[...]` is fiber local, not thread local, if that is what you are expecting.

Issue #2371, opened 2020-07-23
Calling 'has_key?' on the drivers/servers container is deprecated without replacement


thoraxe:
## Meta
Capybara Version:
3.33.0

Driver Information (and browser if relevant):
selenium-webdriver (3.142.7)
webdrivers (4.4.1)
websocket-driver (0.7.1)
turnip (4.1.0)
cucumber-gherkin (10.0.0)
cucumber-messages (10.0.3)
protobuf-cucumber (3.10.8)

Mozilla Firefox 68.10.0esr

## Expected Behavior
More information about deprecation warnings or what is actually wrong

## Actual Behavior
`DEPRECATED: Calling 'has_key?' on the drivers/servers container is deprecated without replacement`

## Steps to reproduce
I'm not entirely sure. I upgraded to Capybara 3.33.0 and, when running my test, got this warning.

3.32.2 does not produce this warning.

I'm not sure if this is related to turnip/cucumber, or something else. The message tells me nothing about which code invoked `has_key?` so I don't know how to troubleshoot further.

Is there a way to get more verbosity out of the warning?

twalpole:
Treating the driver registrations as a Hash is deprecated. Something in your code/included libraries is calling `Capybara.drivers.has_key?` and that won't be supported in Capybara 4.x.  As of now no further information is output although PRs to add that would be accepted.

thoraxe:
@twalpole How the heck can I figure out what's calling that? I have... dozens of gems?

It's not in my code:

```
app/api/result_wrapper.rb:        if parsed_body.has_key?(COLLECTION_KEY)
node_modules/node-sass/src/libsass/src/sass_util.hpp:        if not grouped.has_key?(key) then
```
The above is from a grep on the entire codebase. So, the only thing I can think is that it's in one of the gems related to Capybara?

luke-hill:
@thoraxe If those gems are doing proper semver, then when capybara 4.0 is released they will do fixes their end, or they will fix this up beforehand.

PR #2379, opened 2020-08-12
Add uplelvel to deprecation warning of Capybara

koic:
Follow https://github.com/teamcapybara/capybara/issues/2371.

This PR adds `uplelvel` to deprecation warning of Capybara.
The `uplevel` option has been introduced from Ruby 2.6.

> If the `uplevel` keyword argument is given, the string will be prepended with information for the given caller frame in the same format used by the `rb_warn` C function.

https://ruby-doc.org/core-2.6.0/Kernel.html#method-i-warn

On the other hand, Capybara supports Ruby 2.5.
Therefore, This PR adds `Capybara::Helpers.warn` internal method for emulating `warn 'message', uplevel: 1` in Ruby 2.5 or lower.
This option will clarify where deprecation warnings should be fixed.

## Before

```console
DEPRECATED: Calling 'has_key?' on the drivers/servers container is
deprecated without replacement
```

## After

```console
/Users/koic/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/turnip-4.1.0/
lib/turnip/capybara.rb:10: warning:
DEPRECATED: Calling 'has_key?' on the drivers/servers container is
deprecated without replacement
```

### DIFF
--- a/lib/capybara/helpers.rb
+++ b/lib/capybara/helpers.rb
@@ -15,7 +15,7 @@ module Capybara
     # @return [String]         Normalized text
     #
     def normalize_whitespace(text)
-      warn 'DEPRECATED: Capybara::Helpers::normalize_whitespace is deprecated, please update your driver'
+      Capybara::Helpers.warn 'DEPRECATED: Capybara::Helpers::normalize_whitespace is deprecated, please update your driver'
       text.to_s.gsub(/[[:space:]]+/, ' ').strip
     end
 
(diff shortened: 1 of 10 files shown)


twalpole:
hmm -- is it documented somewhere that it's supported in Ruby 2.5? - https://docs.ruby-lang.org/en/2.5.0/Kernel.html#method-i-warn -  Does the latest JRuby which claims to be 2.5 compliant support it?

koic:
My explanation and first implementation was wrong. `uplevel` option has been introduced by Ruby 2.6.
JRuby 9.2 supports `uplevel`, but Ruby 2.5 does not support.

I updated this PR to add `Capybara::Helpers.warn` internal method for emulating `warn 'message', uplevel: 1` in Ruby 2.5 or lower.

lib/capybara/helpers.rb
```diff
@@ -81,6 +81,16 @@ def filter_backtrace(trace)
       new_trace.first.split(/:in /, 2).first
     end
 
+    # Workaround for emulating `warn '...', uplevel: n` in Ruby 2.5 or lower.
+    def warn(message, uplevel: 0)
```
twalpole, line 85: Everywhere this is called passes 1 - so just default to 1 and don't pass it

koic: Sure! I updated this PR.

ttanimichi:
Are you planning to release this in the next version?

twalpole:
It was merged, so yes

mileslane:
I am running 3.33.0, but I am still seeing the old behavior.
`DEPRECATED: Calling 'has_key?' on the drivers/servers container is deprecated without replacement`
What gives?

twalpole:
@mileslane 3.34 hasn’t been released yet - if you want this now use the master branch

PR #2385, opened 2020-08-31
use Selenium::WebDriver::VERSION vs. Gem.loaded_v

joelhawksley:
This change switches a check against the version
of `selenium-webdriver` to use the version number
defined by the library itself instead of accessing
it via `Gem.loaded_version`, as the latter relies
on the gem being loaded via RubyGems a particular
way that we happen to not be using at in the
Github monolith :)

Co-authored-by: Natasha Umer <natashau@github.com>
Co-authored-by: John Crepezzi <seejohnrun@github.com>

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -21,7 +21,24 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
       require 'capybara/selenium/patches/atoms'
       require 'capybara/selenium/patches/is_displayed'
       require 'capybara/selenium/patches/action_pauser'
-      if Gem::Version.new(Selenium::WebDriver::VERSION) < Gem::Version.new('3.5.0')
+
+      # Look up the version of `selenium-webdriver` to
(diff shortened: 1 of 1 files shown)


twalpole:
Unfortunately IIRC selenium-webdriver doesn't define Selenium::WebDriver::VERSION in the oldest versions we technically currently support which is why this check was done the way it was.  We could dump older versions of selenium-webdriver and only support versions which define their version, or this check needs to be more complicated and combine both methods.  Either way it's not acceptable as currently submitted.

joelhawksley:
@twalpole makes sense. I've rewritten the implementation to only use the constant if the `loaded_specs` approach comes up empty. Let me know what you think ❤️

joelhawksley:
@twalpole I'm having a hard time telling if any of the CI failures are legitimate. I don't believe they are. Can you confirm? ❤️

twalpole:
They don't appear to be related to this -- seems more like a FF version update has broken a few things.

PR #2388, opened 2020-09-09
Clarify Finders#find visible option default value

asavageiv:


asavageiv:
I do not think any of the check failures have to do with my change. The build is broken at the HEAD of master

twalpole:
Thanks

PR #2399, opened 2020-09-29
Add "scroll-behavior: auto" CSS to Capybara::Server::AnimationDisabler

ndbroadbent:
I have some test cases where I need to call `page.execute_script('window.scrollTo(0, 0)')` before taking a screenshot. I'm using [Tabler](https://tabler.io/), which [sets `scroll-behavior: smooth;`](https://github.com/tabler/tabler/blob/542e3541c18ccbb7d26265fd70b9357f4f4e23ea/scss/layout/_core.scss#L4), so any scroll actions are smoothly animated. This is nice in production, but annoying during tests.

This proposed change will cause the scroll to always complete instantly without any animations. (This will only work when using the default `selector_for` value of `'*'`, so that the rule matches the `html` element.)

[`scroll-behavior: auto` is the default value](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior), so I don't think this change should cause any problems.

I've added some tests, including an "inverse" test for `Capybara.disable_animation = false`. I think this important to show that `scroll-behavior: smooth;` does affect the scrolling speed, so that the test will have no false negatives.

EDIT: I just remembered another change that I made in my own version:

```
caret-color: transparent;
```

This is also very helpful during some of my screenshot diff tests where I'm checking that form inputs are rendered and filled in correctly.

Would it be ok to add `caret-color: transparent;`? I would love to start using the built-in `AnimationDisabler` from Capybara, but I really need this `caret-color: transparent;` as well. Or would it be possible to add the option to inject my own custom CSS?

EDIT 2: What do you think about [this `disable_animation_extra_properties` change](https://github.com/DocSpring/capybara/commit/54f333ff20bcbf94ad6e962e0e25bedc60ab0d29), so that I can set `Capybara.disable_animation_extra_properties = 'caret-color: transparent;'`?

### DIFF
--- a/spec/shared_selenium_session.rb
+++ b/spec/shared_selenium_session.rb
@@ -388,6 +388,39 @@ RSpec.shared_examples 'Capybara::Session' do |session, mode|
           @animation_session.find_link('animate me away').right_click
           expect(@animation_session).to have_content('Animation Ended', wait: 0.1)
         end
+
+        it 'should scroll the page instantly', requires: [:js] do
+          @animation_session.visit('with_animation')
+          scroll_y = @animation_session.evaluate_script(<<~JS)
+            (function(){
+              window.scrollTo(0,500);
+              return window.scrollY;
+            })()
+          JS
+          expect(scroll_y).to eq 500
+        end
+      end
+
+      context 'when set to `false`' do
+        before(:context) do # rubocop:disable RSpec/BeforeAfterAll
+          skip "Safari doesn't support multiple sessions" if safari?(session)
+          # NOTE: Although Capybara.SpecHelper.reset! sets Capybara.disable_animation to false,
+          # it doesn't affect any of these tests because the settings are applied per-session
+          Capybara.disable_animation = false
+          @animation_session = Capybara::Session.new(session.mode, TestApp.new)
+        end
+
+        it 'should scroll the page with a smooth animation', requires: [:js] do
+          @animation_session.visit('with_animation')
+          scroll_y = @animation_session.evaluate_script(<<~JS)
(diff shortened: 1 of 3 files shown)


ndbroadbent:
It looks like I'm seeing the same test failure as the master branch, but otherwise all the tests are passing.

* PR: https://travis-ci.org/github/teamcapybara/capybara/jobs/731188374
* master: https://travis-ci.org/github/teamcapybara/capybara/jobs/731188374

spec/shared_selenium_session.rb
```diff
@@ -388,6 +388,31 @@
           @animation_session.find_link('animate me away').right_click
           expect(@animation_session).to have_content('Animation Ended', wait: 0.1)
         end
+
+        it 'should scroll the page instantly', requires: [:js] do
+          @animation_session.visit('with_animation')
```
twalpole, line 393: I think these tests would probably be more reliable if it was done all in one `evaluate_script`. Something like - ```scrollY = @animation_session.evaluate_script(<<~JS)
      (function(){
        window.scrollTo(0,500);
        return window.scrollY;
      })()
    JS```

twalpole:
Thanks for this - I've made one comment around the tests, and you also have rubocop style failures - https://travis-ci.org/github/teamcapybara/capybara/jobs/731188375.  You can run `bundle exec rubocop' before committing

ndbroadbent:
Thanks for the review @twalpole! I've updated the tests with your suggestions, and also fixed all the style failures (sorry about those!)

twalpole:
Thanks! LGTM

PR #2400, opened 2020-09-29
Find <button> elements through <label>

seanpdoyle:
Browsers consider [<button> elements][mdn-button] as
["labelable"][labelable], meaning that a `<button
id="button-with-label">` element can be labelled by a [`<label
for="button-with-label">` element][mdn-label].

This commit extends the button selector to resolve `<button>` elements
with a locator that references a `<label>` element's text. That
`<label>` element can either reference the button through its `[for]`
attribute, or it can be an ancestor of the `<button>` element.

[labelable]: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Form_labelable
[mdn-button]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button
[mdn-label]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

### DIFF
--- a/lib/capybara/selector/definition/button.rb
+++ b/lib/capybara/selector/definition/button.rb
@@ -19,11 +19,12 @@ Capybara.add_selector(:button, locator_type: [String, Symbol]) do
       locator_matchers |= XPath.attr(test_id) == locator if test_id
 
       input_btn_xpath = input_btn_xpath[locator_matchers]
+      input_btn_xpath += XPath.descendant(:label)[XPath.string.n.is(locator)].descendant(input_btn_xpath)
 
       btn_xpath = btn_xpath[locator_matchers |
                             XPath.string.n.is(locator) |
-                            XPath.descendant(:label)[XPath.string.n.is(locator)] |
                             XPath.descendant(:img)[XPath.attr(:alt).is(locator)]]
+      btn_xpath += XPath.descendant(:label)[XPath.string.n.is(locator)].descendant(btn_xpath)
 
       alt_matches = XPath.attr(:alt).is(locator)
       alt_matches |= XPath.attr(:'aria-label').is(locator) if enable_aria_label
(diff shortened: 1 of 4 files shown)


twalpole:
Thanks for the this, however labelable elements can also be labeled by wrapping them inside a label element without a `for` attribute
```
     <label> My label <button>blah</button></label>
```
Please handle this case too - https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selector/selector.rb#L135

lib/capybara/selector/definition/button.rb
```diff
@@ -12,14 +12,17 @@
...
+                         (XPath.attr(:id) == XPath.anywhere(:label)[XPath.string.n.is(locator)].attr(:for))
+
       locator_matchers |= XPath.attr(:'aria-label').is(locator) if enable_aria_label
       locator_matchers |= XPath.attr(test_id) == locator if test_id
 
       input_btn_xpath = input_btn_xpath[locator_matchers]
 
       btn_xpath = btn_xpath[locator_matchers |
                             XPath.string.n.is(locator) |
+                            XPath.descendant(:label)[XPath.string.n.is(locator)] |
```
seanpdoyle, line 25: @twalpole I've added this line as part of locator resolution, but the suite is failing.

I have to admit, I struggle with composing XPath declarations, and would really appreciate any guidance you could provide.

twalpole:
@seanpdoyle I don't have time to do a deep dive into this currently, but at first glance it looks like you're applying the wrapping label as a condition on the btn rather than as a union with it.  You probably need something like 

    btn_xpath = btn_xpath[locator_matchers |  XPath.string.n.is(locator)]  + XPath.descendant(:label)[XPath.string.n.is(locator)].descendant(btn_xpath)

and that probably needs to be done for all 3 of the different "button" types - although I'm not 100% sure on that without spending more time refreshing exactly what it's looking for.

twalpole:
Thanks for this

Issue #2403, opened 2020-10-09
Capybara's GET_XPATH_SCRIPT throws a type error when there is no valid xpath


cam-narzt:
xpath cannot deal with shadow dom, webpages can have shadow dom, when they do capybara throws this:

`Selenium::WebDriver::Error::JavascriptError: TypeError: can't access property "namespaceURI", el is null`

That comes from this script: https://github.com/teamcapybara/capybara/blob/f43b6499ecb2721dd97b63a7fe31ff2518b89b4c/lib/capybara/selenium/node.rb#L491-L518

In order to prevent this, line 505 should be: `if (!el) return '';` or something.

twalpole:
The Webdriver spec doesn't yet specify shadow dom support, and none of Capybaras selectors support it.  Work was started to implement Capybaras selectors in a way that they can all work in either XPath or CSS but it was never completed.

evisong:
> The Webdriver spec doesn't yet specify shadow dom support, and none of Capybaras selectors support it. Work was started to implement Capybaras selectors in a way that they can all work in either XPath or CSS but it was never completed.

Update Sep 2021: Webdriver spec updated with shadow root section.

https://www.w3.org/TR/2021/WD-webdriver2-20210921/
https://www.w3.org/TR/webdriver2/#shadow-root

Is there any plan to follow up? Thanks.

PR #2406, opened 2020-10-10
Detect shadow dom elements when calling `path`

twalpole:


twalpole:
Should fix Issue #2403

PR #2413, opened 2020-10-27
Call conf.clamp for puma options to convert the default :environment Proc into a String

ndbroadbent:
Hello,

I just saw a deprecation warning when running Puma 5.0.3 in Ruby 2.7: `warning: deprecated Object#=~ is called on Proc`. I initially raised the issue on the puma repo: https://github.com/puma/puma/issues/2455
But it turns out that this was a bug in Capybara: https://github.com/puma/puma/issues/2455#issuecomment-716724207

Capybara was not calling `#finalize_values` to convert default Proc into a string. I've also tried to add a test to confirm that this is fixed, but not sure if it's the best way to test this.

### DIFF
--- a/lib/capybara/registrations/servers.rb
+++ b/lib/capybara/registrations/servers.rb
@@ -28,6 +28,7 @@ Capybara.register_server :puma do |app, port, host, **options|
   options = default_options.merge(options)
 
   conf = Rack::Handler::Puma.config(app, options)
+  conf.clamp
   events = conf.options[:Silent] ? ::Puma::Events.strings : ::Puma::Events.stdio
 
   puma_ver = Gem::Version.new(Puma::Const::PUMA_VERSION)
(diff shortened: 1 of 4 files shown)


twalpole:
I don't know that I agree it should be the calling codes responsibility to call `finalize_values`.  I need to find time to go read the puma code and see whether we're using private apis or public.

lib/capybara/registrations/servers.rb
```diff
@@ -28,6 +28,7 @@
   options = default_options.merge(options)
 
   conf = Rack::Handler::Puma.config(app, options)
+  conf.options.finalize_values
```
dentarg, line 31: I think you can do this instead

```suggestion
  conf.clamp
```

See

* https://github.com/puma/puma/blob/v5.0.4/lib/puma/configuration.rb#L127-L130
* https://github.com/puma/puma/blob/v5.0.4/lib/puma/configuration.rb#L224-L228

ndbroadbent:
Thanks @dentarg, you're right! This looks like the public API that we are meant to call. I have updated the PR

ndbroadbent:
Note: I opened a different PR to fix some RuboCop warnings from the master branch: #2417

I've also included that commit in here just to check that the build is passing for this branch

twalpole:
Thanks

PR #2418, opened 2020-10-30
Added `Capybara.disable_animation_extra_css` option to insert extra CSS in Capybara::Server::AnimationDisabler::DISABLE_MARKUP_TEMPLATE

ndbroadbent:
This is just a follow-up change for my previous PR: https://github.com/teamcapybara/capybara/pull/2399

This new option can be used to add additional CSS rules to the `AnimationDisabler` middleware.

I also need to use `caret-color: transparent;` to hide any cursors in inputs. I have some screenshot tests that ensure that some inputs are visually correct on the page.

(I don't know if this is technically "disabling animation" since the cursor is blinking twice per second, and this change makes it possible to remove the blinking cursor.)

I'm currently doing this in my code:

```
  Capybara::Server::AnimationDisabler.class_eval do
    # rubocop:disable Lint/ConstantDefinitionInBlock
    DISABLE_MARKUP_TEMPLATE_CARET_TRANSPARENT = <<~HTML
      <script defer>(typeof jQuery !== 'undefined') && (jQuery.fx.off = true);</script>
      <style>
        %<selector>s, %<selector>s::before, %<selector>s::after {
          transition: none !important;
          animation-duration: 0s !important;
          animation-delay: 0s !important;
          scroll-behavior: auto !important;
          caret-color: transparent;
        }
      </style>
    HTML
    # rubocop:enable Lint/ConstantDefinitionInBlock

    def initialize(app)
      @app = app
      @disable_markup = format(
        DISABLE_MARKUP_TEMPLATE_CARET_TRANSPARENT,
        selector: self.class.selector_for(Capybara.disable_animation)
      )
    end
  end
```

So I'd like to get rid of the class_eval and replace it with this: `Capybara.disable_animation_extra_css = 'caret-color: transparent;'`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "files.trimTrailingWhitespace": true
+}
--- a/History.md
+++ b/History.md
@@ -4,6 +4,8 @@ Release date: unreleased
 ### Added
 
 * Ability to fill in with emoji when using Chrome with selenium driver (Firefox already worked)
+* Added `Capybara.disable_animation_extra_css` option, which can be used to add additional CSS rules
+  to the AnimationDisabler middleware [Nathan Broadbent]
 
 ### Fixed
 
--- a/lib/capybara/server/animation_disabler.rb
+++ b/lib/capybara/server/animation_disabler.rb
@@ -16,7 +16,11 @@ module Capybara
 
       def initialize(app)
         @app = app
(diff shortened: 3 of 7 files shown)


twalpole:
I don't want to add another setting for this.  If you implement it so Capybara.disable_animation optionally accepts multiple arguments I'd accept it

    Capybara.disable_animation = true, 'caret-color: transparent;'

for instance

alan-pie:
I would love to have the feature @twalpole mentioned. Did this ever get implemented? The default `disable_animations` doesn't work for Bootstrap 4.6 modal fades.

PR #2422, opened 2020-11-10
Declare `Session#send_keys`

seanpdoyle:
When covering hotkeys and tab-navigation with test coverage, sending
keys via particular elements (via `Node#send_keys`) can be tedious.

As an alternative, invoke `Session#send_keys` to delegate to the element
with current focus. In most (all?) browsers, `document.activeElement`
will return the `<body>` element when there are no elements with focus.

In Selenium-enabled environments, `Session#send_keys` delegates to the
current element with focus (via [document.activeElement][]). Treat
`Session#send_keys` as a no-operation in Rack Test environments.

[document.activeElement]: https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement

### DIFF
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -133,6 +133,10 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
     unwrap_script_result(result)
   end
 
+  def send_keys(*args)
+    active_element.send_keys(*args)
+  end
+
   def save_screenshot(path, **_options)
     browser.save_screenshot(path)
   end
@@ -453,6 +457,10 @@ private
 
   def find_context
     browser
+  end
+
+  def active_element
+    browser.switch_to.active_element
   end
 
   def build_node(native_node, initial_cache = {})
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -58,7 +58,7 @@ module Capybara
     ].freeze
     SESSION_METHODS = %i[
       body html source current_url current_host current_path
-      execute_script evaluate_script visit refresh go_back go_forward
+      execute_script evaluate_script visit refresh go_back go_forward send_keys
       within within_element within_fieldset within_table within_frame switch_to_frame
       current_window windows open_new_window switch_to_window within_window window_opened_by
       save_page save_and_open_page save_screenshot
@@ -301,6 +301,14 @@ module Capybara
     #
     def go_forward
       driver.go_forward
+    end
+
+    ##
+    # @!method send_keys
+    #   @see Capybara::Node::Element#send_keys
+    #
+    def send_keys(*args, **kw_args)
+      driver.send_keys(*args, **kw_args)
     end
 
     ##
--- a/spec/rack_test_spec.rb
+++ b/spec/rack_test_spec.rb
@@ -145,6 +145,14 @@ RSpec.describe Capybara::Session do # rubocop:disable RSpec/MultipleDescribes
       end
     end
 
+    describe '#send_keys' do
+      it 'raises an UnsupportedMethodError' do
+        session.visit('/form')
+
+        expect { session.send_keys(:tab) }.to raise_error(Capybara::NotSupportedByDriverError)
+      end
+    end
+
     describe '#text' do
       it 'should return original text content for textareas' do
         session.visit('/with_html')
--- a/spec/shared_selenium_session.rb
+++ b/spec/shared_selenium_session.rb
@@ -222,6 +222,18 @@ RSpec.shared_examples 'Capybara::Session' do |session, mode|
       end
     end
 
+    describe '#send_keys' do
+      it 'defaults to sending keys to the document.activeElement' do
+        session.visit('/form')
+
+        expect(session.evaluate_script('document.activeElement')).to eq(session.find('//body'))
+
+        session.send_keys(:tab)
+
+        expect(session.evaluate_script('document.activeElement')).to eq(session.first(:field))
+      end
+    end
+
     describe '#path' do
       it 'returns xpath' do
         # this is here because it is testing for an XPath that is specific to the algorithm used in the selenium driver
--- a/History.md
+++ b/History.md
@@ -7,6 +7,12 @@ Release date: unreleased
 * Current path assetsions/expectations accept optional filter block
 * Animation disabler now specifies `scroll-behavior: auto;` [Nathan Broadbent]
 * :button selector can now find elements by label text [Sean Doyle]
+* In Selenium-enabled environments, `Session#send_keys` delegates to the current
+  element with focus (via [document.activeElement][]). When called from Rack
+  Test environments, `Session#send_keys` raises an error [Sean Doyle]
+
+[document.activeElement]: https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement
+[Sean Doyle]: https://github.com/seanpdoyle/
 
 ### Changed
 
--- a/lib/capybara/driver/base.rb
+++ b/lib/capybara/driver/base.rb
@@ -59,6 +59,10 @@ class Capybara::Driver::Base
     raise Capybara::NotSupportedByDriverError, 'Capybara::Driver::Base#status_code'
   end
 
+  def send_keys(*)
+    raise Capybara::NotSupportedByDriverError, 'Capybara::Driver::Base#send_keys'
+  end
+
   ##
   #
   # @param frame [Capybara::Node::Element, :parent, :top]  The iframe element to switch to


lib/capybara/session.rb
```diff
@@ -303,6 +303,10 @@ def go_forward
       driver.go_forward
     end
 
+    def send_keys(*args, **kw_args)
```
seanpdoyle, line 306: Is there a way to share documentation with the `Node#send_keys` version, or would we need to declare comment docs here as well?

spec/rack_test_spec.rb
```diff
@@ -145,6 +145,13 @@ module TestSessions
       end
     end
 
+    describe '#send_keys' do
+      it 'is a no-operation' do
+        session.visit('/form')
+        session.send_keys(:tab)
```
seanpdoyle, line 151: Since it's a no-op, I'm not sure what the assertion should be. Would an `expect { }.not_to raise_error` be a good fit, or would we be better off skipping it like other JS-enabled tests?

lib/capybara/selenium/driver.rb
```diff
@@ -455,6 +459,10 @@ def find_context
     browser
   end
 
+  def active_element
```
seanpdoyle, line 462: Would this be a good method to add to the `Session` public API?

lib/capybara/rack_test/driver.rb
```diff
@@ -50,6 +50,9 @@ def refresh
     browser.refresh
   end
 
+  def send_keys(*)
```
twalpole, line 53: If this can't do anything it should be an error to call it

spec/rack_test_spec.rb
twalpole: This should raise an error - calling it in the rack test driver makes no sense and would just lead to confusion

lib/capybara/rack_test/driver.rb
seanpdoyle: Thanks for clarifying @twalpole. Is declaring it still worthwhile? Is there a better (more Capybara-specific) alternative to `NoMethodError`?

lib/capybara/selenium/driver.rb
twalpole: I don't think so at the moment

twalpole:
I need to check but I think the selenium api might provide access to the active element which would be preferable to use instead of evaluate_script

lib/capybara/rack_test/driver.rb
twalpole: Yes - we should have a specific error mentioning that the rack test driver doesn't have a concept of active element -- it should not be a nomethoderror

lib/capybara/selenium/driver.rb
```diff
@@ -455,6 +459,10 @@ def find_context
...
+    build_node evaluate_script("document.activeElement")
```
twalpole, line 463: Pretty sure evaluate_script already calls `build_node` when returning  elements

seanpdoyle: You're right: https://github.com/teamcapybara/capybara/blob/22252382421f3126d6e10fa90a63237ff817436a/lib/capybara/selenium/driver.rb#L441-L452

lib/capybara/selenium/driver.rb
```diff
@@ -460,7 +460,7 @@ def find_context
   end
 
   def active_element
-    evaluate_script("document.activeElement")
+    @browser.switch_to.active_element
```
seanpdoyle, line 463: @twalpole I've swapped this with the Selenium method, though I'm not sure of the cross-platform implications of `@browser.switch_to`.

twalpole: hmmm -- I'll have to go look at the selenium code to see exactly what this is doing

spec/shared_selenium_session.rb
```diff
@@ -302,6 +302,15 @@
       end
     end
 
+    describe '#send_keys' do
+      it 'defaults to sending keys to the document.activeElement' do
+        session.visit('/form')
+        session.send_keys(:tab)
```
twalpole, line 308: This should pre-verify the active element - otherwise the test isn't really proving anything

lib/capybara/selenium/driver.rb
twalpole: Found some time, looks fine

lib/capybara/selenium/driver.rb
```diff
@@ -455,6 +459,10 @@ def find_context
     browser
   end
 
+  def active_element
+    @browser.switch_to.active_element
```
twalpole, line 463: Needs to be `browser` not `@browser`

lib/capybara/session.rb
twalpole: Just link to Node#send_keys docs (See ...)

twalpole:
Along with the current code `send_keys` would need to be added to Capybara::Driver::Base

lib/capybara/session.rb
```diff
@@ -303,6 +303,14 @@ def go_forward
       driver.go_forward
     end
 
+    ##
+    # @!method send_keys
+    #   @see Capybara::Node::Element#send_keys
```
seanpdoyle, line 308: @twalpole is this the correct Rubydoc syntax to link?

spec/rack_test_spec.rb
```diff
@@ -145,6 +145,14 @@ module TestSessions
       end
     end
 
+    describe '#send_keys' do
+      it 'raises an UnsupportedMethodError' do
```
seanpdoyle, line 149: @twalpole is a new and domain-specific `UnsupportedMethodError` preferable to a `NotImplementedError`?

twalpole: @seanpdoyle I think for the racktest driver instance something like

    raise Capybara::NotSupportedByDriver, "RackTest driver doesn't have the concept of an active element, please call send_keys on a specific element"

probably makes the most sense

seanpdoyle: Thanks! Done.

lib/capybara/rack_test/driver.rb
```diff
@@ -50,6 +50,10 @@ def refresh
...
+    raise Capybara::NotSupportedByDriverError, "RackTest driver doesn't have the concept of an active element, please call send_keys on a specific element"
```
seanpdoyle, line 54: @twalpole I've updated this error with the message you've suggested.

Reading it back: is it true that Rack Test supports key handlers?

twalpole: Good catch -- I hadn't thought that through -- no it doesn't, and calling `send_keys` on an element would just raise a NotImplementedError, so this might as well be the same via the Capybara::Driver::Base default -- sorry about that.

lib/capybara/selenium/driver.rb
```diff
@@ -133,6 +133,10 @@ def evaluate_async_script(script, *args)
     unwrap_script_result(result)
   end
 
+  def send_keys(*args)
+    active_element.send_keys(*args)
```
seanpdoyle, line 137: @twalpole there is [a test failure in CI](https://travis-ci.org/github/teamcapybara/capybara/jobs/743279918#L976-L985):

```
1) Capybara::Session with remote Chrome with selenium driver #send_keys defaults to sending keys to the document.activeElement

Failure/Error: raise Capybara::NotSupportedByDriverError, 'Capybara::Driver::Base#send_keys'

  Capybara::NotSupportedByDriverError:
    Capybara::Driver::Base#send_keys
  Shared Example Group: "Capybara::Session" called from ./spec/selenium_spec_chrome_remote.rb:73

# ./lib/capybara/driver/base.rb:63:in `send_keys'
# ./lib/capybara/session.rb:311:in `send_keys'
# ./lib/capybara/dsl.rb:58:in `block (2 levels) in <module:DSL>'
# ./spec/shared_selenium_session.rb:231:in `block (4 levels) in <top (required)>'
```

It seems like there are Selenium-enabled tests that _are not_ using the `Selenium::Driver`, and appear to be using `Driver::Base` or `RackTest::Driver`. Are the tests I've added missing a crucial setup step?

twalpole: @seanpdoyle You'll need to tag the test(s) as requiring send_keys  - like https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/session/node_spec.rb#L1062 - although I am curious why the remote chrome driver version of the test would be failing.

twalpole: OH - I see the issue - your test isn't calling `send_keys` on the session - https://github.com/teamcapybara/capybara/pull/2422/files#diff-0794c7a2e4cba5e9448e4dfb1bd723358f79c2e7a1a3a0fc879c6a8e86e793c6R231

This is triggering the creation of a new default session which defaults to the RackTest driver

seanpdoyle: 🤦 , thank you!

With the way that the [DSL module delegates to `#page`](https://github.com/teamcapybara/capybara/blob/e8f7189730428ed9af04610ec4d1933424c6906c/lib/capybara/dsl.rb#L49-L62), will that style of invocation work in consumer applications' test suites (outside this test suite)?

twalpole: Yes -- it's just that in the tests the session is passed in rather than letting a default one be created since it makes it easier for testing against multiple drivers or with multiple sessions.  In most users case they're working with `Capybara.current_session` and wouldn't be passing the actual session around.

twalpole:
Thank you

PR #2427, opened 2020-11-30
Limit ARIA role check on buttons

seanpdoyle:
Add test coverage to ensure that definitions and selectors for other
buttons aren't affected by the `enable_aria_role:` option.

I'm opening this as a draft to run the test suite, as part of trying to dig into an application-local surprising test failure.

### DIFF
--- a/lib/capybara/selector/definition/button.rb
+++ b/lib/capybara/selector/definition/button.rb
@@ -3,33 +3,29 @@
 Capybara.add_selector(:button, locator_type: [String, Symbol]) do
   xpath(:value, :title, :type, :name) do |locator, **options|
     input_btn_xpath = XPath.descendant(:input)[XPath.attr(:type).one_of('submit', 'reset', 'image', 'button')]
-    btn_xpath = XPath.descendant(:button)
-    btn_xpath += XPath.descendant[XPath.attr(:role).equals('button')] if enable_aria_role
+    btn_xpaths = [XPath.descendant(:button)]
+    btn_xpaths << XPath.descendant[XPath.attr(:role).equals('button')] if enable_aria_role
     image_btn_xpath = XPath.descendant(:input)[XPath.attr(:type) == 'image']
 
     unless locator.nil?
       locator = locator.to_s
-      locator_matchers = XPath.attr(:id).equals(locator) |
-                         XPath.attr(:name).equals(locator) |
-                         XPath.attr(:value).is(locator) |
-                         XPath.attr(:title).is(locator) |
-                         (XPath.attr(:id) == XPath.anywhere(:label)[XPath.string.n.is(locator)].attr(:for))
-      locator_matchers |= XPath.attr(:'aria-label').is(locator) if enable_aria_label
-      locator_matchers |= XPath.attr(test_id) == locator if test_id
+      locator_matchers = combine_locators(locator, config: self)
 
       input_btn_xpath = input_btn_xpath[locator_matchers] + locate_label(locator).descendant(input_btn_xpath)
 
-      btn_xpath = btn_xpath[locator_matchers |
-                            XPath.string.n.is(locator) |
-                            XPath.descendant(:img)[XPath.attr(:alt).is(locator)]
-                           ] + locate_label(locator).descendant(btn_xpath)
+      btn_xpaths = btn_xpaths.map do |btn_xpath|
+        btn_xpath[locator_matchers |
+                  XPath.string.n.is(locator) |
+                  XPath.descendant(:img)[XPath.attr(:alt).is(locator)]
+                 ] + locate_label(locator).descendant(btn_xpath)
+      end
 
       alt_matches = XPath.attr(:alt).is(locator)
       alt_matches |= XPath.attr(:'aria-label').is(locator) if enable_aria_label
       image_btn_xpath = image_btn_xpath[alt_matches] + locate_label(locator).descendant(image_btn_xpath)
     end
 
-    %i[value title type].inject(input_btn_xpath.union(btn_xpath).union(image_btn_xpath)) do |memo, ef|
+    %i[value title type].inject([input_btn_xpath, *btn_xpaths, image_btn_xpath].inject(&:union)) do |memo, ef|
       memo.where(find_by_attr(ef, options[ef]))
     end
   end
@@ -50,5 +46,17 @@ Capybara.add_selector(:button, locator_type: [String, Symbol]) do
 
   describe_node_filters do |disabled: nil, **|
     ' that is disabled' if disabled == true
+  end
+
+  def combine_locators(locator, config:)
+    [
+      XPath.attr(:id).equals(locator),
+      XPath.attr(:name).equals(locator),
+      XPath.attr(:value).is(locator),
+      XPath.attr(:title).is(locator),
+      (XPath.attr(:id) == XPath.anywhere(:label)[XPath.string.n.is(locator)].attr(:for)),
+      (XPath.attr(:'aria-label').is(locator) if config.enable_aria_label),
+      (XPath.attr(test_id) == locator if config.test_id)
+    ].compact.inject(&:|)
   end
 end
--- a/lib/capybara/spec/session/has_button_spec.rb
+++ b/lib/capybara/spec/session/has_button_spec.rb
@@ -46,8 +46,24 @@ Capybara::SpecHelper.spec '#has_button?' do
     expect(@session).to have_button('ARIA button', enable_aria_role: true)
   end
 
+  it 'should be true for a role=button within a label when enable_aria_role: true' do
+    expect(@session).to have_button('role=button within label', enable_aria_role: true)
+  end
+
   it 'should be false for role=button when enable_aria_role: false' do
     expect(@session).not_to have_button('ARIA button', enable_aria_role: false)
+  end
+
+  it 'should be false for a role=button within a label when enable_aria_role: false' do
+    expect(@session).not_to have_button('role=button within label', enable_aria_role: false)
+  end
+
+  it 'should not affect other selectors when enable_aria_role: true' do
+    expect(@session).to have_button('Click me!', enable_aria_role: true)
+  end
+
+  it 'should not affect other selectors when enable_aria_role: false' do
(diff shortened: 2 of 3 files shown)


seanpdoyle:
@twalpole I _believe_ that the `:button` selector definition becomes too lax when `Capybara.enable_aria_button = true`, which results in [false negatives for `have_no_button`](https://travis-ci.org/github/teamcapybara/capybara/jobs/746824288#L868-L871) and false positives for `have_button`.

Reading through the definition:

https://github.com/teamcapybara/capybara/blob/da55958746df3dcc3eae7ae238c15fba30e9f90d/lib/capybara/selector/definition/button.rb#L7

the root cause is not obvious to me. Are these test failures genuine? Do you have any ideas of the root cause?

twalpole:
@seanpdoyle I would guess that the  `locate_label(locator).descendant(image_btn_xpath)` at https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selector/definition/button.rb#L20 isn't producing the expected XPath

seanpdoyle:
> I would guess that the locate_label(locator).descendant(image_btn_xpath) at https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selector/definition/button.rb#L20 isn't producing the expected XPath

Thanks @twalpole, I'll investigate further. Is there a simple way to run `rspec` for one test at a time? I've had trouble with annotating line numbers to the file names, since most of the suite relies on shared examples.

twalpole:
tag the test(s) you want to run with `focus_` metadata, and then run the tests

    it 'blah blah', :focus_ do
      ....
    end

then run the basic tests

    bundle exec rake spec_rack

or with chrome

    bundle exec rake spec_chrome

lib/capybara/selector/definition/button.rb
```diff
@@ -21,8 +21,8 @@
 
       btn_xpath = btn_xpath[locator_matchers |
                             XPath.string.n.is(locator) |
-                            XPath.descendant(:img)[XPath.attr(:alt).is(locator)]
-                           ] + locate_label(locator).descendant(btn_xpath)
+                            XPath.descendant(:img)[XPath.attr(:alt).is(locator)]]
+      btn_xpath += locate_label(locator).descendant(btn_xpath)
```
seanpdoyle, line 25: @twalpole your hunch was correct, this seems to resolve the issue.

twalpole: Hmmm -- that doesn't seem right to me since it would mean the locator would have to match in the label and also match against the actual button element. I'll have to look at the XPath that's actually being generated -- probably won't have time to dig into that until next weekend

seanpdoyle: I've pushed up test coverage for this use case.

lib/capybara/selector/definition/button.rb
```diff
@@ -19,17 +19,19 @@
...
-      btn_xpath = btn_xpath[locator_matchers |
-                            XPath.string.n.is(locator) |
-                            XPath.descendant(:img)[XPath.attr(:alt).is(locator)]
-                           ] + locate_label(locator).descendant(btn_xpath)
+      btn_xpaths = btn_xpaths.map do |xpath|
+        xpath[locator_matchers |
+              XPath.string.n.is(locator) |
+              XPath.descendant(:img)[XPath.attr(:alt).is(locator)]
+             ] + locate_label(locator).descendant(xpath)
+      end
```
seanpdoyle, line 27: @twalpole mapping over the `btn_xpaths` instead of union-ing them from the start seems to have done the trick. Unfortunately, this loop and other changes has pushed this block over the rubocop limit. 

What techniques are at our disposal for refactoring long blocks in these DSLs?

lib/capybara/selector/definition/button.rb
twalpole: Reading the spec -- since &lt;label elements can't be associated with aria-role button elements it probably doesn't make sense to support them like that

seanpdoyle: @twalpole oddly enough, the feature work that uncovered this bug is related to `label [role="button"]` elements, inspired by the structure mentioned in [_A styled accessible file upload](https://accessabilly.com/a-styled-accessible-file-upload/):

```html
<label for="fileupload" id="buttonlabel">
  <span role="button" aria-controls="filename" tabindex="0">upload file</span>
</label>
```

Which doesn't get you "button" behavior out of the box, but the structure is accessible and flexible enough for more control.

Would you be able to link to the part of the specification you mentioned?

twalpole:
@seanpdoyle You can define methods inside the add_selector block

lib/capybara/selector/definition/button.rb
twalpole: In that case your label element should be associated with an id of "fileupload" - spec is https://www.w3.org/TR/html52/sec-forms.html#the-label-element - relevant portion is "The for attribute may be specified to indicate a form control with which the caption is to be associated. If the attribute is specified, the attribute’s value must be the ID of a labelable element" where https://www.w3.org/TR/html52/sec-forms.html#labelable-element defines labelable elements

seanpdoyle: This is great information to have, thanks for the links.

With that in mind, I don't think the implementation has been influenced at all by this use case, but this PR _does_ introduce test coverage for it.

Are you suggesting that we omit the tests that cover that behavior, namely:

* https://github.com/teamcapybara/capybara/pull/2427/files#diff-3497e3e50e487b0afcc2d1da0a08faf11991dc5986120253042adea31f3cfb6fR465-R468
* https://github.com/teamcapybara/capybara/pull/2427/files#diff-59d5da4957aa6a81605ec0fc1e0cda49c9734a63e7f099efcdc2efa69713b1bdR50
* https://github.com/teamcapybara/capybara/pull/2427/files#diff-59d5da4957aa6a81605ec0fc1e0cda49c9734a63e7f099efcdc2efa69713b1bdR58
* https://github.com/teamcapybara/capybara/pull/2427/files#diff-59d5da4957aa6a81605ec0fc1e0cda49c9734a63e7f099efcdc2efa69713b1bdR101
* https://github.com/teamcapybara/capybara/pull/2427/files#diff-59d5da4957aa6a81605ec0fc1e0cda49c9734a63e7f099efcdc2efa69713b1bdR109

twalpole: The question really is whether `it 'should be true for a role=button within a label when enable_aria_role: true' do
    expect(@session).to have_button('role=button within label', enable_aria_role: true)
  end` should actually find the element and pass. I'm leaning towards that it shouldn't since the `label` element isn't actually associated with the 'role=button' element (clicking the label element wouldn't actually do anything with regards to the `role=button` element)

seanpdoyle: > since the `label` element isn't actually associated with the 'role=button' element (clicking the label element wouldn't actually do anything with regards to the `role=button` element)

```html
<label for="fileupload" id="buttonlabel">
  <span role="button" aria-controls="filename" tabindex="0">upload file</span>
</label>
<input type="file" id="fileupload">
```

In this [example HTML's case](https://accessabilly.com/a-styled-accessible-file-upload/),  would `click_on "upload file"` resolve to the `span[role="button"]` element, and click within the `label` element to trigger the `input[type="file"]`?

seanpdoyle: In this PR's fixture HTML, `click_on "role=button within label element"` would do nothing, but `click_on "with other text"` would click the `[role=button]` and click the `label`:

```html
<label>
  role=button within label element
  <span role="button">with other text</span>
</label>
```

I can see that this is a use case that is odd and might be worth avoiding.

twalpole: In the exmaple HTML case it would work because 'upload file' would be found as the text in the "button" and the span would be clicked (which is where the click handler should be attached if the role is set as button.  I don't believe the `click_on "role=button within label element`  should work because the label element that contains the text isn't actually associated with the span (other than as a parent/child) - so if the user clicks inside the label but outside the span nothing will actually happen.

seanpdoyle: I've created a JSFiddle: https://jsfiddle.net/u7x1s5qr/ with the following HTML:

```html
<label for="file-upload">
  role=button within label element
  <span role="button" tabindex="0">
    with other text
  </span>
</label>
<input id="file-upload" type="file" hidden>
```

> so if the user clicks inside the label but outside the span nothing will actually happen.

In the JSFiddle, you can click on both the "role=button within label element" and the "with other text" to open the file dialog.

I would expect a Capybara call to `click_on "with other text"` (or `click_button`) to work.

> I don't believe the click_on "role=button within label element should work/find a button because the label element that contains the text isn't actually associated with the span (other than as a parent/child) 

As far as Capybara's behavior goes, I'd expect that a call to `click_on "role=button within label element"` would fail. This is somewhat different from what would happen by actually clicking on the `role=button within label element` text, but understanding that the implementation of `click_on` is an alias for `click_button` and `click_link`, it makes sense.

twalpole: The difference is that clicking on the `label` element works because it's associated with the actual `<input type="file">` element, not because it has the aria role button in it.  `find(:label, 'role=button within label element').click` will work too if you remove the `<span role="button"` element

seanpdoyle: I see, thanks for clarifying.

With that in mind, what needs to be done in the short term to push this over the finish line?

In addition to this test, there was also [a surprising CI failure](https://travis-ci.org/github/teamcapybara/capybara/jobs/747208490#L598).

seanpdoyle:
@twalpole what else could be done to help progress this forward? Do you have any ideas as to [what's causing the CI failure](https://travis-ci.org/github/teamcapybara/capybara/jobs/747208490#L598)?

twalpole:
@seanpdoyle Sorry haven't had any free time to spend on Capybara recently, will try to get to this over the holidays, as for the CI failure I assume Puma has released an update that has changed its behavior to internally call clamp.

twalpole:
I've expanded on this in https://github.com/teamcapybara/capybara/pull/2438 and will merge when tests pass

seanpdoyle:
Thank you!

Issue #2425, opened 2020-11-26
disable_animation injected script causes XHTML parsing error


dmorgan-fa:
Hi, I recently ran into some issues when enabling the `disable_animation` option:

https://github.com/teamcapybara/capybara/blob/f43b6499ecb2721dd97b63a7fe31ff2518b89b4c/lib/capybara/server/animation_disabler.rb#L47

The above line causes XHTML pages to incur a parsing error on the `defer` tag (see [here](https://monkeyraptor.johanpaul.net/2015/11/fixing-xhtml-parsing-error-on-chrome.html)) then on the logical `&&`.

Locally testing, I've found that the following resolves the issue: `<script defer="true">if (typeof jQuery !== 'undefined') {jQuery.fx.off = true};</script>`.

Some context around the issue:
While this might be limited in scope, it impacts some specs we have that are in place to provide support for a third party's import format.

Apologies, I'm unfamiliar with what's necessary to fully resolve this (PR, an appropriate test, etc.).

twalpole:
I didn't realize anyone still cared about XHTML :).  Additionally, I'd say if you're running validators against your document during your tests, you probably shouldn't be running them against documents that test tools have injected markup/code into  - since it's not really validating what would be served to an end-user.

That being said the "correct" solution here (rather than changing the JS) is probably to wrap the script and inline style sections in commented CDATA.  That will mean if the JS/CSS is updated in the future it won't be an issue.  If you could submit a PR for that it'd be great.

dmorgan-fa:
Thanks for the help @twalpole! That makes more sense as a solution.

I've created a PR, let me know if there's anything else I need to do 👍

PR #2428, opened 2020-12-01
Wrapping injected disable_animation script in CDATA to prevent XHTML/…

dmorgan-fa:
Addressing issue raised on [#2425](https://github.com/teamcapybara/capybara/issues/2425) - `disable_animation injected script causes XHTML parsing error`.

Wrapped the injected `disable_animation` script with CDATA to prevent XHTML/XML parser errors (see [here](https://monkeyraptor.johanpaul.net/2015/11/fixing-xhtml-parsing-error-on-chrome.html)).

### DIFF
--- a/lib/capybara/server/animation_disabler.rb
+++ b/lib/capybara/server/animation_disabler.rb
@@ -44,7 +44,11 @@ module Capybara
       end
 
       DISABLE_MARKUP_TEMPLATE = <<~HTML
-        <script defer>(typeof jQuery !== 'undefined') && (jQuery.fx.off = true);</script>
+        <script defer="defer">
+        //<![CDATA[
+          (typeof jQuery !== 'undefined') && (jQuery.fx.off = true);
+        //]]>
+        </script>
         <style>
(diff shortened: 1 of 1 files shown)


twalpole:
Hmmm -- I thought this should be something like

    <script defer>
    //<![CDATA[
    ...code...
    //]]>
   </script>

So I'm confused why the tests pass

dmorgan-fa:
Hey, so I took another pass at this, and think I've resolved the issues. If we continue to pass the `defer` tag without a value, the XML parser still fails (see [here](https://monkeyraptor.johanpaul.net/2015/11/fixing-xhtml-parsing-error-on-chrome.html) and screenshot below).

image

I think you're right that the previous PR wasn't doing what was expected.
I went back and checked, and it wasn't successfully injecting the script, so your question about why the tests were passing seems pretty valid.

I've updated the PR and checked that the script is present in the DOM when `disable_animation` is enabled, and it doesn't trigger the parsing error.

twalpole:
Thank you.

PR #2434, opened 2021-01-06
Add rspec driver selection info to README.md

micdahl:
I was wondering, why `Capybara.default_driver` and `Capybara.javascript_driver` had no effect in my rspec tests. It took me much time to find a way to select the driver for system tests, as I did not find this documented anywhere. I hope others may profit from this additional documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,21 @@ describe 'some stuff which requires js', js: true do
 end
 ```
 
+To permanently select a driver, you can add the following to your `rspec_helper.rb`:
+
+```ruby
+...
+RSpec.configure do |config|
+  ...
+  # select driver for each system spec.
+  # if you use feature specs use type: :feature
+  config.before(:each, type: :system) do
+    # see Drivers section for supported drivers
+    driven_by(:selenium_chrome_headless)
+  end
+end
+```
+
 Capybara also comes with a built in DSL for creating descriptive acceptance tests:
 
 ```ruby


luke-hill:
A better option would be just to do `before { Capybara.use_default_driver }` -> This is what is included in `capybara/cucumber` in the hooks and it makes a lot of sense. Then just set your default driver and hey presto!

micdahl:
> A better option would be just to do `before { Capybara.use_default_driver }` -> This is what is included in `capybara/cucumber` in the hooks and it makes a lot of sense. Then just set your default driver and hey presto!

This unfortunately does not work for me. I have set `Capybara.default_driver = :selenium_chrome_headless` and `Capybara.javascript_driver = :selenium_chrome_headless`. Then I used `before ... Capybara.use_default_driver` but the correct driver is not used and my tests fail.

luke-hill:
That sounds.... wrong, it would be good for you to do some debugging, because that's the intended behaviour. I would try work out where your code is changing the default driver (Because it's likely if a before each block isn't working, then something's weird).

micdahl:
I have created a brand new rails project for testing purposes. I have uploaded a repository: [https://github.com/micdahl/capybara_driver_selection_test](https://github.com/micdahl/capybara_driver_selection_test).
In the branch `driven_by`, `driven_by` is used to select the driver. The branch `use_default_driver` uses `Capybara.use_default_driver` for the same purpose. There are no other differences. `driven_by` does the job, `use_default_driver` fails to do so.

twalpole:
These instructions are only valid if you're using Capybara with `rails` and the `rspec-rails` gem, so the section they've been added to in the docs isn't really correct.  IMHO these belong more in the `rspec-rails` gems docs.

PR #2440, opened 2021-01-22
Fix js animation not disable if load jquery in body

kuei0221:
Animation disabler now can only turn off the animation come from jQuery if jQuery is loaded in head, because `insert_disable` place the template at the end of head.
https://github.com/teamcapybara/capybara/blob/a08e0477ac2cfd3319d8a423d8661b73103acbf3/lib/capybara/server/animation_disabler.rb#L42-L44

If jQuery is loaded with the body, jQuery will be undefined in head so the function won't work.
https://github.com/teamcapybara/capybara/blob/a08e0477ac2cfd3319d8a423d8661b73103acbf3/lib/capybara/server/animation_disabler.rb#L46-L49

By placing it to the end of the body, developer do not have to worry about where they are loading jQuery and give a better experience 😄

### DIFF
--- a/spec/shared_selenium_session.rb
+++ b/spec/shared_selenium_session.rb
@@ -411,6 +411,13 @@ RSpec.shared_examples 'Capybara::Session' do |session, mode|
           JS
           expect(scroll_y).to eq 500
         end
+
+        it 'should scroll the page instantly without jquery animation', requires: [:js] do
+          @animation_session.visit('with_jquery_animation')
+          @animation_session.click_link('scroll top 500')
+          scroll_y = @animation_session.evaluate_script('window.scrollY')
+          expect(scroll_y).to eq 500
+        end
       end
 
       context 'when set to `false`' do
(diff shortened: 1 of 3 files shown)


spec/shared_selenium_session.rb
```diff
@@ -405,7 +405,9 @@
           @animation_session.visit('with_animation')
           scroll_y = @animation_session.evaluate_script(<<~JS)
             (function(){
-              window.scrollTo(0,500);
+              $('html, body').animate({
```
twalpole, line 408: @Why are you modifying this test?

lib/capybara/spec/views/with_animation.erb
```diff
@@ -77,6 +63,20 @@
     <a href='#' class='animation' onclick='this.classList.add("away")' oncontextmenu='this.classList.add("pseudo")'>
       animate me away
     </a>
+    <script src="/jquery.js" type="text/javascript" charset="utf-8"></script>
```
twalpole, line 66: If a new view is needed for new tests please create it rather than modifying the existing one.

twalpole (changes requested):
I believe the original intent was for `defer` to handle this use case, but that isn't happening because `defer` is only for external scripts (ignored for inline).  Please remove the `defer` attribute along when making fixes for this PR

lib/capybara/spec/views/with_animation.erb
kuei0221: 🆗

spec/shared_selenium_session.rb
kuei0221: 🆗

twalpole:
LGTM - thanks

Issue #2442, opened 2021-01-26
Incorrect browser with :selenium_chrome and :selenium_chrome_headless


robinetmiller:
Looks like version 3.35.0 broke the builtin browser configurations a bit. `:selenium_chrome` and `:selenium_chrome_headless` now produce a non-headless Firefox selenium instance.

The changes to `lib/capybara/registrations/drivers.rb`, lines 28 and 41 in commit https://github.com/teamcapybara/capybara/commit/0ad261ad0a9771d2f3ee45555c34caf7e79c563e  appear to be the culprit.

twalpole:
Fixed in 3.35.1

PR #2444, opened 2021-01-26
Use chrome browser in chrome drivers

yuriyalekseyev:
Hey! 
It seems, that 0ad261ad0a9771d2f3ee45555c34caf7e79c563e accidentally changed browser from `chrome` to `firefox` in `:selenium_chrome` and `:selenium_chrome_headless` drivers


lunks:
This fixes #2442

twalpole:
Sorry about that, looks like copy paste bug and we might need some extra tests.  Thanks for this, I'll release a 3.35.1 as soon as the other reported issue is fixed

PR #2445, opened 2021-01-27
Correct year in Release dates [skip ci]

p8:


twalpole:
Thank you - one would think I should know 2020 was finally over

p8:
😄

PR #2468, opened 2021-05-28
Add matrix as a dependency for Ruby 3.1 compatibility

casperisfine:
It was recently removed from the default gems:

  - https://bugs.ruby-lang.org/issues/17873
  - https://github.com/ruby/ruby/pull/4530

### DIFF
--- a/capybara.gemspec
+++ b/capybara.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
               'such as Rails, Sinatra or Merb'
 
   s.add_runtime_dependency('addressable')
+  s.add_runtime_dependency('matrix')
   s.add_runtime_dependency('mini_mime', ['>= 0.1.3'])
   s.add_runtime_dependency('nokogiri', ['~> 1.8'])
   s.add_runtime_dependency('rack', ['>= 1.6.0'])


casperisfine:
Actually this cause double loading warnings for ruby 3.0 and older. I'm not too sure yet how best to fix this.

twalpole:
I haven’t looked in detail yet, but I  can’t think of anywhere that Capybara uses matrix itself. That would imply it’s actually a requirement of a dependency and should be fixed there rather than in Capybara

casperisfine:
> I can’t think of anywhere that Capybara uses matrix itself.

Right here: https://github.com/teamcapybara/capybara/blob/312ab56ac488d872986701fddfeeae120e308fbd/lib/capybara/queries/selector_query.rb#L3

twalpole:
Ah, yes -- we use Vector which is provided by the matrix library

twalpole:
If this is causing double loading warnings on released versions of Ruby then I'm not going to merge it in to fix a dev version of Ruby. For now you'll just need to specify the dependency in your own gemfile

casperisfine:
> If this is causing double loading warnings on released versions of Ruby then I'm not going to merge it

In the end it doesn't, the problem is only with specific gems that bundler use like `digest`, `net/protocol`. See: https://bugs.ruby-lang.org/issues/17873

I got this issue with the `mail` gem, so I posted here quickly too to make sure this wouldn't be merged before I investigate, but I'm confident this can be merged safely now.

> For now you'll just need to specify the dependency in your own gemfile

Yeah it's already worked around on our side, but I'm running ruby-head CI specifically to catch this kind of things early so that upgrading to Ruby 3.1 will as smooth as possible in January. So IMHO if this is included early it's better.

PR #2470, opened 2021-05-28
Remove useless require "matrix"

casperisfine:
Fix: https://github.com/teamcapybara/capybara/pull/2468

It seems that this `require` is useless, there's no reference at all to `Matrix` in `capybara`.


twalpole:
Vector is provided by the matrix library, so can't remove it

PR #2479, opened 2021-06-25
Remove test files from gemspec build

alexeyschepin:
I have a suggestion to delete test files from the gem build. Test files have been removed from many other gems rubygems/rubygems#735

By removing test files from the gem, it becomes a bit lighter (from 364Kb to 323Kb)
It will also stop annoying image screening tools for containing a private key.
```
key.pem | /usr/local/bundle/gems/capybara-3.35.3/spec/fixtures/key.pem
```


twalpole:
The test files are intentionally left in the gem for potential use by driver authors

Issue #2476, opened 2021-06-15
page.scroll_to :top not working


eric-hemasystems:
## Meta
Capybara Version: 3.35.3
Driver Information (and browser if relevant): selenium-webdriver 3.142.7 with Chrome 91

## Expected Behavior

```ruby
page.scroll_to :top
```

Scrolls to top of page.

## Actual Behavior

Stays at current scroll position. This workaround does work:

```ruby
page.execute_script 'window.scrollTo(0, 0)'
```

## Other Notes

I noticed that [`Capybara::Document#scroll_to` operates on the `body` element](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/node/document.rb#L43-L45). I found [this StackOverflow question](https://stackoverflow.com/questions/19635188/why-is-body-scrolltop-deprecated) that indicates calling scroll on the body element may be deprecated in certain situations. I'm wondering if an element is not being operates on if the JS generated should operate on `window` instead of `document.body`?

I can investigate this more and push up a PR to try to fix. But before I did I wanted to run this by the team here to see if someone here might have a bit more info or direction.

twalpole:
The tests for this functionality are in https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/session/scroll_spec.rb - specifically https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/session/scroll_spec.rb which all pass with Chrome 91 for me.  If you can provide a test that fails we'd have somewhere to start.

You can run the tests via something like

    WEBDRIVERS=true bundle exec rake spec_chrome

to run just the scrolling tests easiest way is to add `focus_: true` metadata to line 3 in that test file

eric-hemasystems:
K, thanks for the pointers. I'll see if I can create a re-producible test case. If I can I'll follow up with that. If I cannot then I'll just close this ticket down. Mainly wanted to see if I even got any response by opening a ticket because I can't tell you how many times I've worked up a PR only to have it just sit there for years until it is out of date and I'm not even using the project any more.

twalpole:
I'm going to close this for now, if you create a reproducible example I will reopen and investigate

eric-hemasystems:
@twalpole - Sorry for the radio silence, I got sidetracked but getting back to this. I was able to reproduce pretty easily based on the information in that stack overflow comment above.

[...]

PR #2483, opened 2021-07-22
Test in Standards Mode && Fix Page Scroll

eric-hemasystems:
This is a followup to issue #2476.

The primary goal was to fix page scroll when the page operates in standards mode. Per the discussion on that issue all test pages were all updated to use standards mode to confirm there are no other standard vs quirks mode issues in Capybara.

Commit f20c8de is just that conversion. See the commit notes for more information about that. Commit b2e2a27 fixes the main problem referred to in #2476.

I have also verified that the application I am developing that was attempting to use `page.scroll_to :top` is now able to use that method successfully without any sort of `execute_script` hack.

### DIFF
--- a/lib/capybara/node/document.rb
+++ b/lib/capybara/node/document.rb
@@ -41,7 +41,7 @@ module Capybara
       end
 
       def scroll_to(*args, **options)
-        find(:xpath, '//body').scroll_to(*args, **options)
+        find(:xpath, '/html').scroll_to(*args, **options)
       end
     end
   end
--- a/lib/capybara/spec/session/scroll_spec.rb
+++ b/lib/capybara/spec/session/scroll_spec.rb
@@ -15,7 +15,7 @@ Capybara::SpecHelper.spec '#scroll_to', requires: [:scroll] do
     el = @session.find(:css, '#scroll')
     @session.scroll_to(el, align: :bottom)
     el_bottom = el.evaluate_script('this.getBoundingClientRect().bottom')
-    viewport_bottom = el.evaluate_script('document.body.clientHeight')
+    viewport_bottom = el.evaluate_script('document.documentElement.clientHeight')
     expect(el_bottom).to be_within(1).of(viewport_bottom)
   end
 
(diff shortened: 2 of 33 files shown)


twalpole:
Why update so many of the view erbs if it's being handled in the test app ?

eric-hemasystems:
Some of the existing ERB files already had a layout (were not just a HTML fragment). For example, [the scroll one](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/views/scroll.erb). I'm presuming if it already has a layout then it has specific needs such as setting a title, specific styles, etc. In the interest of making the least disturbance to the test suite, the only update to those was to add a proper doctype.

Other templates are only a HTML fragment such as [the button one](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/spec/views/buttons.erb). These need not just a doctype but also the proper HTML boilerplate (html tag, body tag, etc). Rather than adding all that to each ERB, I created a layout file that all these HTML fragements use. This way if we need to adjust the boilerplate later it is fast to do so.

I am guessing there are some templates that already had a layout but could probably use the common one. But doing that evaluation for each template was beyond the scope of the work I was looking to do on this. The test app looks at the first line of the template. If it has a doctype then it assumes the template comes with its own layout and therefore renders it without a layout. If no doctype is found then it presumes it is a HTML fragment and renders it with the common layout.

The comment on the commit message of f20c8de in the "Implementation" section has more info on all this. If you prefer to do things differently feel free to give direction on that. This was just the simplest I came up with that modified the test suite the least while still ensuring all templates are rendered in standards mode.

I haven't counted but probably a good half of the templates had a existing layout so those were updated the add the doctype. The other half were HTML fragments and are now just using the common layout and therefore were not updated.

If all this seems like too much we could just make the scroll template render in standards mode as I didn't find any other failures due to the mode switch. Leave the others are in quirks mode. Personally I think having them all in standards mode is better as that is the mode people using Capybara are likely in so it makes the test environment more closely match the real world usage.

twalpole:
Thanks for catching and fixing this -- I added in support for quirks mode just in case

PR #2489, opened 2021-08-14
Support `Session#active_element`

seanpdoyle:
As a [follow-up][] to [#2422][], expose the [document.activeElement][]
(that is, the element with focus) via the `Session`.

When called from Rack Test environments, `Session#active_element` raises
an error.

[Document.activeElement]: https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement
[follow-up]: https://github.com/teamcapybara/capybara/pull/2422#discussion_r520646573
[#2422]: https://github.com/teamcapybara/capybara/pull/2422
[document.activeElement]: https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement

### DIFF
--- a/History.md
+++ b/History.md
@@ -9,6 +9,7 @@ Release date: unreleased
 
 * `allow_label_click` accepts click options to be used when clicking an associated label
 * Deprecated `allow_gumbo=` in favor of `use_html5_parsing=` to enable use of Nokogiri::HTL5 when available
+* `Session#active_element` returns the element with focus - Not supported by the `RackTest` driver [Sean Doyle]
 
 ### Fixed
 
--- a/lib/capybara/selenium/driver.rb
+++ b/lib/capybara/selenium/driver.rb
@@ -148,8 +148,12 @@ class Capybara::Selenium::Driver < Capybara::Driver::Base
     unwrap_script_result(result)
   end
 
+  def active_element
+    build_node(browser.switch_to.active_element)
+  end
+
   def send_keys(*args)
-    active_element.send_keys(*args)
+    active_element.native.send_keys(*args)
   end
 
   def save_screenshot(path, **_options)
@@ -472,10 +476,6 @@ private
 
   def find_context
     browser
-  end
-
-  def active_element
-    browser.switch_to.active_element
   end
 
   def build_node(native_node, initial_cache = {})
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -313,6 +313,16 @@ module Capybara
 
     ##
     #
+    # Returns the element with focus.
+    #
+    # Not supported by Rack Test
+    #
+    def active_element
+      driver.active_element
+    end
+
+    ##
+    #
     # Executes the given block within the context of a node. {#within} takes the
     # same options as {Capybara::Node::Finders#find #find}, as well as a block. For the duration of the
     # block, any command to Capybara will be handled as though it were scoped
--- a/lib/capybara/driver/base.rb
+++ b/lib/capybara/driver/base.rb
@@ -63,6 +63,10 @@ class Capybara::Driver::Base
     raise Capybara::NotSupportedByDriverError, 'Capybara::Driver::Base#send_keys'
   end
 
+  def active_element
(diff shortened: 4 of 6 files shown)


History.md
```diff
@@ -9,6 +9,8 @@ Release date: unreleased
 
 * `allow_label_click` accepts click options to be used when clicking an associated label
 * Deprecated `allow_gumbo=` in favor of `use_html5_parsing=` to enable use of Nokogiri::HTL5 when available
+* In Selenium-enabled environments, `Session#active_element` returns the element with focus (via [document.activeElement][]).
```
twalpole, line 12: This should just be for Selenium-enabled environments - there are other drivers which will end up supporting it.  Just say `Session#active_element` returns the element with focus - Not supported by the RackTest driver`

lib/capybara/session.rb
```diff
@@ -311,6 +311,14 @@ def send_keys(*args, **kw_args)
       driver.send_keys(*args, **kw_args)
     end
 
+    ##
+    # @!method active_element
+    #   @see Capybara::Node::Element#active_element
```
twalpole, line 316: There is no Element#active_element

spec/shared_selenium_session.rb
```diff
@@ -234,6 +234,19 @@
       end
     end
 
+    describe '#active_element', requires: [:js] do
```
twalpole, line 237: Lets move this into the non-selenium specific specs - something like lib/capybara/spec/session/active_element_spec.rb

twalpole: Also, please add relevant tabindex attributes in the form since the test depends on the tabbing order.

lib/capybara/selenium/driver.rb
```diff
@@ -148,6 +148,10 @@ def evaluate_async_script(script, *args)
     unwrap_script_result(result)
   end
 
+  def active_element
+    build_node(browser.switch_to.active_element)
```
twalpole, line 152: Technically this changes the behavior of Session#send_keys when using the selenium driver because `build_node` will build instances specialized for the specific browser being used - which may not end up directly calling the Selenium `send_keys` directly - was this intentional? (it may actually be an improvement, just checking)

twalpole (changes requested):

spec/shared_selenium_session.rb
seanpdoyle: The current tests assertions are about the first field that is not marked with `[disabled]`.

> Also, please add relevant tabindex attributes in the form since the test depends on the tabbing order.

What did you have in mind?

lib/capybara/selenium/driver.rb
seanpdoyle: Not intentional. Calling `build_node` was an attempt to transform an instance of `Selenium::WebDriver::Element` into an instance of `Capybara::Node::Element` for the equivalence check in the test suite. Is there a better way or a different abstraction to use?

twalpole: No -- just wanted to make sure you realized that `build_node` would return a ::Capybara::Selenium::ChromeNode instance when using the selenium driver with Chrome, which has some extra code to work around an issue with emoji in send_keys  -- not necessarily a bad thing just different than the previous behavior

lib/capybara/selenium/driver.rb
```diff
@@ -148,8 +148,12 @@ def evaluate_async_script(script, *args)
...
+  end
+
   def send_keys(*args)
-    active_element.send_keys(*args)
+    active_element.native.send_keys(*args)
```
seanpdoyle, line 156: I made this change before https://github.com/teamcapybara/capybara/pull/2489/files#r689119285.

Should we revert it so that the `send_keys` goes through the specialized nodes?

seanpdoyle:
It's worth mentioning that I've opened https://github.com/teamcapybara/capybara/pull/2490 as a follow-up PR. Unfortunately, GitHub won't let me declare the branch off of `seanpdoyle:active-element`, so it also contains commits from this branch.

PR #2490, opened 2021-08-15
Add support for `focused:` filter

seanpdoyle:
[Follow-up to #2489][], adds support for filtering fields, links, and
buttons based on whether or not they're focused.
    
[Follow-up to #2489]: https://github.com/teamcapybara/capybara/pull/2489

### DIFF
--- a/lib/capybara/queries/selector_query.rb
+++ b/lib/capybara/queries/selector_query.rb
@@ -9,7 +9,7 @@ module Capybara
 
       SPATIAL_KEYS = %i[above below left_of right_of near].freeze
       VALID_KEYS = SPATIAL_KEYS + COUNT_KEYS +
-                   %i[text id class style visible obscured exact exact_text normalize_ws match wait filter_set]
+                   %i[text id class style visible obscured exact exact_text normalize_ws match wait filter_set focused]
       VALID_MATCH = %i[first smart prefer_exact one].freeze
 
       def initialize(*args,
@@ -73,6 +73,8 @@ module Capybara
 
         desc << " with id #{options[:id]}" if options[:id]
         desc << " with classes [#{Array(options[:class]).join(',')}]" if options[:class]
+        desc << ' that is focused' if options[:focused]
+        desc << ' that is not focused' if options[:focused] == false
 
         desc << case options[:style]
         when String
@@ -435,6 +437,7 @@ module Capybara
           matches_id_filter?(node) &&
           matches_class_filter?(node) &&
           matches_style_filter?(node) &&
+          matches_focused_filter?(node) &&
           matches_text_filter?(node) &&
           matches_exact_text_filter?(node)
       end
@@ -491,6 +494,16 @@ module Capybara
           options[:class].select { |c| c.is_a? Regexp }.all? do |r|
             classes.any? { |cls| r.match? cls }
           end
+        end
+      end
+
+      def matches_focused_filter?(node)
+        if options.key?(:focused)
+          node_is_focused = node == node.session.active_element
+
+          node_is_focused == options[:focused]
+        else
+          true
         end
       end
 
--- a/lib/capybara/selector.rb
+++ b/lib/capybara/selector.rb
@@ -14,6 +14,7 @@ require 'capybara/selector/definition'
 #   * :left_of (Element) - Match elements left of the passed element on the page
 #   * :right_of (Element) - Match elements right of the passed element on the page
 #   * :near (Element) - Match elements near (within 50px) the passed element on the page
+#   * :focused (Boolean) - Match elements with focus (requires JavaScript)
 #
 # ### Built-in Selectors
 #
--- a/lib/capybara/spec/session/has_button_spec.rb
+++ b/lib/capybara/spec/session/has_button_spec.rb
@@ -73,8 +73,8 @@ Capybara::SpecHelper.spec '#has_button?' do
       expect(@session).to have_button('A Button', focused: true)
     end
 
-    it 'should be false if a field does not have focus when focused: false' do
-      expect(@session).not_to have_button('A Button', focused: false)
+    it 'should be true if a field does not have focus when focused: false' do
+      expect(@session).to have_button('A Button', focused: false)
     end
   end
 end
@@ -131,14 +131,14 @@ Capybara::SpecHelper.spec '#has_no_button?' do
   end
 
   context 'with focused:', requires: [:js] do
-    it 'should be true if a field has focus when focused: true' do
-      expect(@session).to have_no_button('A Button', focused: false)
+    it 'should be true if a button does not have focus when focused: true' do
+      expect(@session).to have_no_button('A Button', focused: true)
     end
 
-    it 'should be false if a field does not have focus when focused: false' do
+    it 'should be false if a button has focus when focused: false' do
       @session.send_keys(:tab)
 
-      expect(@session).to have_no_button('A Button', focused: true)
+      expect(@session).to have_no_button('A Button', focused: false)
     end
   end
 end
--- a/lib/capybara/spec/session/has_button_spec.rb
+++ b/lib/capybara/spec/session/has_button_spec.rb
@@ -65,6 +65,18 @@ Capybara::SpecHelper.spec '#has_button?' do
   it 'should not affect other selectors when enable_aria_role: false' do
     expect(@session).to have_button('Click me!', enable_aria_role: false)
   end
+
+  context 'with focused:', requires: [:js] do
+    it 'should be true if a field has focus when focused: true' do
+      @session.send_keys(:tab)
+
+      expect(@session).to have_button('A Button', focused: true)
+    end
+
+    it 'should be false if a field does not have focus when focused: false' do
+      expect(@session).not_to have_button('A Button', focused: false)
+    end
+  end
 end
 
 Capybara::SpecHelper.spec '#has_no_button?' do
@@ -116,5 +128,17 @@ Capybara::SpecHelper.spec '#has_no_button?' do
 
   it 'should not affect other selectors when enable_aria_role: false' do
     expect(@session).to have_no_button('Junk button that does not exist', enable_aria_role: false)
+  end
+
+  context 'with focused:', requires: [:js] do
+    it 'should be true if a field has focus when focused: true' do
+      expect(@session).to have_no_button('A Button', focused: false)
+    end
+
+    it 'should be false if a field does not have focus when focused: false' do
+      @session.send_keys(:tab)
+
+      expect(@session).to have_no_button('A Button', focused: true)
+    end
   end
 end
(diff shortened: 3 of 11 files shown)


lib/capybara/selector/definition/button.rb
```diff
@@ -40,6 +40,8 @@
     builder(xpath).add_attribute_conditions(name: val)
   end
 
+  node_filter(:focused) { |node, value| (node.native == node.session.active_element.native) == value }
```
seanpdoyle, line 43: @twalpole this is a duplicate of the `_field` filter set.

In practice, _any_ element with a `[tabindex]` attribute can have focus. Is there a way to declare a global filter?

lib/capybara/selector.rb
```diff
@@ -186,6 +194,7 @@ class Capybara::Selector; end # rubocop:disable Lint/EmptyClass
       add_error("Expected validation message to be #{msg.inspect} but was #{vm}") unless res
     end
   end
+  node_filter(:focused) { |node, value| (node.native == node.session.active_element.native) == value }
```
seanpdoyle, line 197: Is there a way to detect or disabled a `node_filter` declaration when JavaScript is unavailable? This could either no-op, always return `false`, or `raise` and error from Rack Test. What options do we have?

seanpdoyle:
I'm not sure why CI isn't running. Is there an extra step I need to take?

twalpole:
Need to move the project off Travis but haven't had time

lib/capybara/selector/definition/button.rb
twalpole: global filters are defined in SelectorQuery - see `matches_id_filter?`, ``matches_class_filter?`, etc

twalpole: Although it might be better to just add it to a fieldset and them import it into the different selectors.

lib/capybara/selector.rb
twalpole: It should already raise an error on the calling of `active_element` if the user users the `focused` filter.  Also any reason you're calling `native` here?

twalpole: `node_filter(:focused, :boolean)...`

twalpole:
Please add tabindex attributes to all relevant form fields.  The tests depend on tabbing order,  adding tabindex attributes will make sure the order stays even if the forms are modified in the future.

lib/capybara/selector.rb
seanpdoyle: > Also any reason you're calling native here?

Comparing the `node` and `node.session.active_element` is evaluating to false, since `node` is an instance of `Capybara::Node::Element`, and `node.session.active_element` evaluates to a Selenium-specialized instance.

Comparing the underlying native nodes evaluates to true.

seanpdoyle: > It should already raise an error on the calling of active_element if the user users the focused filter. Also any reason you're calling native here?

From a developer's perspective, always raising makes a lot of sense. I'm concerned about CI and our test suite failing. Does this diff properly mark the tests to only execute for JS-enabled browsers?

seanpdoyle: I've pushed up a change to add a `matches_focused_filter?` to `SelectorQuery`. Having a global filter was most appealing, but that means it's run on every query for every driver. Does a `SelectorQuery` instance have access to a mechanism to detect JavaScript? Would a `node_filter(:focused) { ... }` block?

twalpole: I fixed the code I merged around active_element into master to return an Element instance so the `.native` shouldn't be necessary anymore

lib/capybara/spec/session/has_button_spec.rb
```diff
@@ -65,6 +65,18 @@
   it 'should not affect other selectors when enable_aria_role: false' do
     expect(@session).to have_button('Click me!', enable_aria_role: false)
   end
+
+  context 'with focused:', required: [:js] do
```
twalpole, line 69: requires: [:js]

lib/capybara/spec/session/has_button_spec.rb
```diff
@@ -117,4 +129,16 @@
   it 'should not affect other selectors when enable_aria_role: false' do
     expect(@session).to have_no_button('Junk button that does not exist', enable_aria_role: false)
   end
+
+  context 'with focused:', required: [:js] do
```
twalpole, line 133: requires

spec/rack_test_spec.rb
```diff
@@ -12,6 +12,7 @@ module TestSessions
   screenshot
   frames
   windows
+  active_element
```
twalpole, line 15: This would only make a difference if you did `requires: [:active_element]` instead of :js on the tests

twalpole (changes requested):

lib/capybara/queries/selector_query.rb
```diff
@@ -494,6 +497,12 @@ def matches_class_filter?(node)
         end
       end
 
+      def matches_focused_filter?(node)
```
twalpole, line 500: This should return true if no focused option was passed -- see `matches_id_filter?` and `use_default_id_filter?`

lib/capybara/selector.rb
```diff
@@ -41,6 +41,7 @@
 #       * :multiple (Boolean) - Match fields that accept multiple values
 #       * :valid (Boolean) - Match fields that are valid/invalid according to HTML5 form validation
 #       * :validation_message (String, Regexp) - Matches the elements current validationMessage
+#       * :focused (Boolean) - Match fields with focus
```
twalpole, line 44: these can be mentioned at the top with the other global filters

lib/capybara/selenium/driver.rb
```diff
@@ -148,8 +148,12 @@ def evaluate_async_script(script, *args)
     unwrap_script_result(result)
   end
 
+  def active_element
```
twalpole, line 151: Please rebase off master

twalpole (changes requested):

lib/capybara/spec/session/has_button_spec.rb
```diff
@@ -65,6 +65,18 @@
...
+
+  context 'with focused:', requires: [:js] do
+    it 'should be true if a field has focus when focused: true' do
+      @session.send_keys(:tab)
+
+      expect(@session).to have_button('A Button', focused: true)
+    end
+
+    it 'should be false if a field does not have focus when focused: false' do
+      expect(@session).not_to have_button('A Button', focused: false)
```
twalpole, line 77: Assuming the above test is correct and tab has to be sent to focus the button, then this test doesn't seem right to me - if the button isn't focused then it should match and it should be `expect(@session).to have_button('A button', focused: false)`

twalpole:
The logic of some of the tests is wrong, and addition of new elements with conflicting tabindex has broken existing tests - 
```
(husky) ➜  capybara git:(focused-filter) HEADLESS=true WEBDRIVERS=true bundle exec rake spec_chrome
/Users/twalpole/.rbenv/versions/3.0.2/bin/ruby -I/Users/twalpole/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rspec-core-3.10.1/lib:/Users/twalpole/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rspec-support-3.10.2/lib /Users/twalpole/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rspec-core-3.10.1/exe/rspec --pattern ./spec/\{selenium_spec_chrome.rb\} --color
Capybara starting Puma...
* Version 5.4.0 , codename: Super Flight
* Listening on http://127.0.0.1:52678
Run options:
  include {:focus_=>true}
  exclude {:requires=>#<Proc:>}

All examples were filtered out; ignoring {:focus_=>true}

Randomized with seed 15508
..........................................................................................................................*..........................
  1) Capybara::Session with chrome with selenium driver #send_keys defaults to sending keys to the document.activeElement
     Failure/Error: expect(session.evaluate_script('document.activeElement')).to eq(session.first(:field))

       expected: #<Capybara::Node::Element tag="input" path="/HTML/BODY[1]/LABEL[1]/INPUT[1]">
            got: #<Capybara::Node::Element tag="button" path="/HTML/BODY[1]/BUTTON[1]">

       (compared using ==)

       Diff:
       @@ -1 +1 @@
       -#<Capybara::Node::Element tag="input" path="/HTML/BODY[1]/LABEL[1]/INPUT[1]">
       +#<Capybara::Node::Element tag="button" path="/HTML/BODY[1]/BUTTON[1]">

     Shared Example Group: "Capybara::Session" called from ./spec/selenium_spec_chrome.rb:94
     # ./spec/shared_selenium_session.rb:233:in `block (4 levels) in <top (required)>'
FCapybara starting Puma...
* Version 5.4.0 , codename: Super Flight
* Listening on http://127.0.0.1:52737
..Capybara starting Puma...
* Version 5.4.0 , codename: Super Flight
* Listening on http://127.0.0.1:52755
..Capybara starting Puma...
* Version 5.4.0 , codename: Super Flight
* Listening on http://127.0.0.1:52768
.....Capybara starting Puma...
* Version 5.4.0 , codename: Super Flight
* Listening on http://127.0.0.1:52781
........**..................................................................................................................................................................................................................Capybara starting Puma...
* Version 5.4.0 , codename: Super Flight
* Listening on http://127.0.0.1:52869
..**............................................................
  2) Capybara::Session selenium_chrome #active_element should return the active element
     Failure/Error: expect(@session.active_element).to match_selector(:css, '#form_title')
       Item does not match the provided selector
     # ./lib/capybara/spec/session/active_element_spec.rb:8:in `block (2 levels) in <top (required)>'
F.....................................................................................................................................................................................
  3) Capybara::Session selenium_chrome #has_no_field with focused should be true if a field does not have focus
     Failure/Error: expect(@session).to have_field('An Input', focused: true)
       expected to find visible field "An Input" that is focused that is not disabled but there were no matches. Also found "", which matched the selector but not all filters.
     # ./lib/capybara/spec/session/has_field_spec.rb:202:in `block (3 levels) in <top (required)>'
F.....................Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead. /Users/twalpole/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rspec-core-3.10.1/exe/rspec:4
..................................Capybara starting Puma...
```

seanpdoyle:
Thank you for serving as CI, @twalpole.

I've addressed the failures you mentioned and have pushed up changes.

I've run `HEADLESS=true WEBDRIVERS=true bundle exec rake spec_chrome` locally, and all the tests passed!

```
Finished in 11 minutes 27 seconds (files took 3.8 seconds to load)
1704 examples, 0 failures, 5 pending

Randomized with seed 45070
```

PR #2491, opened 2021-08-15
Support `described_by:` filter

seanpdoyle:
Filter fields, buttons, and links by descriptions constructed from their
[aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute) attribute.

For example, given the following HTML:

```html
<label id="price">Price</label>
<span id="price_currency">USD</price>
<input id="price" type="number" name="price" value="-1" aria-describedby="price_currency price_errors">
<span id="price_errors">must be greater than $0</span>
```

consider the following:

```ruby
expect(page).to have_field(described_by: "USD")
expect(page).to have_field("USD must be greater than $0")
expect(page).to have_field("Price", described_by: "USD")
expect(page).to have_field("Price", described_by: "must be greater than $0")
expect(page).to have_field("Price", described_by: "USD must be greater than $0")
```

### DIFF
--- a/History.md
+++ b/History.md
@@ -9,6 +9,7 @@ Release date: unreleased
 
 * `allow_label_click` accepts click options to be used when clicking an associated label
 * Deprecated `allow_gumbo=` in favor of `use_html5_parsing=` to enable use of Nokogiri::HTL5 when available
+* Filter fields, buttons, and links by descriptions constructed from their `aria-describedby` attribute [Sean Doyle]
 
 ### Fixed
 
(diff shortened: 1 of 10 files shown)


twalpole:
I don't really want to add any more aria stuff to Capybara at the moment.  Aria basically allows any element to be anything and makes trying to write efficient selectors for everything nearly impossible.  IMO this should be functionality in a separate gem that uses Capybara.modify_selector to add whatever aria based filters you want

PR #2494, opened 2021-08-25
Add description of Cuprite into README

YusukeIwaki:
Cuprite is already introduced into Rails.
https://github.com/rails/rails/pull/42790

Some Rails users will refer the README of Capybara, and it would be kind to include the guide of Cuprite in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ GitHub): http://groups.google.com/group/ruby-capybara
     - [RackTest](#racktest)
     - [Selenium](#selenium)
     - [Apparition](#apparition)
+    - [Cuprite](#cuprite)
 - [The DSL](#the-dsl)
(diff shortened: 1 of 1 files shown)


guillaumewrobel:
Any blocker @twalpole ?

sdhull:
apparition seems cool but looks like cuprite has much more activity (more stars, more recent releases, etc). Seems like the readme _should_ point folks to cuprite rather than apparition. Any explanation as to why this was closed without any further discussion?

https://www.ruby-toolbox.com/compare/apparition,cuprite?display=table&order=score

YusukeIwaki:
https://github.com/teamcapybara/capybara/commit/0b43cb90ae633ce9b2008e07b997c76833fa570e
It seems Apparition is silently removed from the description.

PR #2495, opened 2021-08-27
Fix incompatibility with React 17 when using attach_file

trappar:
React 17 makes some [changes to how they handle event delegation](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation).

One result of these changes is that:

> If you add manual DOM listeners with `document.addEventListener(...)`, you might expect them to catch all React events. In React 16 and earlier, even if you call e.stopPropagation() in a React event handler, your custom document listeners would still receive them because the native event is already at the document level. With React 17, the propagation would stop (as requested!), so your document handlers would not fire

(see: https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues)

This is an issue with Capybara since `attach_file` (when used with a block) executes the following script to find the actual file input: https://github.com/teamcapybara/capybara/blob/3.20.2/lib/capybara/node/actions.rb#L395-L403

In my case I found the issue was present when using Capybara to test code which used `react-dropzone` due to its [event handling calling `event.stopPropagation`](https://github.com/react-dropzone/react-dropzone/blob/master/src/index.js#L746-L748).

React's docs suggest passing `{capture: true}` as the third argument to `addEventListener`. In my testing I found the following:

**When using `{capture: false}`** (current/default behavior)
* When using React 16 and calling `attach_file` - Test passes
* When using React 17 and calling `attach_file` - Test fails with error `Capybara was unable to determine the file input you're attaching to`

**When using `{capture: true}`** (PR's behavior)
* When using React 16 and calling `attach_file` - Test passes
* When using React 17 and calling `attach_file` - Test passes

I am not sure if enabling the capture flag on addEventListener could have other undesirable side effects, but at least according to the React docs it doesn't sound like it. Maybe someone who understands event delegation / bubbling better than I do could speak to this?

I hope that even if this PR isn't accepted, it at least shines some light on the issue!

### DIFF
--- a/lib/capybara/node/actions.rb
+++ b/lib/capybara/node/actions.rb
@@ -413,7 +413,7 @@ module Capybara
             this.removeEventListener('click', file_catcher);
             e.preventDefault();
           }
-        })
+        }, {capture: true})
       JS
     end
   end


twalpole:
Seems reasonable to me 👍

PR #2496, opened 2021-08-30
Remove include Capybara::DSL and include Capybara::RSpecMatchers

YusukeIwaki:
'system spec' and 'feature spec' are defined in rspec-**rails**, and it is better to include them in the context of Rails.

rspec-rails actually includes Capybara::DSL properly here,
https://github.com/rspec/rspec-rails/blob/v5.0.2/lib/rspec/rails/vendor/capybara.rb#L13-L27
so I think we don't need to include Capybara::DSL anymore.

(Note that non-Rails users will be affected if they use Capybara without `config.include Capybara::DSL`.)

### DIFF
--- a/lib/capybara/rspec.rb
+++ b/lib/capybara/rspec.rb
@@ -7,12 +7,6 @@ require 'capybara/rspec/features'
 require 'capybara/rspec/matcher_proxies'
 
 RSpec.configure do |config|
-  config.include Capybara::DSL, type: :feature
-  config.include Capybara::RSpecMatchers, type: :feature
-  config.include Capybara::DSL, type: :system
-  config.include Capybara::RSpecMatchers, type: :system
-  config.include Capybara::RSpecMatchers, type: :view
-
   # The before and after blocks must run instantaneously, because Capybara
   # might not actually be used in all examples where it's included.
   config.after do


twalpole:
This would be a breaking change and I see no real upside to it

YusukeIwaki:
@twalpole Hi, thank you for taking a look.
I know this change is no venefit for actual users, but I believe the change would be valuable for us (contributors and maintainers of Capybara).

Capybara is essentially not responsible for including Capybara::DSL, because it is properly included by rspec-rails and Systemtestcase in Rails when needed.
However, in current implementation Capybara is too kind, and it does too much business, which increases complexity of Capybara and results to essentially unnecessary implementation like this https://github.com/teamcapybara/capybara/commit/5b68043516451c3831fcc9833442a9feb4f2415d

If breaking change should be avoided, I have another solution with no breaking changes. (just adding `unless defined?(RSpec::Rails)` for backward compatibility)

I know you are so busy and I want to contribute for Capybara to reduce its complexity.
Is it acceptable if this PR comes with no breaking change?

twalpole:
@YusukeIwaki I'm not going to be removing code that some users may currently depend on without at least some concrete example of where it's causing any issues for anyone.  All this code is doing is setting some RSpec configuration - It's really not an increased complexity for anyone that I know of.  If anything, adding unless defined?(RSpec::Rails) just increases the complexity by adding more potential load order issues - so no this will not be acceptable without a concrete example of how this causes any issues.

YusukeIwaki:
Thank you for the detailed explanation.

I thought this https://github.com/teamcapybara/capybara/commit/5b68043516451c3831fcc9833442a9feb4f2415d is also not originated from user issues because rspec-rails is responsible for it. Also this commit produced undocumented behavior for non-Rails users when type: :system happens to be set.

I'm really confused when I studied  and tried to understand Capybara and rspec-rails, especially about the usage of include Capybara::DSL.


However, now I understand Capybara only accepts some improvement originated from user issues.

PR #2498, opened 2021-09-26
Fix invalid HTML generated by AnimationDisabler

javierm:
While I was experimenting with the [W3C rspec validations](https://github.com/Goltergaul/w3c_rspec_validators) gem, it reported all pages had invalid HTML due to the contents added by Capybara's AnimationDisabler. The W3C Validator was reporting an error:

> Element style not allowed as child of element body in this context.
>
> Contexts in which element style may be used:
>    * Where metadata content is expected.
>    * In a noscript element that is a child of a head element.
>
> Content model for element body:
>    * Flow content.

It looks like [there can't be any style tags inside the body element](https://html.spec.whatwg.org/multipage/semantics.html#the-style-element).

The `<script>` part of the markup template remains at the end of the `<body>` element because it only makes sense if jQuery has already been defined, and jQuery could be defined anywhere inside the `<head>` or `<body>` elements.

I'm not familiar with your source code at all, so comments about either the implementation or the tests are much appreciated :pray:.

### DIFF
--- a/lib/capybara/server/animation_disabler.rb
+++ b/lib/capybara/server/animation_disabler.rb
@@ -16,7 +16,10 @@ module Capybara
 
       def initialize(app)
         @app = app
-        @disable_markup = format(DISABLE_MARKUP_TEMPLATE, selector: self.class.selector_for(Capybara.disable_animation))
+        @disable_css_markup = format(DISABLE_CSS_MARKUP_TEMPLATE,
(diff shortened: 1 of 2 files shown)


twalpole:
Thanks for this --- While <style> tags are technically illegal outside the <head> element, all modern browsers support it. That said, I don't think it will hurt anything to make it technically legal - just want to think about any possible ramifications for a few days --- If I don't come up with any will merge by end of week

PR #2509, opened 2021-10-29
Document `server_host` option

pablobm:
I needed this option recently but didn't find it on the docs. I saw it was mentioned on `lib/capybara/session/config.rb`, but that's a less obvious location.

The description includes the detail of why you would want to change this, typically to `"0.0.0.0"`, as it strikes me as the most obvious use case. Thoughts?

### DIFF
--- a/lib/capybara.rb
+++ b/lib/capybara.rb
@@ -96,6 +96,7 @@ module Capybara
     #   {Capybara::Session#save_and_open_page save_and_open_page}, or {Capybara::Session#save_and_open_screenshot save_and_open_screenshot}.
     # - **server** (Symbol = `:default` (which uses puma)) - The name of the registered server to use when running the app under test.
     # - **server_port** (Integer) - The port Capybara will run the application server on, if not specified a random port will be used.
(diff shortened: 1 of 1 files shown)


lib/capybara.rb
```diff
@@ -96,6 +96,7 @@ class << self
     #   {Capybara::Session#save_and_open_page save_and_open_page}, or {Capybara::Session#save_and_open_screenshot save_and_open_screenshot}.
     # - **server** (Symbol = `:default` (which uses puma)) - The name of the registered server to use when running the app under test.
     # - **server_port** (Integer) - The port Capybara will run the application server on, if not specified a random port will be used.
+    # - **server_host** (Integer = "127.0.0.1") - The IP address Capybara will bind the application server to. If the test application is to be accessed from an external host, you will want to change this to "0.0.0.0" or to a more specific IP address that your test client can reach.
```
twalpole, line 99: This should be a String type, not Integer

twalpole (changes requested):

lib/capybara.rb
pablobm: Ooops. Fixed now.

PR #2519, opened 2021-12-11
add support for setting the server to ssl and skip the HTTP request

dorianmariefr:
It's to skip the HTTP requests that result in:

```
2021-12-11 11:58:11 +0100 SSL error, peer: 127.0.0.1, peer cert: : #<Puma::MiniSSL::SSLError: HTTP connection?>
```


twalpole:
Is the motivation for this only to get rid of the SSL Request log message?  If so just specify the Silent option -- I'd really prefer to start getting rid of some of Capybaras settings rather than adding more

twalpole:
Closing due to lack of response.  We can reopen if there is a real reason for this change.

dorianmariefr:
it's to get rid of the extra request, probably costing time

PR #2520, opened 2021-12-30
Allow selenium-webdriver 4.1.0 or higher

JunichiIto:
This PR fixes the following warning when running with selenium-webdriver 4.1.0.

```
WARN Selenium [DEPRECATION] [:browser_options] :options as a parameter for driver initialization is deprecated. Use :capabilities with an Array of value capabilities/options if necessary instead.
```


twalpole:
Thanks

PR #2527, opened 2022-02-01
Add description of `Capybara.server=(callable)`

donv:


twalpole:
Thanks for this, however server= taking a callable object will be removed in v4 so I don't really want to advertise it currently

PR #2553, opened 2022-05-20
Use #create not #make in README example

jalada:
Very minor point...happy to have it ignored.

These tests look a lot like they are testing a Rails app, but `User.make` is not standard Rails and just felt a little out of place to me. I looked back over the README's history and issues, discussion etc. and couldn't see that it was a conscious decision _not_ to make it look like Rails. So figured it might be less confusing to newbies to have it look like familiar Rails code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ You can now write your specs like so:
 ```ruby
 describe "the signin process", type: :feature do
   before :each do
-    User.make(email: 'user@example.com', password: 'password')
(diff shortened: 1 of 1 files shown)


twalpole:
IIRC `make` was a Machinist method, which was popular a while ago

jalada:
> IIRC `make` was a Machinist method, which was popular a while ago

You learn something new every day! 🙏🏻

Issue #2556, opened 2022-05-27
With rack test driver, unfilled file fields clobber hidden fields of same name, unlike real browser


bjeanes:
Hello :)

I have noticed a bug with rack test driver which is a significant departure from real browser behaviour when it comes to form submissions. At first, I thought this was an issue `rack-test` and opened https://github.com/rack/rack-test/issues/301 but further digging has shown the issue to be on the Capybara side.

The bug is that when there is a hidden field named `foo` with value `bar` followed by a file field also named `foo`, a form submitted without attaching a file does not submit `foo=bar` as a parameter.

This appears to be due to [`NilUploadedFile`](https://github.com/teamcapybara/capybara/blob/e145ec5ab7a9688c9a42077f3b981e753e981f17/lib/capybara/rack_test/form.rb#L4-L21), first introduced in 29898295dc7a65c6587026fac4b3991420f0378d, 12 years ago.

As best I can tell, this was to work around the fact that `rack-test` historically had no way to declare that the params should be encoded as multipart other than by sniffing for a `Rack::Test::UploadedFile` (from which `NilUploadedFile` descends).

https://github.com/rack/rack-test/pull/303 has been opened to support a `multipart` key in the env to force encoding the params as multipart. Given this, I'd hope that a future release of Capybara which depends on a `rack-test` release with this change included would be able to drop `NilUploadedFile`, set `:multipart` according to the `<form>`'s `enctype`, and avoid the historical workaround.

Barring that, I wonder if a special-case to avoid clobbering existing keys in the `params` in this code path is merited? It's trading one incorrect browser behaviour (posting the wrong param values) for another (sometimes incorrectly setting the content-type as `application/x-www-form-urlencoded` -- though only when a hidden field is also present).

The context for why I have a field that is both a file and hidden is a submitted form which did not pass validation stores the `ActiveRecord::Blob#signed_id` in a hidden field to allow the user to avoid re-uploading the file to re-submit. However, I keep the file field on the page (but non-required) so they may still choose to change its value.

Prior to https://github.com/teamcapybara/capybara/commit/29898295dc7a65c6587026fac4b3991420f0378d, it seems as though `value`-less file fields were omitted for the params, which is more representative of browser behaviour (other than I presume they would encode the params as multipart anyway).

## Meta

Capybara Version: 3.37.1 (and probably every version prior for at least a decade)
Driver Information (and browser if relevant): rack-test

## Expected Behavior

Submitting the following form fields:

```html
[...]

PR #2557, opened 2022-05-27
Proof of concept fix for #2556

bjeanes:
This fixes the issue I reported in #2556.

It would however not post as multipart without another present file _unless_ run against (https://github.com/rack/rack-test/pull/303. A workaround could be to do something like:

```rb
merge_param!(params, "_multipart_workaround", NilUploadedFile.new))
```

when not running under a new enough `rack-test`, if any future Capybara version which might include such a fix needs to maintain compatibility with older `rack-test`s.

However, I am not sure which is more problematic: submitting as single part when a `<form enctype="multipart/form-data">` has no filled file inputs or submitting params which were not present in the form in the first place.

My gut is that the latter is actually worse because a single-part form with no file fields should only be a problem if a user is asserting the lower-level data submitted -- which is patently not in the spirit of testing with Capybara. The user wouldn't care, nor should any Capybara test, I would think.

### DIFF
--- a/lib/capybara/rack_test/driver.rb
+++ b/lib/capybara/rack_test/driver.rb
@@ -50,8 +50,8 @@ class Capybara::RackTest::Driver < Capybara::Driver::Base
     browser.refresh
   end
 
-  def submit(method, path, attributes)
-    browser.submit(method, path, attributes)
+  def submit(method, path, attributes, multipart: false)
+    browser.submit(method, path, attributes, multipart: multipart)
   end
 
   def follow(method, path, **attributes)
--- a/lib/capybara/rack_test/form.rb
+++ b/lib/capybara/rack_test/form.rb
@@ -1,25 +1,6 @@
 # frozen_string_literal: true
 
 class Capybara::RackTest::Form < Capybara::RackTest::Node
-  # This only needs to inherit from Rack::Test::UploadedFile because Rack::Test checks for
-  # the class specifically when determining whether to construct the request as multipart.
-  # That check should be based solely on the form element's 'enctype' attribute value,
-  # which should probably be provided to Rack::Test in its non-GET request methods.
-  class NilUploadedFile < Rack::Test::UploadedFile
-    def initialize # rubocop:disable Lint/MissingSuper
-      @empty_file = Tempfile.new('nil_uploaded_file')
-      @empty_file.close
-    end
-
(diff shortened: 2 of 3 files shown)


lib/capybara/rack_test/driver.rb
```diff
@@ -50,8 +50,8 @@ def refresh
     browser.refresh
   end
 
-  def submit(method, path, attributes)
-    browser.submit(method, path, attributes)
+  def submit(method, path, attributes, multipart: false)
+    browser.submit(method, path, attributes, multipart: multipart)
   end
```
bjeanes, line 55: This was the easiest way to communicate the `multipart` knowledge from the form processing to the browser instance, however this interface change would be breaking if there are any callers which rely on implicit Hash arguments:

```rb
def submit(method, path, attributes, multipart: false)
  # ...
end

submit("GET", "/", "foo" => "bar")   # ! ArgumentError: wrong number of arguments (given 2, expected 3)
submit("GET", "/", {"foo" => "bar"}) # OK
submit("GET", "/", params_hash)      # OK
```

However, the headers sent for a form submission by any browser ultimately _do_ depend on the form value, so I am not sure what would be a superior way to do this.

Perhaps implementing a `submit_multipart()` instead?

lib/capybara/rack_test/form.rb
```diff
@@ -88,6 +69,8 @@ def add_input_param(field, params)
 
       Capybara::RackTest::Node.new(driver, field).value.to_s
     when 'file'
+      return if value.blank?
```
bjeanes, line 72: I would have to set up a test case with other browsers to be sure, but I think this kind of short circuit would need to be in place for other `<input>` fields with blank values.

I'm not sure if there is in practice a distinction between an empty string value and a `nil` value, but the code here already does `field['value'].to_s` above.

bjeanes (commented):
Just added some of my thoughts along the way about my proof-of-concept to seed discussion and weigh caveats.

twalpole:
This would be a breaking change and not one I'm willing to make.  If/when rack releases a new version which allows to automatically trigger multipart upload without having to pass the nil upload file then we will support that for those versions, however we will not be dropping current support for existing rack versions.  This driver attempts to mimic a real browser but is not 100% the same and isn't guaranteed to be.

PR #2563, opened 2022-06-06
Reset @scopes when session is reset

### LINKED ISSUE #2562: Using switch_to_frame(some_frame) persists across tests, even when current test has finished and the browser window is closed/reset
## Meta
Capybara Version: 3.32.2

Driver Information (and browser if relevant): selenium-webdriver 3.142.7

## Expected Behavior

Test 1 visits a url with an iframe, calls `switch_to_frame(find('#test_iframe'))`, does some assertions, then ends.
Test 2 visits the same url, does not switch to the iframe, and instead opens a new window and switches to it, does some assertions, then ends.

Both tests pass regardless of the order they're run in.

## Actual Behavior

When tests are run in order (1 then 2), the following error is raised when Test 2 calls `within_window`:
```
Capybara::ScopeError:
  Window cannot be switched inside a `within_frame` block
# /Users/me/.rbenv/versions/2.6.10/gemsets/my_gemset/gems/capybara-3.32.2/lib/capybara/session.rb:910:in `_switch_to_window'
# /Users/me/.rbenv/versions/2.6.10/gemsets/my_gemset/gems/capybara-3.32.2/lib/capybara/session.rb:529:in `within_window'
# /Users/me/.rbenv/versions/2.6.10/gemsets/my_gemset/gems/capybara-3.32.2/lib/capybara/dsl.rb:58:in `block (2 levels) in <module:DSL>'
```

When tests are run in reverse order (2 then 1), no error is raised.

## Steps to reproduce

Run the following suite with RSpec and Capybara and a browser driver capable of executing JavaScript, with the tests run in order of defined. It should all pass, but the third test fails as per above.

The tests "1" and "2" explained above are the second and third tests in this file below; I added the third test at the start just to confirm that the method by which I'm opening and switching to a new actually works (which it does).

`bundle exec rspec --order=defined this_test_suite_spec.rb`
```rb
# this_test_suite_spec.rb
describe 'visiting example.com and inserting an iframe' do

  before do
    visit 'http://example.com:80'
    execute_script("document.body.innerHTML = '<iframe id=\"test_iframe\" src=\"http://example.com:80\"></iframe>';")

    if in_iframe
      switch_to_frame(find('#test_iframe'))
    end
  end

  context 'when not in iframe' do
    let(:in_iframe) { false }

    it 'can open a new window' do
      expect do
        within_window(open_new_window) { puts "this will pass" }
      end.not_to raise_error
    end
  end

  context 'when in iframe' do
    let(:in_iframe) { true }

    it 'cannot open a new window' do
      expect do
        within_window(open_new_window) { puts "this will fail" }
      end.to raise_error(Capybara::ScopeError, 'Window cannot be switched inside a `within_frame` block')
    end
  end

  context 'when not in iframe after being in iframe' do
    let(:in_iframe) { false }

    it 'can open a new window' do
      expect do
        within_window(open_new_window) { puts "this should pass even though the previous test switches to an iframe and does not switch back" }
      end.not_to raise_error
    end
  end
end
```

henrahmagix:
Otherwise they persist between RSpec tests.

Fix #2562

### DIFF
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -129,6 +129,7 @@ module Capybara
       if @touched
         driver.reset!
         @touched = false
+        @scopes = [nil]
       end
       @server&.wait_for_pending_requests
       raise_server_error!
--- a/lib/capybara/spec/session/reset_session_spec.rb
+++ b/lib/capybara/spec/session/reset_session_spec.rb
@@ -95,6 +95,19 @@ Capybara::SpecHelper.spec '#reset_session!' do
     expect(@session.windows.size).to eq 1
   end
 
+  it 'does not block opening a new window after a frame was switched to and not switched back', requires: [:windows] do
+    @session.visit('/with_iframe?id=test_iframe&url=/')
+    @session.switch_to_frame(@session.find(:frame, 'test_iframe'))
+    within_window_test = lambda do
+      @session.within_window(@session.open_new_window) do
+        @session.visit('/')
+      end
+    end
+    expect(&within_window_test).to raise_error(Capybara::ScopeError)
+    @session.reset_session!
+    expect(&within_window_test).not_to raise_error
+  end
+
   context 'When reuse_server == false' do
     let!(:orig_reuse_server) { Capybara.reuse_server }
 
--- a/lib/capybara/spec/test_app.rb
+++ b/lib/capybara/spec/test_app.rb
@@ -177,6 +177,22 @@ class TestApp < Sinatra::Base
     HTML
   end
 
+  get '/with_iframe' do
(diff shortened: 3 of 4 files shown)


twalpole:
Thanks for this - I'll look at it in more detail later this week, but I can't currently think of any reason the scopes shouldn't be reset - if you have time to add a test to the PR that would be great.

henrahmagix:
@twalpole Done, and ran locally with chromedriver (couldn't get firefox driver to show up in my path, and need to get back to other work – can this PR be run in travis to ensure it works? ta!)

henrahmagix:
Thanks @twalpole !

PR #2566, opened 2022-06-16
Fix wrong counter behavior with changed uri

alex-konoval:
### Motivation
In some cases, we get an issue when a request is not finished for Capybara, but in reality, it's finished. It looks like we have a case when URI was changed because of server redirection or something like this and the `Counter#decrement` method is going to use a fallback `@value.length`. In the original PR the `- 1` was missed, so for value `['/example']` removing of `@value.length` will resolve to `@value.delete_at(1)` instead of `@value.delete_at(0)`.

### Change log and testing strategy
Fix fallback for Middleware::Counter#decrement which sometimes resolves to `Requests did not finish in 60 seconds` error.

### Related history link
https://github.com/teamcapybara/capybara/pull/2250#discussion_r326843512

### DIFF
--- a/lib/capybara/server/middleware.rb
+++ b/lib/capybara/server/middleware.rb
@@ -14,7 +14,7 @@ module Capybara
         end
 
         def decrement(uri)
-          @mutex.synchronize { @value.delete_at(@value.index(uri) || @value.length) }
+          @mutex.synchronize { @value.delete_at(@value.index(uri) || @value.length - 1) }
         end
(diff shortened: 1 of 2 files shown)


alex-konoval:
@twalpole @juanca please review.

juanca (approved):
Code looks good, but I'm not a capybara maintainer 😅

twalpole:
Thanks for this

PR #2572, opened 2022-07-23
Fix a duplicate options suggested in `have_xxxx` method at invalid option error message

ydah:
This PR eliminates the duplication of options proposed in the `have_xxxx` method's invalid option error message.

If the following code is executed:
```ruby
expect(page).to have_link(invalid: true)
```

The following error message appears (e.g. :href is displayed in duplicate)
```
Failure/Error: expect(page).to have_link(invalid: true)

     ArgumentError:
       Invalid option(s) :invalid, should be one of :above, :below, :left_of, :right_of, :near, :count, :minimum, :maximum, :between, :text, :id, :class, :style, :visible, :obscured, :exact, :exact_text, :normalize_ws, :match, :wait, :filter_set, :focused, :href, :href, :alt, :title, :download
```

### DIFF
--- a/lib/capybara/queries/selector_query.rb
+++ b/lib/capybara/queries/selector_query.rb
@@ -355,7 +355,7 @@ module Capybara
         return if unhandled_options.empty?
 
         invalid_names = unhandled_options.map(&:inspect).join(', ')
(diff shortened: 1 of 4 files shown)


twalpole:
Thank you

PR #2584, opened 2022-09-22
Fix spelling

ydah:
Several typos were corrected in this PR.

### DIFF
--- a/History.md
+++ b/History.md
@@ -830,7 +830,7 @@ Release date: 2018-03-23
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


twalpole:
Thanks for this but I shouldn't accept most of these changes, specifically the ones that are in test names.  Other drivers use the test suite in lib/capybara/spec and may use the names for filtering specific tests so changing large numbers of them outside of a major release isn't great.  I'll create a new commit with the other ones, although the change to code logic concerns me that it wasn't caught by a test

PR #2589, opened 2022-10-10
consider Selenium::WebDriver::Error::DetachedShadowRootError as invalid_element_error

perryn:


perryn:
I am introducing a widget built with a shadow DOM into our app.

Some of our tests were flaky with occasional errors like this

```
     Failure/Error: tab_nav.find("a.active").text01:23
     01:23
     Selenium::WebDriver::Error::DetachedShadowRootError:01:23
       detached shadow root: detached shadow root01:23
         (Session info: chrome=105.0.5195.125)01:23

     <SNIP>

     # --- Caused by: ---01:23
     # Selenium::WebDriver::Error::StaleElementReferenceError:01:23
     #   stale element reference: element is not attached to the page document01:23
     #     (Session info: chrome=105.0.5195.125)01:23
     #   #0 0x5595906ce693 <unknown>
```

Once I
        - 1) made this patch to consider DetachedShadowRootError as an 'invalid_element_error'
        - 2) wrapped the selector code in a `.synchronize` block
The test became much more stable

twalpole:
Thanks for this but it would have broken backwards compatibility with Selenium versions which don't define that error.  I added it via https://github.com/teamcapybara/capybara/pull/2592

Issue #2576, opened 2022-08-29
html5_drag does not fire dragenter


sfnelson:
## Meta
Capybara Version:
3.37.1

Driver Information (and browser if relevant):
Chrome

## Expected Behavior
When simulating a drag and drop, Capybara should fire a dragenter event before firing dragover
See spec: https://www.w3.org/html/wg/spec/dnd.html

## Actual Behavior
The `dragenter` event does not fire.

## Steps to reproduce
Add a listener for dragenter events and verify that when using the `drag_to` the listener does not fire.

This is because `lib/capybara/selenium/extensions/html5_drag.rb` does not implement firing dragenter.

I would expect this event to be fire from `dragEnterTarget` before firing `dragover`.

PR #2601, opened 2022-10-19
Add `dragenter` event and test to `drag_to`

matt17r:
Fire a `dragenter` event before the two `dragover` events.

The [spec](https://www.w3.org/html/wg/spec/dnd.html#introduction-6) requires a `dragenter` event to be fired first:

> To accept a drop, the drop target has to listen to at least three events. First, the [dragenter](https://www.w3.org/html/wg/spec/dnd.html#event-dragenter) event, which is used to determine whether or not the drop target is to accept the drop.

Addresses #2576 

This is my first code contribution and I'm much more comfortable in Ruby than JavaScript so please let me know if I've done this badly or misunderstood the JavaScript side of things!

### DIFF
--- a/lib/capybara/selenium/extensions/html5_drag.rb
+++ b/lib/capybara/selenium/extensions/html5_drag.rb
@@ -165,6 +165,9 @@ class Capybara::Selenium::Node
           }
           opts[key + 'Key'] = true;
         }
+        
+        var dragEnterEvent = new DragEvent('dragenter', opts);
+        target.dispatchEvent(dragEnterEvent);
 
         // fire 2 dragover events to simulate dragging with a direction
         var entryPoint = pointOnRect(sourceCenter, targetRect)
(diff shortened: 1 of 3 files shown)


boris-petrov:
Perhaps also add a test that verifies that the `dragenter` is dispatched *before* the `dragover`?

matt17r:
How about checking for both at the same time, @boris-petrov? Looks a bit ugly to me but since I need to check their order anyway, this finds the element in questions then checks the dragover occurs next using `preceding-sibling`

boris-petrov:
Isn't it possible to do something like `expect(page).to have_text(/DragEnter.*DragOver/)`?

matt17r:
That's much simpler! Events are listed in reverse chronological order...

image

 It's not obvious reading the test and I couldn't think of a way to make it explicit in code so I added a comment. There's a similar comment on what's currently line 562

matt17r:
Whoops, now the test isn't passing. Bear with me 🙃

matt17r:
Ugh, I'm trying to rebase against master, squash it all into a single commit and then force push to my branch but I clearly don't know what I'm doing

boris-petrov (approved):

PR #2611, opened 2022-11-04
Fix `Driver#submit` to forward the required args

dsantosmerino:
## Issue
It looks like the recent release `3.38.0` contains a [small bug](https://github.com/teamcapybara/capybara/commit/a3ff5514f40f3af0ec6655354d3596f6d9ea8964#diff-7ba980837bac726069131342051c1571ce448f8c54887109543c0fbc85e30bc8R57) when forwarding params from the `Capybara::RackTest::Driver` to the `Capybara::RackTest::Browser`.

From our spec suite:
```bash
     Failure/Error: Capybara::RackTest::Form.new(page.driver, form.native).submit({})
     
     ArgumentError:
```

After checking it a bit, it looks related to this change a3ff5514f40f3af0ec6655354d3596f6d9ea8964, when the `Capybara::RackTest::Browser#submit` signature was changed, but not the `Capybara::RackTest::Driver` one.

## Approach
Replacing the existing signature to just forward the received params to the final class. I guess we could use this same approach to same methods present in this class, but I didn't want to mix things.

_Note: I know that it doesn't contain any spec, but I am not familiar with the code base. Let me know if you have a clear idea of what I should cover here and I'll do my best to add them ASAP._

### DIFF
--- a/lib/capybara/rack_test/driver.rb
+++ b/lib/capybara/rack_test/driver.rb
@@ -50,8 +50,8 @@ class Capybara::RackTest::Driver < Capybara::Driver::Base
     browser.refresh
   end
 
-  def submit(method, path, attributes)
-    browser.submit(method, path, attributes)
+  def submit(...)
+    browser.submit(...)
   end
 
   def follow(method, path, **attributes)


twalpole:
What version of rack are you using?

dsantosmerino:
>  What version of rack are you using?

We're using Rack v2.2.4, @twalpole.

twalpole:
@dsantosmerino Capybara tests pass without this change, what code of yours is triggering an error here?

twalpole:
@dsantosmerino I don't think your issue is where you think it is since the Capybara tests all pass and do exercise this method.  I think the issue is that you're passing an empt hash to `submit` rather than the button that was clicked to trigger the submit

john-h-k:
@twalpole we are having the same issue due to someone in the past having created the same hack. However, it is common enough to have a form without a submit button, so it seems very odd to not support that

twalpole:
@john-h-k If you're using the rack-test driver then you're running in a non-JS supporting environment -- with no submit button how is the form being submitted?

ndbroadbent:
I also ran into this error after updating to `3.38.0`. I was using a workaround to test a search input that doesn't have a submit button:

Screen Shot 2022-11-18 at 12 47 39 PM


```
      scenario 'query has matching folder on root folder' do
        visit templates_path
        fill_in 'query', with: 'root folder two'
        Capybara::RackTest::Form.new(page.driver, find('form.search-form').native).submit
```

It would be great if it was possible to add support for this case and a public API for submitting a form with the rack-test driver.

I had probably found this workaround on StackOverflow or a Gist:

* https://stackoverflow.com/questions/8757477/capybara-submit-a-form-without-button
* https://gist.github.com/beccasaurus/989533

PR #2612, opened 2022-11-05
Use GitHub Actions for CI

mvz:
This is a work-in-progress attempt to migrate from Travis to GitHub Actions.


mvz:
Actual results of the actions can be checked here for now: https://github.com/mvz/capybara/actions

mvz:
@twalpole there's one failure left which I think means one more exception has to be added to the spec runners. I'm putting this up for review so we can discuss.

mvz:
All jobs are green now: https://github.com/mvz/capybara/actions/runs/3829507147/jobs/6516284770

twalpole:
@mvz I don't like the moving of the skip/pending settings of tests from the selenium_spec_chrome.rb, etc  files into node_spec.rb.  The files in lib/capybara/spec are used by other drivers to run the tests against them, and the conditional skipping shouldn't be in them.

mvz:
@twalpole alright, I have some questions on how that works but that's something for another issue. For now, I'll remove that change.

twalpole:
Thank you very much

PR #2617, opened 2022-11-18
Forward given filter block to matcher

### LINKED ISSUE #2616: Different behavior between {..} and do..end block with filter block of rspec matchers
## Meta
Capybara Version: 3.38.0

Driver Information: selenium-webdriver 4.6.1

## Expected Behavior

These two expectations should behave identically as only the block notation differs:

```ruby
# will fail
expect(page).to have_selector("input") { false }

# will pass
expect(page).to have_selector("input") do
  false
end
```

Currently they behave differently: first one is failing, second one is passing. I think second one should fail too.

Why does it pass in the second case: as `do..end` block having lower precedence than `{..}` block, the block in the first expectation is attached to the `have_selector` matcher while the block in the second expectation is attached to the `#to` method.

Adding parens can make the second expectation fail as expected:
```ruby
# will fail
expect(page).to(have_selector("input") do
  false
end)
```

## Actual Behavior
```ruby
expect(page).to have_selector("input") do |input|
  input.value == "42"
end
```

This expectation will pass regardless of the actual value in the input element, as the filter block is never evaluated.

## Steps to reproduce

Something like this should fail:
```ruby
expect(page).to have_selector("input") do
  false
end
```

I'll submit a PR for this.

cbliard:
Fixes #2616

### DIFF
--- a/lib/capybara/rspec/matchers/base.rb
+++ b/lib/capybara/rspec/matchers/base.rb
@@ -47,15 +47,19 @@ module Capybara
       end
 
       class WrappedElementMatcher < Base
-        def matches?(actual)
-          element_matches?(wrap(actual))
+        def matches?(actual, &filter_block)
+          with_filter(filter_block) do
+            element_matches?(wrap(actual))
+          end
         rescue Capybara::ExpectationNotMet => e
           @failure_message = e.message
           false
         end
 
-        def does_not_match?(actual)
-          element_does_not_match?(wrap(actual))
+        def does_not_match?(actual, &filter_block)
+          with_filter(filter_block) do
+            element_does_not_match?(wrap(actual))
+          end
         rescue Capybara::ExpectationNotMet => e
           @failure_message_when_negated = e.message
           false
@@ -70,6 +74,14 @@ module Capybara
           else
             Capybara.string(actual.to_s)
           end
+        end
+
+        def with_filter(filter_block)
+          previous_filter_block = @filter_block
+          @filter_block = filter_block if filter_block
+          yield
+        ensure
+          @filter_block = previous_filter_block
         end
       end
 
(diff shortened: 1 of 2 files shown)


twalpole:
Sorry for the delay, this will be looked at/considered this coming weekend

twalpole:
I don't think the tests added are actually validating the filter blocks are run since the `have_selector` calls would pass with or without the filter blocks.

cbliard:
You're right. When writing the tests, I knew that some of them were not actually accurate as they would pass with or without the block.

This one for instance will happily pass with or without the block:
```ruby
expect(page).to have_selector(:css, 'input#test_field') { |input| input.value == 'monkey' }
```

Contrary to this one which would not pass without the block:
```ruby
expect(page).to have_no_selector(:css, 'input#test_field') { |input| input.value == 'not a monkey' }
```

How should the tests be formulated then? I saw in `lib/capybara/spec/session/element/match_css_spec.rb` that some tests are written like `expect { expect(page).to match_something }.to raise_exception(RSpec::Expectations::ExpectationNotMetError)`. Following it, the test can be written like this:

```ruby
expect { expect(page).to have_selector(:css, 'input#test_field') { |input| input.value == 'not a monkey' } }
  .to raise_exception(RSpec::Expectations::ExpectationNotMetError)
```

Would it be better? Any other suggestions?

twalpole:
Either that or changing the tests to consider number of matched elements like

    expect(page).to have_selector(:css, 'input', count: 1) { |input| input.value == 'monkey' }

which would be easier to read/understand

twalpole:
Looking at this further, this breaks the default binding of the `do` block to `to` which is defined to allow a proc for failure messages.  Not really sure what the best solution here actually is.  This PR is overriding Rubys default binding precedence and could break existing code...

twalpole:
Merged via cacd032489b443fe5d9580752776d7e34fea4203

cbliard:
Thanks @twalpole for merging.

I'm still feeling bad about the tests. Do you mind if I open another PR to provide better tests with `count: 1` as you suggested?

twalpole:
@cbliard I had already fixed the tests

PR #2618, opened 2022-11-20
Skip shadow root related spec for #evaluate_script on older Selenium

mvz:
This makes specs for Chrome pass with the base dependency versions:

```
BUNDLE_GEMFILE='gemfiles/Gemfile.base-versions' bundle exec rake spec_chrome
```


twalpole:
Sorry for the delay, this will be looked at/considered this coming weekend

Issue #2558, opened 2022-06-01
evaluate_async_script not available in Capybara::DSL


henrahmagix:
## Meta
Capybara Version: 3.32.2

Driver Information (and browser if relevant): selenium-webdriver (3.142.7)

## Expected Behavior

`evaluate_async_script` will call `page.evaluate_async_script`

## Actual Behavior

```
NameError: undefined local variable or method `evaluate_async_script' for #<RSpec::ExampleGroups::MySuite::WhenInMyTest:0x0000000109475338>
Did you mean?  evaluate_script
from /Users/me/.rbenv/versions/2.6.10/gemsets/my_project/gems/rspec-expectations-3.10.1/lib/rspec/matchers.rb:965:in `method_missing'
```

## Steps to reproduce

Setup a Capybara test and call all the *_script methods on `page`:
```rb
page.execute_script('window._page_execute_script = true')
puts page.evaluate_script('"window._page_execute_script equals: " + window._page_execute_script')
puts page.evaluate_async_script('var callback = arguments[arguments.length-1]; setTimeout(() => callback("async done"), 500);')
```
This will succeed. Then repeat but remove the reference to `page`:
```rb
execute_script('window._page_execute_script = true')
puts evaluate_script('"window._page_execute_script equals: " + window._page_execute_script')
puts evaluate_async_script('var callback = arguments[arguments.length-1]; setTimeout(() => callback("async done"), 500);')
```
The first two will succeed, the third will error as per "Actual Behaviour" described above.

henrahmagix:
Thanks @twalpole :)

PR #2627, opened 2022-12-27
add: `Session#execute_async_script`

seanpdoyle:
Similar to [teamcapybara/capybara#2558][] and
[teamcapybara/capybara#2559][], extends the `Session` interface to implement `execute_async_script`.

[teamcapybara/capybara#2558]: https://github.com/teamcapybara/capybara/issues/2558
[teamcapybara/capybara#2559]: https://github.com/teamcapybara/capybara/pull/2559

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+### Added
+
+* Support for `Sesison#execute_async_script`
+
 # Version 3.39.0
 Release date: unreleased
 
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -59,7 +59,7 @@ module Capybara
     ].freeze
     SESSION_METHODS = %i[
       body html source current_url current_host current_path
-      execute_script evaluate_script evaluate_async_script visit refresh go_back go_forward send_keys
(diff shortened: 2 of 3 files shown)


seanpdoyle:
@twalpole I'm not sure why, but this change didn't execute the test suite in CI.

seanpdoyle:
The motivation behind this change involved calls to `execute_async_script` made from the [axe-core-api](https://github.com/dequelabs/axe-core-gems/blob/develop/packages/axe-core-api/lib/webdriver_script_adapter/execute_async_script_adapter.rb#L97) gem, but that gem might be invoking `execute_async_script`, then acting upon the return value.

Despite the fact that [selenium-webdriver](https://github.com/SeleniumHQ/selenium/blob/trunk/rb/lib/selenium/webdriver/common/driver.rb#L224-L241) defines an `execute_async_script` to return the script's value, Capybara's `execute`-prefixed methods are deliberate about **not** returning a value.

Would it be best to keep `execute_async_script` out of the API and continue to require that callers use `evaluate_async_script` instead?

twalpole:
Yes - Capybara specifically has execute vs evaluate because of the overhead of returning values.  There is no point in an `execute_async_script` in Capybara since it can just be done via `execute_script`.  If you need the return value then use `evaluate_async_script`

PR #2630, opened 2023-01-06
Added dependabot for GitHub Actions and bundler

ydah:
How about using dependabot in this way?
When this works, a PullRequest is created as follows:
- ruby/bigdecimal#242

It was created with reference to the following:
- ruby/csv#265

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


mvz:
How about also enabling it for the gem dependencies?

ydah:
Indeed it is, I believe Gemfile also created a PR for me if I had specified a version limit, but it doesn't seem to be necessary this repository.

simi:
@mvz even for gemspec based constraints? :thinking:

mvz:
@simi yes, I use it for nearly all my gems.

twalpole:
I'm not interested in this right now -- thanks anyway

PR #2632, opened 2023-01-07
ExpectedError is not defined in Selenium 4+

twalpole:


foobear:
Thanks for the fix! Will there be an official gem release?

PR #2633, opened 2023-01-13
Scope net_http_connect_on_start in WebMock gotcha

spikeheap:
While debugging the [net_http_connect_on_start gotcha](https://github.com/teamcapybara/capybara#gotchas) I found it's sometimes preferable to restrict the scope of `net_http_connect_on_start` to prevent network connections to external services while allowing (and fixing) connections to the local test server.


twalpole:
Capybara.server_host always has a value, so I'm not sure what the point of using it to set `net_http_connect_on_start` is?

spikeheap:
Hi @twalpole, `net_http_connect_on_start` accepts a host, so we're not using it as a truthy value here. We're saying "Enable `net_http_connect_on_start` only for `server_host`, and retain the default behaviour for other hosts".

Issue #2634, opened 2023-01-15
`all_text` performance characteristics and proposal


baweaver:
## Overview

While benchmarking against Capybara specs I had noticed that flamegraphs were showing a significant amount of time on `gsub`, corresponding to the following code:

```ruby
def all_text
  text = driver.evaluate_script('arguments[0].textContent', self) || ''
  text.gsub(/[\u200b\u200e\u200f]/, '')
      .gsub(/[\ \n\f\t\v\u2028\u2029]+/, ' ')
      .gsub(/\A[[:space:]&&[^\u00a0]]+/, '')
      .gsub(/[[:space:]&&[^\u00a0]]+\z/, '')
      .tr("\u00a0", ' ')
end
```

...which can be found in either `Capybara::Selenium::Node` or `Capybara::RackTest::Node`.

## Areas of Interest

There were a few potential areas of interest with this code:

1. **Inline Regex construction** - Every call of `all_text` creates 4 regex objects which could be solved with hoisting constants
2. **`gsub` slowness** - `gsub` can be slower than methods like `squeeze` or `tr`, and for lead/trailing removals `sub` could be used instead
3. **Regex slowness** - In some of these cases `tr` and a literal string could be used instead, offering gains on regex usage
4. **Immutable Element** - Some parts of `Capybara::Node::Element` are immutable, or cache invalidation could be used, to not recompute text on every search
5. **Hot Path** - `all_text` is called _a lot_ which means that any improvements to this method will have dramatic performance impacts.

Given that and the flamegraph observations I decided to see how much of a difference this would make, and I believe it may be of interest to the Capybara team.

## Benchmarks

The following benchmarks include both the current variant of the `all_text` method as well as a variant which avoids usage of `gsub` and hoists constants:

```ruby
require "benchmark/ips"

NO_CACHING = true

class Node
  NON_BREAKING_SPACE = "\u00a0"
  LINE_SEPERATOR = "\u2028"
  PARAGRAPH_SEPERATOR = "\u2029"

  BREAKING_SPACES = "[[:space:]&&[^#{NON_BREAKING_SPACE}]]"

  SQUEEZED_SPACES = "\ \n\f\t\v#{LINE_SEPERATOR}#{PARAGRAPH_SEPERATOR}"
  LEADING_SPACES = /\A#{BREAKING_SPACES}+/
  TRAILING_SPACES = /#{BREAKING_SPACES}+\z/

  ZERO_WIDTH_SPACE = "\u200b"
[...]

PR #2635, opened 2023-01-15
Optimize `all_text` and `visible_text` for Nodes

baweaver:
Optimizes the `all_text` and `visible_text` methods for Selenium and RackTest Nodes by hoisting constants, consolidating substitutions, and switching out instances of `gsub` with more efficient methods or by removing the need for regex.

In response to issue #2634

### DIFF
--- a/lib/capybara/driver/node.rb
+++ b/lib/capybara/driver/node.rb
@@ -5,6 +5,24 @@ module Capybara
     class Node
       attr_reader :driver, :native, :initial_cache
 
+      NON_BREAKING_SPACE = "\u00a0"
+      LINE_SEPERATOR = "\u2028"
+      PARAGRAPH_SEPERATOR = "\u2029"
+
+      BREAKING_SPACES = "[[:space:]&&[^#{NON_BREAKING_SPACE}]]"
+
+      SQUEEZED_SPACES = "\ \n\f\t\v#{LINE_SEPERATOR}#{PARAGRAPH_SEPERATOR}"
+      LEADING_SPACES = /\A#{BREAKING_SPACES}+/
+      TRAILING_SPACES = /#{BREAKING_SPACES}+\z/
+
+      ZERO_WIDTH_SPACE = "\u200b"
+      LEFT_TO_RIGHT_MARK = "\u200e"
+      RIGHT_TO_LEFT_MARK = "\u200f"
+
+      REMOVED_CHARACTERS = [ZERO_WIDTH_SPACE, LEFT_TO_RIGHT_MARK, RIGHT_TO_LEFT_MARK].join
+
+      EMPTY_LINES = /[\ \n]*\n[\ \n]*/
+
       def initialize(driver, native, initial_cache = {})
         @driver = driver
         @native = native
@@ -137,6 +155,27 @@ module Capybara
 
       def ==(other)
         eql?(other) || (other.respond_to?(:native) && native == other.native)
+      end
+
+      protected
+
+      def normalize_spacing(text)
+        text
+          .delete(REMOVED_CHARACTERS)
+          .tr(SQUEEZED_SPACES, ' ')
+          .squeeze(' ')
+          .sub(LEADING_SPACES, '')
+          .sub(TRAILING_SPACES, '')
+          .tr(NON_BREAKING_SPACE, ' ')
+      end
+
+      def normalize_visible_spacing(text)
+        text
+          .squeeze(' ')
+          .gsub(EMPTY_LINES, "\n")
+          .sub(LEADING_SPACES, '')
+          .sub(TRAILING_SPACES, '')
+          .tr(NON_BREAKING_SPACE, ' ')
       end
     end
   end
--- a/lib/capybara/driver/node.rb
+++ b/lib/capybara/driver/node.rb
@@ -5,24 +5,6 @@ module Capybara
     class Node
       attr_reader :driver, :native, :initial_cache
 
-      NON_BREAKING_SPACE = "\u00a0"
-      LINE_SEPERATOR = "\u2028"
-      PARAGRAPH_SEPERATOR = "\u2029"
-
-      BREAKING_SPACES = "[[:space:]&&[^#{NON_BREAKING_SPACE}]]"
-
-      SQUEEZED_SPACES = "\ \n\f\t\v#{LINE_SEPERATOR}#{PARAGRAPH_SEPERATOR}"
-      LEADING_SPACES = /\A#{BREAKING_SPACES}+/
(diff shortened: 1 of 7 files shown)


lib/capybara/driver/node.rb
```diff
@@ -5,6 +5,24 @@ module Driver
     class Node
       attr_reader :driver, :native, :initial_cache
 
+      NON_BREAKING_SPACE = "\u00a0"
+      LINE_SEPERATOR = "\u2028"
+      PARAGRAPH_SEPERATOR = "\u2029"
```
baweaver, line 10: Made some of the unicode characters into named constants to know what each one did, as looking them up takes you away from the code.

lib/capybara/driver/node.rb
```diff
@@ -138,6 +156,27 @@ def inspect
       def ==(other)
         eql?(other) || (other.respond_to?(:native) && native == other.native)
       end
+
+      protected
+
+      def normalize_spacing(text)
+        text
+          .delete(REMOVED_CHARACTERS)
```
baweaver, line 164: These were straight deletions on the previous `gsub`

lib/capybara/driver/node.rb
```diff
@@ -138,6 +156,27 @@ def inspect
...
+          .tr(SQUEEZED_SPACES, ' ')
+          .squeeze(' ')
```
baweaver, line 166: These were `v+` regexes before. Not _as_ sure on this one.

lib/capybara/driver/node.rb
```diff
@@ -138,6 +156,27 @@ def inspect
...
+          .sub(LEADING_SPACES, '')
+          .sub(TRAILING_SPACES, '')
```
baweaver, line 168: These ones are `\A` and `\z` beginning and end bound, so there can only really be one instance of them in a string. Probably a _slightly_ faster way to do this

lib/capybara/driver/node.rb
```diff
@@ -138,6 +156,27 @@ def inspect
...
+          .tr(NON_BREAKING_SPACE, ' ')
```
baweaver, line 169: Then preserve NBSP

lib/capybara/node/element.rb
```diff
@@ -56,9 +56,12 @@ def native
...
       def text(type = nil, normalize_ws: false)
-        type ||= :all unless session_options.ignore_hidden_elements || session_options.visible_text_only
-        txt = synchronize { type == :all ? base.all_text : base.visible_text }
-        normalize_ws ? txt.gsub(/[[:space:]]+/, ' ').strip : txt
+        unless session_options.ignore_hidden_elements || session_options.visible_text_only
+          type ||= :all
+        end
+
+        extracted_text = synchronize { type == :all ? base.all_text : base.visible_text }
+        normalize_ws ? extracted_text.gsub(/[[:space:]]+/, ' ').strip : extracted_text
```
baweaver, line 64: Should probably undo this section, as that was left over from experimenting with caching.

lib/capybara/rack_test/node.rb
```diff
@@ -153,9 +144,11 @@ def displayed_text(check_ancestor: true)
...
       ''
     elsif native.text?
-      native.text
-            .gsub(/[\u200b\u200e\u200f]/, '')
-            .gsub(/[\ \n\f\t\v\u2028\u2029]+/, ' ')
+      native
+        .text
+        .delete(REMOVED_CHARACTERS)
+        .tr(SQUEEZED_SPACES, ' ')
+        .squeeze(' ')
```
baweaver, line 151: Kinda wonder if the `visible_text` cleaning should be done here instead?

lib/capybara/driver/node.rb
```diff
@@ -138,6 +156,27 @@ def inspect
       def ==(other)
         eql?(other) || (other.respond_to?(:native) && native == other.native)
       end
+
+      protected
+
+      def normalize_spacing(text)
```
technicalpickles, line 162: Could we include some comments here or around the constants about why we are using the constants? Thinking about future readers of this, and the temptation to reverse it because they are only used in one place 😅

lib/capybara/driver/node.rb
twalpole: Why are you putting these in this class? I get that it's a base but it's also used by a number of 3rd party drivers.  These should be defined in any driver specific nodes they are needed/wanted in (or into a module which can be included into those driver specific classes)

lib/capybara/driver/node.rb
twalpole: None of these new methods should be in this class

lib/capybara/node/element.rb
twalpole: Yes - undo this

twalpole (changes requested):

lib/capybara/driver/node.rb
```diff
@@ -138,6 +156,27 @@ def inspect
...
+          .tr(NON_BREAKING_SPACE, ' ')
+      end
+
+      def normalize_visible_spacing(text)
+        text
+          .squeeze(' ')
+          .gsub(EMPTY_LINES, "\n")
+          .sub(LEADING_SPACES, '')
+          .sub(TRAILING_SPACES, '')
+          .tr(NON_BREAKING_SPACE, ' ')
```
technicalpickles, line 178: I wonder if using `!` versions on a `text.dup` would be any faster 🤔  It would be only 1 object created vs 5.

lib/capybara/driver/node.rb
baweaver: Mostly base class, yes, considering that Selenium and RackTest both used it. Module makes sense, some sort of whitespace cleaner, can get that done tonight when I get back.

lib/capybara/driver/node.rb
baweaver: Will move all of these to a new file and comment them all fairly heavily on what/why they're around.

lib/capybara/driver/node.rb
baweaver: Will move to the implied new module mentioned above.

lib/capybara/node/element.rb
baweaver: Will do.

lib/capybara/driver/node.rb
baweaver: May be worth trying out.

baweaver:
Changes made according to feedback.

baweaver:
@twalpole - Heya, checking back in on this one, any other improvements we should make?

PR #2643, opened 2023-02-08
Add `hover` to `session`

boris-petrov:
Similar to https://github.com/teamcapybara/capybara/commit/85205619833e7bee1dc8c131bf149494e56a22f9

### DIFF
--- a/lib/capybara/session.rb
+++ b/lib/capybara/session.rb
@@ -41,7 +41,7 @@ module Capybara
     NODE_METHODS = %i[
       all first attach_file text check choose scroll_to scroll_by
       click double_click right_click
-      click_link_or_button click_button click_link
+      click_link_or_button click_button click_link hover
       fill_in find find_all find_button find_by_id find_field find_link
       has_content? has_text? has_css? has_no_content? has_no_text?
       has_no_css? has_no_xpath? has_xpath? select uncheck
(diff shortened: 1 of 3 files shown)


twalpole:
I don't get the use of this --- clicking somewhere randomly I understand -- not really sure what anyone is going to test with random hovering -- would seem to make more sense to be specific in a test as to what you are hovering over

boris-petrov:
Sorry, I'm not sure what you mean by "random hovering". I have the following use-case:

```ruby
within 'div', exact_text: 'foobar' do
  hover
  find('.something-that-appears-on-hover').click
end
```

This doesn't work right now in Capybara. One has to use `page.current_scope.hover` instead. I tried implementing that with this PR. Did I instead do something else? :smile:

twalpole:
I'm saying that `hover` alone doesn't really tell someone reading the test what it's actually doing.  As a normal user the mouse would generally be stationary ('hovering') - so seeing the line 

    ...
    hover
    ....

doesn't really explain what is being done, or how it's different from just leaving the mouse where it is.  I think it would be much better to write your code as

    within 'div', exact_text: 'foobar' do |my_element|
      my_element.hover
      find('.something-that-appears-on-hover').click
    end

Overall I've come around to think we should probably be looking to remove methods from `Session` rather than adding more

PR #2653, opened 2023-03-13
Allow fill_in to work with reactjs

iainbeeston:
React overrides the native setter of html input elements, so it doesn't fire change events when the value is changed. Because of that, for those input types where capybara uses javascript to change the value, change events are not fired as you'd expect when using react.

I've updated the code to use the un-overridden value setter instead as recommended here:

https://stackoverflow.com/a/46012210

I believe this shouldn't adversely affect apps using other javascript frameworks or no frameworks at all


twalpole:
I think this is a bit too risky to make the default immediately - may provide it as an optional behavior

iainbeeston:
My biggest issue is how to test this, do you have any advice for that?

PR #2665, opened 2023-05-08
fix argument error from change in Selenium 4.9.1

titusfortner:
This fixes the error, I'm not sure it is the best solution.

Not sure how it would work into the Capybara approach, but you can also use: `logger.ignore(:deprecations)`.

### DIFF
--- a/lib/capybara/selenium/logger_suppressor.rb
+++ b/lib/capybara/selenium/logger_suppressor.rb
@@ -3,7 +3,7 @@
 module Capybara
   module Selenium
     module DeprecationSuppressor
-      def initialize(*)
+      def initialize(*, **)
         @suppress_for_capybara = false
         super
       end


aramprice:
See also 
- https://github.com/teamcapybara/capybara/issues/2666
- https://github.com/teamcapybara/capybara/pull/2667

twalpole:
@titusfortner thanks --- went with https://github.com/teamcapybara/capybara/pull/2667 instead --  breaking changes in a patch release aren't great though ... :)

titusfortner:
Yup, that one is better, just wanted to highlight the issue.

And Selenium will never be able to do semver. We're prioritizing keeping the features consistent between minor versions across the bindings. It would be easier to release minor versions instead of patch versions if we had active/dedicated people on each of the languages like we used to.

titusfortner:
Also also, I didn't expect that to be a breaking change because we were already using a double splat. 😄

PR #2667, opened 2023-05-08
selenium log suppression: forward both positional and keyword args

kyrofa:
Ruby 3 separated positional and keyword arguments, which changes the way we need to forward them. Just using `*` isn't enough, as exposed by #2666. We either need to use `...` or handle both positional and keyword. We opt for the latter approach here since it's compatible with the most Ruby versions.

### DIFF
--- a/lib/capybara/selenium/logger_suppressor.rb
+++ b/lib/capybara/selenium/logger_suppressor.rb
@@ -3,7 +3,7 @@
 module Capybara
   module Selenium
     module DeprecationSuppressor
-      def initialize(*)
+      def initialize(*, **)
         @suppress_for_capybara = false
         super
       end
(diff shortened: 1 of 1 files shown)


lib/capybara/selenium/logger_suppressor.rb
```diff
@@ -3,7 +3,7 @@
 module Capybara
   module Selenium
     module DeprecationSuppressor
-      def initialize(*, **)
+      def initialize(...)
```
kyrofa, line 6: I was about to point out that this is incompatible with Ruby 2.5, but then I took a closer look at the gemspec and saw the `required_ruby_version` :stuck_out_tongue: . I'm sorry, I should have done it this way from the beginning.

PR #2671, opened 2023-05-18
Fix #2670 Wrong number of arguments exception

### LINKED ISSUE #2670: Capybara::RackTest::Form.submit throws exception using Capybara::RackTest::Driver
## Meta
Capybara Version: 3.39.1
Driver Information (and browser if relevant): rack-test 3.39.1

## Expected Behavior
Calling Capybara::RackTest::Form.new(page.driver, form.native).submit({}) when driver is Capybara::RackTest::Driver should not throw an exception.

## Actual Behavior
Exception is thrown because of change in this commit https://github.com/teamcapybara/capybara/commit/a3ff5514f40f3af0ec6655354d3596f6d9ea8964 where it is assumed that the driver will be a browser.

```
     ArgumentError:
     # /bundle/ruby/3.1.0/gems/capybara-3.39.1/lib/capybara/rack_test/driver.rb:53:in `submit'
     # /bundle/ruby/3.1.0/gems/capybara-3.39.1/lib/capybara/rack_test/form.rb:57:in `submit'
```

## Steps to reproduce
Capybara::RackTest::Form.new(page.driver, form.native).submit({}) with Capybara::RackTest::Driver

cschilbe:
## Changes
Resolve https://github.com/teamcapybara/capybara/issues/2670 by
passing content-type argument to submit in attributes instead of as another param to preserve interface in use by other drivers.

### DIFF
--- a/lib/capybara/rack_test/browser.rb
+++ b/lib/capybara/rack_test/browser.rb
@@ -31,7 +31,8 @@ class Capybara::RackTest::Browser
     request(last_request.fullpath, last_request.env)
   end
 
-  def submit(method, path, attributes, content_type: nil)
+  def submit(method, path, attributes)
+    content_type = attributes.delete(:content_type)
     path = request_path if path.nil? || path.empty?
     uri = build_uri(path)
     uri.query = '' if method.to_s.casecmp('get').zero?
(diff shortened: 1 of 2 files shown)


cschilbe:
Alternatively, we could update [the Capybara::RackTest::Driver](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/rack_test/driver.rb#L53) to take a fourth argument and pass that along - I just wondered if there may be other drivers that are actually broken with this change.

twalpole:
Thanks, but this has been discussed multiple times in the issues - I have no desire to support functionality which encourages bad testing practices

PR #2677, opened 2023-06-13
Remove deprecation warning about Rack::Handler

etiennebarrie:
Testing a gem against Rails main with Rack 3 [showed a deprecation warning](https://github.com/Shopify/maintenance_tasks/actions/runs/5256129811/jobs/9496981233#step:8:13) when Capybara refers to `Rack::Handler`. [Rack defines an autoload for Handler to `"rack/handler"`](https://github.com/rack/rack/blob/v3.0.7/lib/rack.rb#L34) and that file causes a deprecation warning. This uses a similar fix as 2db340d184782912d0cd2e0fa3ab68b731c439dd, storing the module in a variable on which we get the `:Handler` constant.

If `require "rackup"` works without a `LoadError`, it means that `Rackup` defines `Handler` (I'm not sure why it's normal, but `require "rackup"` in rackup 1.0 always raises because it tries to [`require_relative` a file that doesn't exist](https://github.com/rack/rackup/blob/v1.0.0/lib/rackup.rb#L6)).

BTW should capybara depend on rackup since it uses it?

### DIFF
--- a/lib/capybara/registrations/servers.rb
+++ b/lib/capybara/registrations/servers.rb
@@ -19,16 +19,18 @@ Capybara.register_server :webrick do |app, port, host, **options|
 end
 
 Capybara.register_server :puma do |app, port, host, **options| # rubocop:disable Metrics/BlockLength
-  begin
+  base_module = begin
     require 'rackup'
+    Rackup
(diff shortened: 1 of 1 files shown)


etiennebarrie:
Fixed in #2706

PR #2680, opened 2023-06-14
Enable compatibility with ruby 3.2:

barrerajl:
Use metaprogaming in session with key params detection Enable ruby 3, 3.1 and 3.2 for all tests


twalpole:
Capybara is already compatible with Ruby 3.2 - this patch appears to be adding back in ruby 2.7 keywords support via a hack which is unwanted.  As commented in the issue ticket configure your app to properly use the Capybara matchers

PR #2682, opened 2023-07-10
Add ability to add more capybara server middleware

tubaxenor:
Expose `extra_middleware` as a configuration option instead of only apply `Capybara::Server::AnimationDisabler` when `disable_animation` was set. 

Capybara::Server has the ability to inject middleware after e8b49b6eed31547ab3138b6c18ecc423b1ae5405 but was limited for `disable_animation` option only. This PR extend the middleware setup to a global config so we can benefit from the design and add customized middleware to the stack. This is very useful when we have test specific logs or debugging tools that needs to wrap between requests 🙇‍♂️


twalpole:
You can just add middleware to the rack app you tell Capybara to run, what justifies adding the extra complexity to Capybara? We want to start getting rid of some of the config options...

PR #2687, opened 2023-07-18
table_row selector matches cell values in order when passed an array …

### LINKED ISSUE #2686: table_row selector does not work with an array longer than two values
## Meta
Capybara Version: 3.40.0

## Expected Behavior

`have_selector(:table_row, %w[Thomas Walpole Oceanside])` should match against a row with those values

## Actual Behavior
The selector works if there are only two values as demonstrated in the tests but fails if there are more than two values.

## Steps to reproduce
Add the third column to the test as shown here: https://github.com/jparr/capybara/commit/e7bac39e0855eae5c815c942b10ab29c7c296ccd

If you look at how this is implemented for the `:table, with_rows: [...]` selector you can see that it builds a nested xpath here: https://github.com/teamcapybara/capybara/blob/5521da18729c3337d63351e7c4a59faf13538837/lib/capybara/selector/definition/table.rb#L106

The table_row selector builds a different shape for its xpath. The commit above makes the necessary change.

jparr:
…Fixes #2686

This corrects the shape of the generated xpath for matching against an array of values. 

If you look at how this is implemented for the :table, with_rows: [...] selector you can see that it builds a nested xpath here: https://github.com/teamcapybara/capybara/blob/5521da18729c3337d63351e7c4a59faf13538837/lib/capybara/selector/definition/table.rb#L106

### DIFF
--- a/lib/capybara/selector/definition/table_row.rb
+++ b/lib/capybara/selector/definition/table_row.rb
@@ -14,7 +14,7 @@ Capybara.add_selector(:table_row, locator_type: [Array, Hash]) do
     else
       initial_td = XPath.descendant(:td)[XPath.string.n.is(locator.shift)]
       tds = locator.reverse.map { |cell| XPath.following_sibling(:td)[XPath.string.n.is(cell)] }
-                   .reduce { |xp, cell| xp.where(cell) }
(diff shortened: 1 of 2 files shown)


twalpole:
Thanks

PR #2688, opened 2023-07-18
table_row and with_rows ensure that the header matches locator for first column

### LINKED ISSUE #2685: Row selectors ignore the value of the first header
## Meta
Capybara Version: 3.40.0

## Expected Behavior

`have_table('Caption', with_rows: [{'this does not matter' => 'cell value'}]` should only match when the header matches the key passed.

`have_selector(:table_row, 'this does not matter' => 'Thomas')` should only match when the header matches the key passed.

## Actual Behavior
both selectors ignore the text of the header when it is in the first column

## Steps to reproduce
I wrote some tests here: https://github.com/jparr/capybara/commit/03e2f7502444509a961666a7bb2fb31790104cc3

Both matchers do some previous sibling counting to verify the position of the header and cell which resolve to zero for the first column when there is a match and zero when there is no match.

jparr:
Fixes #2685

This adds a check to ensure that the `header_xp` exists before checking the that the position of the cell matches the position of the header. If `header_xp` does not exist then `header_xp.preceding_sibling.count.plus(1)` will still equal 1 and `XPath.position.equals` will be true for the first column.

### DIFF
--- a/lib/capybara/selector/definition/table.rb
+++ b/lib/capybara/selector/definition/table.rb
@@ -93,7 +93,7 @@ Capybara.add_selector(:table, locator_type: [String, Symbol]) do
     row.map do |header, cell|
       header_xp = XPath.ancestor(:table)[1].descendant(:tr)[1].descendant(:th)[XPath.string.n.is(header)]
       XPath.descendant(:td)[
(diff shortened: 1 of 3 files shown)


twalpole:
Thank you

PR #2696, opened 2023-08-07
Do not auto-submit forms if they have multiple inputs

mitchellhenke:
Thanks for capybara!

I am attempting to update to the latest version, but am having some test failures from the change in https://github.com/teamcapybara/capybara/commit/6b05e5cb30e9807483c2641e9a60d89692bb0487 as I have forms where I set multiple inputs in the same form to values that end with `\n`.

I think ideally I'd be able to opt-out of the auto-submit behavior, but I'm unsure what the preferred path is. I'm happy to make updates to the branch here as well. Thanks again for all the work that goes into capybara!

### DIFF
--- a/lib/capybara/rack_test/node.rb
+++ b/lib/capybara/rack_test/node.rb
@@ -225,7 +225,7 @@ private
       native.remove
     else
       value.to_s.tap do |set_value|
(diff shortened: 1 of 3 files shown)


twalpole:
This is how it was supposed to be in the first place - thanks for catching

mitchellhenke:
Thank you so much!

PR #2701, opened 2023-08-18
Do not pass an explicit nil CONTENT_TYPE in the rack_test browser

jeremyevans:
This violates the Rack SPEC and causes failures if the application you are testing uses Rack::Lint.

I could not find any existing specs that test the environment that capybara passes to rack-test, so this does not include specs.

### DIFF
--- a/lib/capybara/rack_test/browser.rb
+++ b/lib/capybara/rack_test/browser.rb
@@ -35,12 +35,13 @@ class Capybara::RackTest::Browser
     path = request_path if path.nil? || path.empty?
     uri = build_uri(path)
     uri.query = '' if method.to_s.casecmp('get').zero?
+    env = {'HTTP_REFERER' => referer_url}
+    env['CONTENT_TYPE'] = content_type if content_type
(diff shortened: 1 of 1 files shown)


twalpole:
Would you mind pointing out where this violates the rack spec -- I'm likely missing something but I don't see any such restriction in https://github.com/rack/rack/blob/main/SPEC.rdoc#rack-applications-

jeremyevans:
> Would you mind pointing out where this violates the rack spec -- I'm likely missing something but I don't see any such restriction in https://github.com/rack/rack/blob/main/SPEC.rdoc#rack-applications-

Sure: "The CGI keys (named without a period) must have String values."

jeremyevans:
Note that the Rack SPEC is extracted from lib/rack/lint.rb, and Rack::Lint exists solely to check that servers/middleware/applications are compliant with the Rack SPEC.  Here's the line showing the problem in Rack::Lint: https://github.com/rack/rack/blob/main/lib/rack/lint.rb#L311

PR #2708, opened 2023-09-23
Introduce `data:` attribute filter

seanpdoyle:
The Rails provided `tag` and `content_tag` methods will transform [`data:` options][rails-data-option] into `data-` prefixed counterparts.

For example:

```ruby
      tag.div data: { value: 1 }
 # => <div data-value="1"></div>
```

Rails will even JSON-encode `Hash` and `Array` instances:

```ruby
      tag.div data: { object: { value: 1 }, array: [1, 2] }
 # => <div data-object="{&quot;value&quot;:1}" data-array="[1, 2]"></div>
```

This commit introduces similar filter-level support for a global `data:` filter option to transform `Hash` instances into compound `String` keys:

```ruby
 #=> <div data-object="{&quot;value&quot;:1}" data-array="[1, 2]"></div>
 
 expect(page).to have_css "div", data: { object: { value: 1 }, array: [1, 2] } }
```

[rails-data-option]: https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-tag-label-Options

### DIFF
--- a/History.md
+++ b/History.md
@@ -3,6 +3,7 @@ Release date: unreleased
 
 * Dropped support for Ruby 2.7, 3.0+ is now required
 * Dropped support for Selenium < 4.8
+* Support `:data` system filter option
 
 # Version 3.39.2
 Release date: 2023-06-10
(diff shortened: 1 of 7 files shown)


twalpole:
This is rails specific behavior and doesn't belong in the default base Capybara

PR #2709, opened 2023-09-23
Introduce `aria:` attribute filter

seanpdoyle:
The Rails provided `tag` and `content_tag` methods will transform [`aria:` options][rails-aria-option] into `aria-` prefixed counterparts.

For example:

```ruby
      tag.div aria: { label: "Label" }
 # => <div aria-label="Label"></div>
```

Rails will even JSON-encode `Hash` and `Array` instances:

```ruby
      tag.div aria: { labelledby: ["an_id", "another_id"] }
 # => <div aria-labelledby="an_id another_id"></div>
```

This commit introduces similar filter-level support for a global `aria:` filter option to transform `Hash` instances into compound `String` keys.

```ruby
 # => <div aria-labelledby="an_id another_id" aria-selected="true"></div>

 expect(page).to have_css "div", aria: { labelledby: ["an_id", "another_id"], selected: "true" }
```

[rails-aria-option]: https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-tag-label-Options

### DIFF
--- a/History.md
+++ b/History.md
@@ -3,6 +3,7 @@ Release date: unreleased
 
 * Dropped support for Ruby 2.7, 3.0+ is now required
 * Dropped support for Selenium < 4.8
+* Support `:aria` system filter option
 
 # Version 3.39.2
 Release date: 2023-06-10
--- a/lib/capybara/queries/selector_query.rb
+++ b/lib/capybara/queries/selector_query.rb
@@ -9,7 +9,7 @@ module Capybara
 
       SPATIAL_KEYS = %i[above below left_of right_of near].freeze
       VALID_KEYS = SPATIAL_KEYS + COUNT_KEYS +
-                   %i[text id class style visible obscured exact exact_text normalize_ws match wait filter_set focused]
+                   %i[text id class style visible obscured exact exact_text normalize_ws match wait filter_set focused aria]
       VALID_MATCH = %i[first smart prefer_exact one].freeze
 
       def initialize(*args,
@@ -79,6 +79,7 @@ module Capybara
 
         desc << " with id #{options[:id]}" if options[:id]
         desc << " with classes [#{Array(options[:class]).join(',')}]" if options[:class]
+        desc << " with aria #{options[:aria]}" if options[:aria]
         desc << ' that is focused' if options[:focused]
         desc << ' that is not focused' if options[:focused] == false
 
@@ -446,6 +447,7 @@ module Capybara
         matches_visibility_filters?(node) &&
           matches_id_filter?(node) &&
           matches_class_filter?(node) &&
+          matches_aria_filter?(node) &&
           matches_style_filter?(node) &&
           matches_focused_filter?(node) &&
           matches_text_filter?(node) &&
@@ -511,6 +513,21 @@ module Capybara
         return true unless use_default_focused_filter?
 
         (node == node.session.active_element) == options[:focused]
+      end
+
+      def matches_aria_filter?(node)
+        options[:aria].to_h.all? do |key, filter|
+          value = node["aria-#{key}"]
+          filter = filter.join(' ') if filter.is_a?(Array)
+
+          case filter
+          when Regexp then filter.match?(value)
+          when true then !value.nil?
+          when false then value.nil?
+          else
+            value.to_s == filter.to_s
+          end
+        end
       end
 
       def need_to_process_classes?
--- a/lib/capybara/selector.rb
+++ b/lib/capybara/selector.rb
@@ -8,6 +8,7 @@ require 'capybara/selector/definition'
 # All Selectors below support the listed selector specific filters in addition to the following system-wide filters
 #   * :id (String, Regexp, XPath::Expression) - Matches the id attribute
 #   * :class (String, Array<String | Regexp>, Regexp, XPath::Expression) - Matches the class(es) provided
+#   * :aria (Hash<Symbol, Array | String | true | false>) - Matches the aria-* prefixed attributes provided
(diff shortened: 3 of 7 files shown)


twalpole:
This is Rails specific formatting behavior and doesn't belong in base Capybara

seanpdoyle:
@twalpole thank you for that guidance. Is there a way to define a global filter?

I'm happy to open a PR to Rails for this diff as well as https://github.com/teamcapybara/capybara/pull/2708, but I'm unsure about what the integration touchpoint would be.

seanpdoyle:
@twalpole I've opened https://github.com/rails/rails/pull/49665. Outside of feedback on Rails-ism, I'm particularly interested in whether or not it extends Capybara through its "public" API. If you're available and interested, I'd appreciate review on the way it installs the expression filters.

twalpole:
@seanpdoyle Yes `modify_selector` is public API, however I don't believe those changes belong as part of Rails either.  IMO they should be part or a separate gem users can opt to use or not until it's proven out they're valuable changes and are stable, non problem causing.  There's also a chance the `aria` part could conflict with Capybara in the future which, if included directly in rails, would lock people out of Capybara updates.  The `data` attribute I don't like because it encourages people to test implementation rather than results.

seanpdoyle:
@twalpole thank you for sharing that guidance.

> The `data` attribute I don't like because it encourages people to test implementation rather than results.

Projects I've worked on have had a lot of success using Capybara selectors outside of System Testing environments. For example, we have Action View and Action Dispatch Integration Tests that use Capybara selectors instead of [rails-dom-testing](https://github.com/rails/rails-dom-testing).

In those environments, tests aren't driving the page (or even "visiting" the page) at all, they're strictly making assertions about structure from static HTML without JS or CSS.

Having access to semantically meaningful built-in selectors like `:link`, `:button`, etc. combined with the ability to define our own semantically meaningful selectors (or depend on pre-defined selectors like [capybara_accessible_selectors](https://github.com/citizensadvice/capybara_accessible_selectors) has been **immeasurably** valuable when compared. The thought process has been: "our test suites already depend on Capybara, and use the same meaningful selectors in browser tests, why not share the same testing language wherever HTML is present"?

In static HTML tests, we've found the higher-level abstractions really valuable (`:link`, `:button`, `:field`), **especially** when the same assertion can be made with lower level primitives. For example, `assert_checked_field "Enable the feature"` instead of `assert_field "Enable the feature", type: "checkbox", checked: true` or even `assert_selector :element, "input", type: "checkbox", checked: true`.

Having acknowledged that, there are still times where we want to be specific about `aria-` and `data-` HTML attributes when a higher-order selector is unavailable. This is especially true when writing view and integration tests for applications that utilize `data-` attribute driven frameworks like [Turbo](https://turbo.hotwired.dev) and [Stimulus](https://stimulus.hotwired.dev). Having the thorough coverage that asserts that an element has a `[data-controller]` attribute matching a value at the view test layer frees up time and attention to test less rigorously at the browser test layer, where tests are more expensive.

In my experience, applications that use Capybara at the browser test layer **should actively discourage** asserting particular `[data-*]` or `[class]` or `[id]` attributes, and should instead encourage the higher-order selectors that have end-user facing semantics. Sometimes, `[aria-]` and `[role]` attributes **are** end-user facing, so that can be tricky and usually requires a judgment call instead of a blanket rule.

All this is to say that Capybara has become **the most important dependency** on most Rails applications that I work on that server-generate HTML. That isn't hyperbole. The fact that selectors can be available at every layer combined with it's sophistication make Capybara irreplaceable. I have a sense that utilizing the selector system in this kind of way is at odds with Capybara's stated goals, and the fact that it prioritizes browser environment tests above all else.

Lately, I've wished for a version of Capybara that exists on a continuum between the gem as its structured today and a gem like [teamcapybara/xpath](https://github.com/teamcapybara/xpath).

Is there any interested in making it possible to consume Capybara in separate layers?

* the layer that abstracts over Nokogiri with `XPath` and its selector system
* the layer that abstracts over Selenium / Webdriver / etc with its browser-driving behavior

If not, is there interest in improving support for Capybara in browser-less environments?

twalpole:
I believe what you've written reinforces that the changes/enhanvements you're pushing for should really be in their own gem which people would then be free to include into the projects if they want the same functionality you're going for.

seanpdoyle:
@twalpole thank you for that feedback.

> In spite of the tremendous value we get out of using the tool in this way, I have the feeling that how we've utilized the selector system is at odds with Capybara's stated goals, and that we're pulling in opposite directions.

Am I correct in my understanding of where Capybara fits into a broader test suite?

> the changes/enhancements you're pushing for should really be in their own gem 

Is [Capybara::Selector](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selector/selector.rb) part of the gem's public interface? Is `Capybara::Selector.all` a reliable way for third-party gem's to extend Capybara, or is that breaching layers of encapsulation?

twalpole:
I see Capybara providing the most value with browser based tests.  The selectors can be used in view tests, and are useful there, however I see browser based behavior tests being much more valuable than view tests.  Testing that views have specific data-* values in elements is meaningless if you're not actually testing that those data-* values are correctly used by the JS in the browser, and if you are testing the behavior in the browser, what's the point of testing they exist in the view?

Yes Capybara::Selector is public interface, although I wouldn't use Capybara::Selector.all to loop through and modify all the selectors unless you give people a way to opt out of modifying specific selectors. There's way too high a chance of conflicting with peoples custom selectors.  As a user I think I would actually prefer an opt-in per selector configuration.

PR #2715, opened 2023-11-04
Capybara::Minitest::Assertions: Expose definition helpers

seanpdoyle:
Define module-level `.asserts`, `.asserts_no`, `.assert_selector`, and `.asserts_no_selector` methods to abstract how assertions are defined.

Then, define the out-of-the-box assertions in terms of these new methods, while documenting how to invoke them from consumer-side applications for bespoke selectors.


twalpole:
I'm not interested in expanding the public API around this.

PR #2721, opened 2023-11-30
Update README

artur-intech:
- Do not mention about dismissing an alert. You cannot really dismiss it since there is only one button named "OK".
- Add more examples on how to dismiss `confirm` and `prompt` dialogs.
- Enhance other examples with optional `text` arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -618,10 +618,10 @@ JS
 
 In drivers which support it, you can accept, dismiss and respond to alerts, confirms, and prompts.
 
-You can accept or dismiss alert messages by wrapping the code that produces an alert in a block:
+You can accept alert messages by wrapping the code that produces an alert in a block:
 
 ```ruby
-accept_alert do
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -618,26 +618,38 @@ JS
...
 
 ```ruby
-accept_prompt(with: 'Linus Torvalds') do
+accept_prompt('optional text', with: 'Linus Torvalds') do
+  click_link('Show Prompt About Linux')
+end
+```
+
+```ruby
+dismiss_prompt('optional text', with: 'Linus Torvalds') do
```
twalpole, line 652: Doesn't really make sense to pass a `with` to dismiss_prompt

twalpole (changes requested):

README.md
artur-intech: Fixed!

Issue #2725, opened 2023-12-08
Chromedriver option needs update since chrome 120


aandis:
## Meta
Capybara Version: 3.39.2

The following code breaks -
```
Capybara.javascript_driver = :selenium_chrome_headless
page.driver.browser.logs.get(:browser)

#=> undefined method `logs' for #<Selenium::WebDriver::Chrome::Driver:0x75d3dfd5c1800834 browser=:"chrome-headless-shell">
```

Works if your chrome version is < 120 though

See
- https://github.com/SeleniumHQ/selenium/issues/13112
- https://www.selenium.dev/blog/2023/headless-is-going-away/

aandis:
https://github.com/teamcapybara/capybara/blob/6267ff65a294a7dcbe9d7281c30cc2d91b08b093/lib/capybara/registrations/drivers.rb#L35 needs to be changed to `--headless=new`

PR #2726, opened 2023-12-08
Use the new headless Chrome on the pre-registered driver

neilvcarvalho:
Earlier this year (2023), Chrome released Chrome for Testing, a new Chrome flavor that targets testing and automation use cases. This flavor has been accessible using the `--headless=new` option since Chrome 109, released on January 1, 2023. This new mode supports the full browser capabilities, including running extensions.

The default headless mode is still pointed to `--headless=old` for now, but it's planned to be changed to `new` over time.

This commit explicitly chooses the `new` headless mode, that uses Chrome for Testing.

Solves #2725


jagthedrummer:
Just curious if there's an ETA on releasing a new version with this fix.

neilvcarvalho:
For folks following this and waiting for a gem update: Selenium released this week a version [supporting the old headless mode](https://github.com/SeleniumHQ/selenium/pull/13271)

simi:
IMHO it is too early to make this default. New headless mode broke (or make flaky) a lot of Chrome automated usage I have seen (random, not even using Selenium or Capybara).

twalpole:
Then you have the freedom to setup your own config

PR #2734, opened 2024-01-12
Add `Capybara::Minitest::Assertions#within`

seanpdoyle:
Introduce support for `#within` inside the
`Capybara::Minitest::Assertions` module.

This method brings support for scoping assertions when included in tests
without `Capybara::DSL` and outside of a `Capybara.current_session`
context.

```ruby
class MinitestAssertionsTest < Minitest::Test

  attr_reader :page

  def test_within
    @page = Capybara.string <<~HTML
      <div id="outside">Outside</div>
      <div id="parent">
        Inside
        <div id="child">Child</div>
        <div id="sibling">Sibling</div>
      </div>
    HTML

    within :element, id: "parent" do
      assert_text "Inside"
      assert_text "Child"
      assert_text "Sibling"
      assert_no_text "Outside"

      within :element, id: "child" do
        assert_text "Child"
        assert_no_text "Inside"
        assert_no_text "Outside"
        assert_no_text "Sibling"
      end

      assert_raises Capybara::ElementNotFound do
        within :element, id: "outside"
      end
    end
  end
end
```

### DIFF
--- a/lib/capybara/minitest.rb
+++ b/lib/capybara/minitest.rb
@@ -2,10 +2,17 @@
 
 require 'minitest'
 require 'capybara/dsl'
+require 'capybara/scoping'
 
 module Capybara
   module Minitest
     module Assertions
+      include Capybara::Scoping
+
+      def root_scope
+        page
+      end
+
       ##
       # Assert text exists
       #
@@ -384,7 +391,7 @@ module Capybara
         when ->(arg) { arg.respond_to?(:to_capybara_node) }
           [args.shift.to_capybara_node, args]
         else
-          [page, args]
+          [current_scope, args]
         end
       end
 
--- a/History.md
+++ b/History.md
@@ -4,6 +4,10 @@ Release date: unreleased
 * Dropped support for Ruby 2.7, 3.0+ is now required
 * Dropped support for Selenium < 4.8
 
+### Added
+
+* Introduce `Capybara::Scoping` to share `#within` between `Capybara::Session` and `Capybara::Minitest::Assertions`
+
 # Version 3.39.2
 Release date: 2023-06-10
 
--- a/lib/capybara/scoping.rb
+++ b/lib/capybara/scoping.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+module Capybara
+  module Scoping
+    ##
+    #
+    # Executes the given block within the context of a node. {#within} takes the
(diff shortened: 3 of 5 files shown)


seanpdoyle:
To highlight some prior art that would stand to benefit from this built-in support:

* [ActionView::TestCase](https://edgeguides.rubyonrails.org/testing.html#testing-view-partials) could support Capybara assertions outside of a request-response cycle
* [Action Mailer Tests](https://edgeguides.rubyonrails.org/testing.html#the-basic-test-case) could utilize `#within` in a similar way
* the View Component package originally [implemented a `ViewComponent::CapybaraSimpleSession` class](https://github.com/ViewComponent/view_component/pull/1389) to make `#within` (and other helpers) available to [ViewComponent::TestCase](https://viewcomponent.org/guide/testing.html). That bespoke implementation [was later reverted](https://github.com/ViewComponent/view_component/pull/1427).



This kind of built-in support would enable those types of use cases.

dorianmarie (approved):
Looking good, could we turn:

```
  def render(html)
    @page = Capybara.string(html)
  end
```

Into an helper like `Capybara.render`

Also what about rspec?

lib/capybara/minitest.rb
```diff
@@ -2,10 +2,17 @@
...
+require 'capybara/scoping'
 
 module Capybara
   module Minitest
     module Assertions
+      include Capybara::Scoping
+
+      def root_scope
+        page
+      end
```
seanpdoyle, line 14: @twalpole does this feel like an appropriate abstraction to introduce to share this behavior?

twalpole:
@seanpdoyle you seem fixated on trying to make Capybara do more than it's primary focus. I'm more interested in focusing purely on testing of web applications how users would interact with them. As I've stated before it seems like you'd be better off writing your own gem for extending functionality in the ways you find value in.

G-Rath:
@seanpdoyle do you know if it's possible to make this usable in same way without requiring changes within Capybara?

I personally have been on the lookout for something similar to [testing library](https://testing-library.com/) in Rails to enable more robust DOM testing and smaller individual tests while keeping overhead low, which I think Capybara in its current form almost fully realizes but there's still a few aspects that hold it back from being an easy experience, this being one of them (and maybe the only one?); so I'd love to hear if you know of a way to make this work with today's Capybara.

Issue #2745, opened 2024-02-06
Selenium raises WebDriverError, not JavascriptError, when access is denied to clear sessionStorage


DanielHeath:
## Meta
Capybara Version: 3.40.0
Driver Information (and browser if relevant): selenium-webdriver (4.17.0), chrome 121.0.6167.139

## Expected Behavior
Visiting `about:blank` and then calling `clear_storage` should silently fail (thanks to the `rescue JavascriptError` line in lib/capybara/selenium/driver.rb:351 )

## Actual Behavior
```
     Selenium::WebDriver::Error::WebDriverError:
       <unknown>: Failed to read the 'sessionStorage' property from 'Window': Access is denied for this document.
         (Session info: chrome-headless-shell=120.0.6099.224)
     # #0 0x5558656adf83 <unknown>
     # #1 0x555865366cf7 <unknown>
     # #2 0x55586536d2a3 <unknown>
     # #3 0x55586536fbb4 <unknown>
     # #4 0x5558653ffba3 <unknown>
     # #5 0x5558653e00b2 <unknown>
     # #6 0x5558653ff006 <unknown>
     # #7 0x5558653dfe53 <unknown>
     # #8 0x5558653a7dd4 <unknown>
     # #9 0x5558653a91de <unknown>
     # #10 0x555865672531 <unknown>
     # #11 0x555865676455 <unknown>
     # #12 0x55586565ef55 <unknown>
     # #13 0x5558656770ef <unknown>
     # #14 0x55586564299f <unknown>
     # #15 0x55586569b008 <unknown>
     # #16 0x55586569b1d7 <unknown>
     # #17 0x5558656ad124 <unknown>
     # #18 0x7fbc6bc93ac3 <unknown>
     # .gem/ruby/3.2.2/gems/selenium-webdriver-4.17.0/lib/selenium/webdriver/remote/response.rb:55:in `assert_ok'
     # .gem/ruby/3.2.2/gems/selenium-webdriver-4.17.0/lib/selenium/webdriver/remote/response.rb:34:in `initialize'
     # .gem/ruby/3.2.2/gems/selenium-webdriver-4.17.0/lib/selenium/webdriver/remote/http/common.rb:83:in `new'
     # .gem/ruby/3.2.2/gems/selenium-webdriver-4.17.0/lib/selenium/webdriver/remote/http/common.rb:83:in `create_response'
[...]

PR #2746, opened 2024-02-06
Handle "access denied" when clearing session storage

DanielHeath:
Fixes https://github.com/teamcapybara/capybara/issues/2745


twalpole:
Thanks, handled with a single `rescue` in another commit

PR #2751, opened 2024-02-19
Remove a has_element example with wrong usage.

YusukeIwaki:
https://github.com/teamcapybara/capybara/pull/2700 introduced a useful matcher.
However its spec have a wrong usage

```
expect(@session).to have_element('a', text: :'A link', href: :'/with_simple_html')
```

It generates a XPath query `(.//*[(local-name(.) = 'a')][(./@href = '/with_simple_html')])[]`.
**It doesn't have `text: A link` condition since `Capybara::Queries::SelectorQuery` assumes the text option to be a String or a Regexp**.

So I just removed the line because no other specs have `text: :xxx` (Symbol).


twalpole:
Thanks

PR #2758, opened 2024-03-27
Consistently support passing a symbol as a test_id

rzane:
I found a few places where passing a symbol as a test ID doesn't work as expected. All of the other selector types support symbols as for test ID selectors.


twalpole:
Thank you

PR #2768, opened 2024-08-16
Disable the Chrome search engine selector by default

dgmstuart:
In recent versions of Chrome, a search engine selection modal is shown when first opening the browser, which causes specs to fail  with errors like this:

    Selenium::WebDriver::Error::UnknownError:
            unknown error: failed to close window in 20 seconds
              (Session info: chrome=128.0.6613.36)

Presumably this is because the spec is trying to interact with the browser but can't get past the selection screen.

This seems to be a common issue:

- https://stackoverflow.com/questions/78831994/how-to-skip-choose-search-engine-with-capybara-selenium-chrome
- https://www.reddit.com/r/learnpython/comments/1eoapcb/selenium_dosent_find_google_search_engine_option/

Adding this option appears to fix it, but I've only tested by defining my own driver locally with this argument - not sure how I'd write a spec for this driver definition?

Also I'm not sure if this option is defined for older versions of Chrome (or is it chromedriver?) - would this result in errors when using a version where this option isn't supported?

### DIFF
--- a/lib/capybara/registrations/drivers.rb
+++ b/lib/capybara/registrations/drivers.rb
@@ -23,6 +23,7 @@ Capybara.register_driver :selenium_chrome do |app|
   browser_options = Selenium::WebDriver::Chrome::Options.new.tap do |opts|
     # Workaround https://bugs.chromium.org/p/chromedriver/issues/detail?id=2650&q=load&sort=-id&colspec=ID%20Status%20Pri%20Owner%20Summary
     opts.add_argument('--disable-site-isolation-trials')
+    opts.add_argument('--disable-search-engine-choice-screen')
   end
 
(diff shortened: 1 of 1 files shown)


vrodic:
I confirm this is a problem specifically on MacOS. It breaks headless tests. On Linux the latest Chrome doesn't show this nag screen. But it's safe to add this option on all platforms.

dgmstuart:
> I confirm this is a problem specifically on MacOS. It breaks headless tests. On Linux the latest Chrome doesn't show this nag screen. But it's safe to add this option on all platforms.

Interesting - If it's not there yet on Linux, then I guess it's coming - seems like it's a legal requirement that's been imposed on them by the EU so I'd expect it to eventually be on all platforms: https://www.google.com/chrome/choicescreen/

twalpole:
Github wouldn't let me fix conflicts so I copied into a different change

PR #2794, opened 2025-02-11
Ensure puma config load process is respected

joshuay03:
Ref: https://github.com/puma/puma/pull/3616

In v7, puma will be validating that its config objects are both loaded (loads config files) and clamped (resolves config procs and finalizes defaults) before its options are accessed to begin the launch process. This will be a breaking change for users utilising capybara as it builds its own puma config, but only clamps it. This change ensures a future version of capybara can be compatible with puma v7, while also being backward compatible with previous versions.

**Note:** The side effect of this change is that options specified in users' puma config files that are not specified in capybara's [user options](https://github.com/teamcapybara/capybara/blob/0480f90168a40780d1398c75031a255c1819dce8/lib/capybara/registrations/servers.rb#L40-L41) will be [taken into account](https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/server.rb#L87-L100) when capybara boots its server. ~A workaround would be to only pass in the user options hash [here](https://github.com/teamcapybara/capybara/blob/0480f90168a40780d1398c75031a255c1819dce8/lib/capybara/registrations/servers.rb#L61) (`conf.options.user_options`) so only those options will be picked up by puma and [merged with its default options](https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/server.rb#L84). Let me know what you think.~ See https://github.com/teamcapybara/capybara/pull/2794#issuecomment-2661172136.

### DIFF
--- a/spec/server_spec.rb
+++ b/spec/server_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
+require 'puma'
 
 RSpec.describe Capybara::Server do
   it 'should spool up a rack server' do
@@ -102,7 +103,6 @@ RSpec.describe Capybara::Server do
   it 'should call #clamp on the puma configuration to ensure that environment is a string' do
     Capybara.server = :puma
     app_proc = proc { |_env| [200, {}, ['Hello Puma!']] }
-    require 'puma'
     allow(Puma::Server).to receive(:new).and_wrap_original do |method, app, events, options|
       # If #clamp is not called on the puma config then this will be a Proc
       expect(options.fetch(:environment)).to be_a(String)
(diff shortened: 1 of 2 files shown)


joshuay03:
> A workaround would be to only pass in the user options hash [here](https://github.com/teamcapybara/capybara/blob/0480f90168a40780d1398c75031a255c1819dce8/lib/capybara/registrations/servers.rb#L61) (`conf.options.user_options`) so only those options will be picked up by puma and [merged with its default options](https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/server.rb#L84). Let me know what you think.

Actually, a much simpler workaround would be to specify `config_files: ['-']` with the user options which will take precedence over the [default file lookup](https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/configuration.rb#L251-L262). I've gone ahead and actioned this.

spec/server_spec.rb
```diff
@@ -118,6 +118,25 @@
...
+    File.open('config/puma.rb', 'w') do |file|
+      file.puts 'debug' # puma default is debug: false, which we don't override
+    end
+    allow(Puma::Server).to receive(:new).and_call_original
+    app_proc = proc { |_env| [200, {}, ['Hello Puma!']] }
+    server = described_class.new(app_proc).boot
+    expect(Puma::Server).to have_received(:new).with(
+      anything,
+      anything,
+      satisfy { |opts| opts.final_options[:debug] == false }
```
joshuay03, line 133: With `load` but without `config_files` override:
Screenshot 2025-02-16 at 12 13 06 pm

joshuay03:
Bumping this, cc @twalpole

joshuay03:
This is no longer needed, see: https://github.com/puma/puma/pull/3616#issuecomment-3200854300

PR #2809, opened 2025-05-12
Remove unused `cgi` require for Ruby 3.5

Earlopain:
On Ruby 3.5, requiring `cgi` will emit a warning. `capybara` uses nothing from it, so just remove it.

https://bugs.ruby-lang.org/issues/21258

### DIFF
--- a/lib/capybara/rack_test/driver.rb
+++ b/lib/capybara/rack_test/driver.rb
@@ -4,7 +4,6 @@ require 'rack/test'
 require 'rack/utils'
 require 'mini_mime'
 require 'nokogiri'
-require 'cgi'
 
 class Capybara::RackTest::Driver < Capybara::Driver::Base
   DEFAULT_OPTIONS = {


Earlopain:
Thanks

tomstuart:
Would it be possible to get a release containing this change so we’re ready for the release of Ruby 4.0 next week?

artur-intech:
FYI: The warning is displayed in the version 4.0.2 already. Based on the phrase "The users need to run gem install cgi after Ruby 3.5 if they want to use CGI library" I conclude that it will be removed starting from that version, not deprecated.

simi:
Yeah, I think Ruby 3.5 was actually renamed to 4.0.

artur-intech:
@twalpole What needs to be done to release what has been merged so far?

simi:
I assume we need passing CI and @twalpole or @jnicklas to do the release.

PR #2811, opened 2025-06-08
Run CI for master branch push and PRs.

simi:


voxik:
Looking into the CI failures, specifically to [rack_3 (3.2, spec_rack)](https://github.com/teamcapybara/capybara/actions/runs/15519323258/job/43690442545?pr=2811#logs), I believe this is due to https://github.com/sinatra/sinatra/commit/cd3e00de20ddaff34ea30f7a74a7b9dad189d1d8 introduced by (Sinatra /) Rack::Protection 4.1.0, where the `www.example.com` is not on the list of enabled hosts.

simi:
Thanks @voxik, I seen also some date related issues. Locally CI looks much better. I'll keep exploring, feel free to share patches (or open PR to merge into this branch) to fix any test you're aware of why is failing.

simi:
Any idea what to use instead of example.com or better to re-configure to enable?

voxik:
Already looks better 👏 At least something is passing 🥳

simi:
Ahh, I think we can remove 3.0. WDYT?

voxik:
> Ahh, I think we can remove 3.0. WDYT? Or worth to keep?

For me personally (and looking into this due to Fedora), I don't mind removing the 3.0 support. Nevertheless I can't say the xpath code would be more readable with just `&` instead of `&block`, so I wonder if the change is worth of it.

However, wearing my Red Hat fedora, RHEL users would likely not mind longer support for Ruby 3.0 which is default in RHEL 9 and will be supported for years. But I don't think I can stop the train (after all, RHEL 10 is out fresh) 😉

simi:
OK, I'll revert in xpath for now. I just blindly follower rubocop in there.

PR #2815, opened 2025-08-07
Start getting this repo back to where it should be

twalpole:


twalpole:
@simi Great, I'll take a look, thanks

lib/capybara/registrations/drivers.rb
```diff
@@ -36,6 +36,11 @@
     opts.add_argument('--disable-gpu') if Gem.win_platform?
     # Workaround https://bugs.chromium.org/p/chromedriver/issues/detail?id=2650&q=load&sort=-id&colspec=ID%20Status%20Pri%20Owner%20Summary
     opts.add_argument('--disable-site-isolation-trials')
+
+    # https://github.com/teamcapybara/capybara/issues/2796#issuecomment-2678172710
+    opts.add_argument('disable-background-timer-throttling')
```
connorshea, line 41: Out of curiosity, is it not necessary to have the `--` before the flag name here? Does this still work?

PR #2817, opened 2025-08-11
Validate whether just catching here is enough

twalpole:


connorshea:
I tried running my [test suite for vglist](https://github.com/connorshea/vglist/tree/b34f4c7b451286ce1d8619fb7e5b54db663dfa80/spec/features) a few times after pointing my Gemfile at the capybara master branch, and unfortunately this doesn't seem to fully fix the issue. Regardless, thank you for the recent work updating the gem!

PR #2826, opened 2025-09-17
Fix `assert_selector` support for `Hash` locator

seanpdoyle:
Prior to this commit, the following style of assertion fails:

```ruby
assert_selector :table_row, "First Name" => "John", "Last Name" => "Doe"
```

The assertion cites that the `"First Name"` is an invalid option, signaling that it's being treated as a filter rather than a locator value:

```
ArgumentError: Invalid option(s) "First Name", should be one of :above,
  :below, :left_of, :right_of, :near, :text, :id, :class, :style,
  :visible, :obscured, :exact, :exact_text, :normalize_ws, :match, :wait,
  :filter_set, :focused
```

To resolve this issue, modify the private
`Capybara::Node::Matchers#_set_query_session_options` to treat the last positional argument as a positional argument (rather than a pseudo keyword argument) if it's a `Hash` with any `String` keys.

### DIFF
--- a/lib/capybara/node/matchers.rb
+++ b/lib/capybara/node/matchers.rb
@@ -892,7 +892,7 @@ module Capybara
       def _set_query_session_options(*query_args)
         query_args, query_options = query_args.dup, {}
         # query_options = query_args.pop if query_options.empty? && query_args.last.is_a?(Hash)
-        query_options = query_args.pop if query_args.last.is_a?(Hash)
+        query_options = query_args.pop if query_args.last.is_a?(Hash) && query_args.last.keys.none?(String)
         query_options[:session_options] = session_options
         [query_args, query_options]
       end
--- a/History.md
+++ b/History.md
@@ -1,3 +1,7 @@
+### Fixed
+
+* `assert_selector` can accept `Hash` instance as locator [Sean Doyle]
+
 # Version 3.40.0
 Release date: 2024-01-26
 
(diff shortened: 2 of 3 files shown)


lib/capybara/node/matchers.rb
```diff
@@ -892,7 +892,7 @@ def _verify_text(type = nil, expected_text, **query_options) # rubocop:disable S
       def _set_query_session_options(*query_args)
         query_args, query_options = query_args.dup, {}
         # query_options = query_args.pop if query_options.empty? && query_args.last.is_a?(Hash)
-        query_options = query_args.pop if query_args.last.is_a?(Hash)
+        query_options = query_args.pop if query_args.last.is_a?(Hash) && query_args.last.keys.none?(String)
```
seanpdoyle, line 895: I'm pretty dissatisfied with this solution. Since it's related to positional-to-keyword argument coercion, I wonder if there's a more wholistic approach that could be take to resolve this. For example, this would still fail for symbol keys that aren't filter options:

```ruby
assert_selector :table_row, First: "John", Last: "Doe"
```

twalpole:
I don't think this is a good solution. Selectors are allowed to be defined that don't require a locator, so globally forcing something that might be a filter option to be a locator is a bad idea.

twalpole:
I believe current usage would need to be soemthing like:

    assert_selector :table_row, { "First Name" => "John", "Last Name" => "Doe" }, {}

twalpole:
https://github.com/teamcapybara/capybara/pull/2846 does the same but with more restrictions when it works (only when selector specifies support of a Hash locator).  It's still hacky but, I think,  less likely to break anyones existing code. Closing this PR

PR #2829, opened 2025-10-06
Fix kwargs in `assert_*_of_selectors` methods in minitest

Jealrock:
When calling `assert_*_of_selectors` on self in minitest the kwargs get pushed to args.

To reproduce replace `assert_all_of_selectors(:css, 'select#form_other_title', 'input#form_last_name')` with `assert_all_of_selectors(:css, 'select#form_other_title', "input#form_super_secret", visible: false)` in [minitest_spec.rb](https://github.com/teamcapybara/capybara/blob/b3325b198464b806f07ec2011ceb532d6d5cf4ab/spec/minitest_spec.rb#L113).

The fix is to just use the right generator method in [minitest.rb](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/minitest.rb)

### DIFF
--- a/lib/capybara/minitest.rb
+++ b/lib/capybara/minitest.rb
@@ -48,7 +48,26 @@ module Capybara
       # @!method assert_no_current_path
       #   See {Capybara::SessionMatchers#assert_no_current_path}
 
-      %w[text no_text title no_title current_path no_current_path].each do |assertion_name|
+      ##
+      # Assert all of the provided selectors exist on page
+      #
+      # @!method assert_all_of_selectors
+      #   See {Capybara::Node::Matchers#assert_all_of_selectors}
+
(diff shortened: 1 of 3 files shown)


twalpole:
The fact that 'select#form_other_title' is still passing in this test tells me that the 'visible:false' isn't actually being applied since that element is visible. I'm not sure why yet but this isn't valid as submitted. Also please don't remove the previous assertion from the tests just add a second one to the same test.

twalpole:
Actually,  ignore tthe above, been too long and I was misreading the test as written (don't check visibility as opposed to check for non visibility).

twalpole:
Merged at f8d38c29d26e5b3697fd483dace140945e7d19a3

Issue #2807, opened 2025-04-23
The default url host for testing is not same as mentioned in the doc.


aditya-vishwakarma-lab:
# Default URL Options Mismatch between Rails and Capybara

## Description
Rails 8.0.2 has changed its default test environment URL to `example.com`, but Capybara still defaults to `www.example.com`. This causes issues when testing mailer URLs in system tests.

### Current Behavior
- Rails default_url_options in test env: `example.com`
- Capybara default host: `www.example.com`
- This mismatch causes URL comparison failures in tests

### Example
```ruby
# config/environments/test.rb
config.action_mailer.default_url_options = { host: "example.com" }

# spec/system/some_mailer_spec.rb
expect(email.body.to_s).to include(some_url(...))  # generates www.example.com
# but email contains example.com
```

### Expected Behavior
Both Rails and Capybara should use the same default host to ensure consistent URL generation in tests.

## Proposed Solution
Either:
1. Update Capybara's default host to match Rails' `example.com`, or
2. Add documentation about this mismatch and how to handle it ( I know it has been mentioned in the [gotchas](https://github.com/teamcapybara/capybara/blob/master/README.md#gotchas) but I think that is incorrect, it assumes the default host in rails test environment is www.example.com )

## Environment
- Rails version: 8.0.2
- Capybara version: 3.40.o
- Ruby version: 3.4.1

## Additional Context
This issue particularly affects system tests that need to verify email content containing absolute URLs.

twalpole:
We won't be changing Capybaras default host currently since that could be a breaking change for a lot of people.  If you want to submit a PR with documentation updates you would consider sufficient it will be considered

twalpole:
Mreged documentation PR

PR #2830, opened 2025-10-15
docs: Correct default url options mismatch between capybara and rails

khhizr007:
Rails 8.0.2 has changed its default test environment URL to example.com, but Capybara still defaults to www.example.com. This causes issues when testing mailer URLs in system tests. Made changes to gotchas to clarify about the this mismatch.

This should address the issue of #2807


simi (approved):

khhizr007:
Hi @simi can we get this merged?

PR #2837, opened 2026-04-10
Add support for coordinates on `dragstart`, and related drag_to adjustments to work with react-dnd's HTML5 implementation

### LINKED ISSUE #2662: drag_to is broken in 3.39.0 due to `dragenter` event is dispatched without coordinates
## Meta
Capybara Version: 3.39.0

## Expected Behavior
The dispatched `dragenter` event contains the coordinates of the target.

## Actual Behavior
The dispatched `dragenter` event doesn't contais the coordinates of the target.

## Steps to reproduce
Fix of https://github.com/teamcapybara/capybara/issues/2576 doesn't include the mouse pointer coordinates causing issues in JS code listening to `dragenter` event and relying on those coordinates.

ragesoss:
I've had some drag-and-drop capybara specs in my test suite for https://github.com/WikiEducationFoundation/WikiEduDashboard that I had to disable years ago when we upgraded `react-dnd` to a version that uses HTML5. At the time, I couldn't find a way to make those specs work. I took a pass with Claude Code today to see if I could find a way to get the specs working, and ended up with a [drag helper](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/6707e1f1cd510fa9b512c402278101fd498cf596) that worked, in place of Capybara's `drag_to`.

I'm happy to keep that, but I figured since I had a good test case, I would try implementing an upstream fix to make Capybara's `drag_to` compatible with how my app uses `react-dnd`. Here's where I ended up. (Happy to squash these if you'd prefer, but I thought it better to start with the full history of the AI-written commits to be as transparent about the process for getting here as possible.)

I've tested this change locally, and I'm able to successfully run my long-disabled drag-and-drop spec, as long as `delay: 0` is added to the `drag_to` calls. It tool several iterations going between making changes to Capybara, making sure the old and new Capyabara specs still worked, and testing the changes against the spec in my project.

I think this fixes #2662, which was part of the root problem (but not all of it) for why my spec didn't work with HTML5 react-dnd.

Note the potential breaking change in the second commit: https://github.com/teamcapybara/capybara/commit/b305e2946cb12b0ee6b159954a71f366fa166ead

I don't know enough about Capybara or drag-and-drop to be confident about whether this is safe or reasonable, but it seems to make sense and I've watched both my own tests and the affected parts of the Capybara test suite run successfully with these changes.

### DIFF
--- a/lib/capybara/node/element.rb
+++ b/lib/capybara/node/element.rb
@@ -407,7 +407,11 @@ module Capybara
       # @param [Capybara::Node::Element] node     The element to drag to
       # @param [Hash] options  Driver specific options for dragging. May not be supported by all drivers.
       # @option options [Numeric] :delay   (0.05) When using Chrome/Firefox with Selenium and HTML5 dragging this is the number
-      #                                    of seconds between each stage of the drag.
+      #                                    of seconds between each stage of the drag. Libraries that use
+      #                                    requestAnimationFrame for hover scheduling (e.g. react-dnd) may require
+      #                                    `delay: 0` to prevent RAF callbacks and React effects from running
+      #                                    between events, which can cause listeners to be removed and reattached
+      #                                    mid-sequence.
       # @option options [Boolean] :html5   When using Chrome/Firefox with Selenium enables to force the use of HTML5
       #                                    (true) or legacy (false) dragging. If not specified the driver will attempt to
       #                                    detect the correct method to use.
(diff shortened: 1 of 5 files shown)


ragesoss:
Hmm... i see this didn't pass in CI, so i'll try to get that sorted.

ragesoss:
Okay! Here's a passing build that has this set of changes on top of the separate PR focused on making the build work: https://github.com/ragesoss/capybara/actions/runs/24291646486

And here's the build with just the changes from this PR, which fails but I think only for already-broken things: https://github.com/ragesoss/capybara/actions/runs/24291591529

ragesoss:
@simi as with the other commit, this one is all agentic code. i've tested it against my use case (and it doesn't break existing Capybara tests), but if it's not worth it to review, i completely understand; feel free to close.

simi:
@ragesoss I'll check, seems like a good feature.

PR #2843, opened 2026-07-07
Fix Rubocop violations

wynksaiddestroy:
I tried to fix all the Rubocop violations that cause the CI to fail.
I tried to make small commits to make the changes more understandable.


twalpole:
Thanks for this, I was working on a PR to do the same thing, but mine doesn't change the names of the methods being tested (have_text/have_content) which makes the tests invalid.  I'm going to go with my PR instead.

PR #2849, opened 2026-08-25
Add Element#enabled?

luke-hill:
Opposite of disabled, but cannot chain `all` with a `not_to` so adding this in to facilitate it


luke-hill:
@twalpole should be a small review

